PR #20, opened 2020-05-05
First steps for search box

jeroenvandijk:
# A first version of a search box. 

## TODO
- [ ] Add click handler to search results
- [ ] Filtering of results need to be tweaked
- [ ] Sorting of results need to be tweaked
- [ ] Uses db global, should be use passed argument
- [ ] Styling of output

## Screenshot

Athens_Research

### DIFF
--- a/src/cljs/athens/views.cljs
+++ b/src/cljs/athens/views.cljs
@@ -72,45 +72,60 @@
      :page  page/main
      pages-panel)])
 
+(defn highlight-match [query txt]
+  (let [query-pattern (re-pattern (str "(?i)"  "((?<=" query ")|(?=" query "))"))]
+    (map-indexed (fn [i part]
+           (if (re-find query-pattern part)
+             [:span {:key i :style {:background-color "yellow"}} part]
+             part))
+         (clojure.string/split txt query-pattern))))
+
 (defn search-box []
-  (let [*matches (atom [])]
+  (let [*cache (atom {})
+        *match (atom nil)]
     [:div {:style {:position "relative"
                    :display "inline-block"}}
      [:input#find-or-create-input.bp3-input
       {:type "search"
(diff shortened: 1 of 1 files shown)


jeroenvandijk:
With highlighting

Athens_Research

jeroenvandijk:
For later, consider full text search, maybe https://github.com/theronic/datascript-fulltext/

Especially the first hit is a bit slow. I can imagine this gets worse when the db grows. Might need some fulltext search indexing at some point. A first optimization would be to just limit the text to a minimum of characters, but Roam doesn't do this either.

jelmerderonde:
@jeroenvandijk maybe something like this http://elasticlunr.com/ can be integrated quite easily? It isn't ClojureScript but it doesn't look that hard to write a simple wrapper / use it directly using interop.

jeroenvandijk:
Thanks for the suggestion @jelmerderonde! I think we could combine that library with the work that was done in [datascript fulltext approach mentioned before](https://github.com/theronic/datascript-fulltext/blob/master/src/com/theronic/datascript/fulltext.cljs)

I think for now the current search is good enough. We have to focus on the presentation of the results. With the current amount of data in the database we wouldn't really have a proper test case either. But I'm sure your suggestion will be of help in the next phase when we can import full databases and can write new data into Athens

tangjeff0:
Closed by #128, main feature issue now #126

jelmerderonde:
Just came across this article: https://www.notion.so/Product-Search-in-Clojure-DataScript-3d467efc55244f3693127f8d9da5f262

Interesting stuff we might want to use if we ever revisit this issue.

tangjeff0:
@jelmerderonde incredible find. Loved how he did fuzzy search, tokenization, and direct index access.

PR #29, opened 2020-05-06
fix: closes #11

### LINKED ISSUE #11: Refreshing Athens on a non-root page causes app to freezes
Non-root page example: `#/page/AVOHMK6bK`

My guess for why this is is that when the app is mounting, the datoms have not yet been loaded from Github. The solution would probably be to show a loading UI, and let all datascript queries wait in a queue until the datoms have been loaded.

tangjeff0:
closes #11 

Uses `day8.re-frame.async-flow-fx`

### DIFF
--- a/src/cljs/athens/views.cljs
+++ b/src/cljs/athens/views.cljs
@@ -68,12 +68,15 @@
      pages-panel)])
 
 (defn main-panel []
-  (let [current-route (subscribe [:current-route])]
+  (let [current-route (subscribe [:current-route])
+        loading (subscribe [:loading])]
     (fn []
-      [:div
-       [alert]
-       ;;[:h1 (str "Hello World")]
-       [:div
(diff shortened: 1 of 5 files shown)


src/cljs/athens/views.cljs
```diff
@@ -68,12 +68,15 @@
...
-      [:div
-       [alert]
-       ;;[:h1 (str "Hello World")]
-       [:div
-        [:a {:href (rfee/href :pages)} "All /pages"]
-        [:span {:style {:margin 0 :margin-left 10}} "Current Route: " [:b (-> @current-route :path)]]]
-       [match-panel (-> @current-route :data :name)]])))
+      [alert]
+      (if @loading
+        [:h4 "Loading... (at least it'll be faster than Roam)"]
```
jeroenvandijk, line 76: 🤣 🤣 🤣 🤣 🤣

PR #40, opened 2020-05-08
Add build step to export app artifacts

jeroenvandijk:
This build step to export the html app can help us to do quick tests and compare versions

Note that this production release is not working well (not due the build script). I'll make a separate pull request to fix that.

Artifact can be found here https://github.com/jeroenvandijk/athens/actions/runs/99358303


jeroenvandijk:
Closed by #41

PR #47, opened 2020-05-10
Add: trigger vercel deploy in gh action

yqrashawn:
This commit add the github action to deploy `resource/public` to vercel. There's a demo PR [here](https://github.com/yqrashawn/athens-fork-for-gh-actions/pull/5). I have to create an unforked repo to tweak the gh action.

There're several things need to be done before making this possible.
1. create a vercel account at https://vercel.com/
2. create a vercel token at https://vercel.com/account/tokens as `VERCEL_TOKEN`
3. install the [vercel cli](https://github.com/zeit/now/tree/master/packages/now-cli) with `yarn global add vertical` or `npm install vercel -g`
4. cleanup `resource/public` folder, remove files generated by `lein dev` (too much files may reach vercel's free account upload limit)
5. at root dir of athens, run `vercel --token=$VERCEL_TOKEN`
	- `Set up and deploy “/path/to/athens”? [Y/n]`: **yes**
	- `Link to existing project? [y/N]`: **no**
	- `What’s your project’s name?` **athens**
	- `In which directory is your code located?` **./**
	- `Want to override the settings? [y/N]`: **yes**
	- `Which settings would you like to overwrite (select multiple)? Build Command, Output Directory, Development Command`: **choose all three options with space and arrow key**
		- `Build Command`: **leave empty**
		- `Output Directory`: **resources/public**
		- `Development Command`: **leave empty**

6. there will be a `.vercel/project.json` file created in athens root dir
7. open https://github.com/athensresearch/athens/settings/secrets
	- add `VERCEL_TOKEN` same as above
	- add `VERCEL_ORG_ID`, `VERCEL_PROJECT_ID` with the value `orgId`, `projectId` from `.vercel/project.json`

Then the action should work fine.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,6 +96,20 @@ jobs:
       - name: Release 
         run: |
           RELEASE_NAME=${{ steps.athens-version.outputs.release-name }} script/build/athens-app
+
+      - uses: amondnet/now-deployment@v2
+        # add github secrets at https://github.com/athensresearch/athens/settings/secrets
+        with:
+          # only comment on pr events
+          github-comment: ${{ contains(github.event_name, 'pull_request') }}
+          # vercel token, generated here https://vercel.com/account/tokens
+          zeit-token: ${{ secrets.VERCEL_TOKEN }} 
+          # gh action built-in env
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # below two is based on vercel's project linking feature
+          # instructions: https://github.com/amondnet/now-deployment#project-linking
+          now-org-id: ${{ secrets.VERCEL_ORG_ID}}
+          now-project-id: ${{ secrets.VERCEL_PROJECT_ID}}
           
       - uses: actions/upload-artifact@v1
         with:
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@
 
 # Emacs files
 .#*
+
+.vercel


cpaika:
For what it's worth, looks good to me.  Also didn't know Github had a secrets management feature thats awesome.

jeroenvandijk:
@yqrashawn Thank you for the research and the well structured pull request! I'll look into this tomorrow when I'm fresh again :)

jeroenvandijk:
@yqrashawn I followed your instructions; created a Vercel account and added the secrets to the Athens accounts. It doesn't seem to have an effect yet. Any idea? I tried it here after adding the secrets https://github.com/athensresearch/athens/pull/51

Maybe it has something to do with the fact that pull requests from other repos cannot access these ENV vars? ([source](url))

yqrashawn:
@jeroenvandijk It seems there’s missing env in the Vercel deploy section in the gh action log. You need to add `VERCEL_TOKEN` `VERCEL_ORG_ID` `VERCEL_PROJECT_ID` in github secrets.  

Checkout the difference between the below two build. 

https://github.com/athensresearch/athens/runs/664750937?check_suite_focus=true#step:8:3

https://github.com/yqrashawn/athens-fork-for-gh-actions/runs/660499402?check_suite_focus=true#step:7:4

yqrashawn:
> Maybe it has something to do with the fact that pull requests from other repos cannot access these ENV vars?

I’ll check that later. There’s some security logic behind this for sure. Maybe PR from forked repo can’t access gh secret.

yqrashawn:
https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/td-p/30678

GitHub don’t allow PR from forked repo to read the secret yet.

cpaika:
@yqrashawn Makes sense they can't disclose secrets to fork's but that's a bummer they haven't implemented a solution.  

What do you think is the best workflow then?  People could be granted contributor status to be able to push branches to this repo, so they're PR's would be able to run the github action.  But that hurts the contribution experience for anyone new who submits a PR from a fork since they wont get access to PR deploys.

Other than that we could adopt a separate CI/CD tool other than Github actions?  I've never used it, but the envoy proxy project uses circleci and it seems to integrate well 🤷 https://github.com/envoyproxy/envoy/pull/11148

That being said, Github actions are *so* easy so it's a real bummer to have to adopt something more advanced

cpaika:
Apparently they do have a free tier for open source that looks pretty robust, but its hard to know how many credits we would be using https://circleci.com/open-source/

tangjeff0:
I've used open-source for CircleCI before. It worked well, but I don't know if we exceeded credits 🤷

yqrashawn:
@tangjeff0 The credits will be enough. We have a browser plugin running e2e tests on circleci [here](https://app.circleci.com/pipelines/github/Conflux-Chain/conflux-portal/295/workflows/34b6f768-c86e-4e8f-afbd-b5759024bf54). The credits never run out.

jeroenvandijk:
@yqrashawn I added the ENV's: 
Secrets
Potentially, I could have made a mistake their, but the whole output looks different than in your pull request. So I'm assuming that it's restricted.

As for CircleCI, how would that change things? They don't use env vars? They have a separate deploy step that is fully controlled in the admin?

Regardless of how we get this to work technically, this made me think about security. We actually don't want anyone to be able to make a pull request and have it deploy under our name. This is promoting phishing attacks in our name. **So let's think carefully about that.**

Although, it's not ideally CD, right now every Github Action is push html artifacts that can be downloaded quickly. I hope to have something better than this later.

yqrashawn:
> Regardless of how we get this to work technically, this made me think about security. We actually don't want anyone to be able to make a pull request and have it deploy under our name. This is promoting phishing attacks in our name. So let's think carefully about that.

Vercel has a feature that asks for confirmation before running the ci if the `now.json` file is changed. As far as I know, circleci don't support this feature.

The problem is vercel only supports building projects using a specific framework. I doubt there's java env in their container for building projects.

I think we'll just use the current artifact approach till there's a better solution.

jeroenvandijk:
@yqrashawn Thank you for trying though. Maybe we can revisit Vercel later

Issue #10, opened 2020-04-30
Decide on a CSS library: Tachyons or Tailwind


tangjeff0:
Both [Tachyons](https://tachyons.io/#style) and [Tailwind](https://tailwindcss.com/docs/utility-first) are more so libraries (i.e. a composable design language) than they are frameworks (e.g. not composable Bootstrap or Bulma). Functional programmers probably like the former more 😄.

jeroenvandijk:
I have no experience on the mentioned libraries (only bootstrap). Composable sounds good :) Do you know examples of websites that were build with the libraries? It could be worth to check how it looks and how the underlying CSS and html is structured.

tangjeff0:
I also haven't used these libraries and haven't looked at examples, so there may be even better options. The ideal would be writing CSS with EDN!

tangjeff0:
@teodorlu Have you used any of these libraries?

jeroenvandijk:
I gave it some thought and I wonder if Tachyons and Tailwind are more modular than something like [Sass](https://sass-lang.com/guide). 

Another usefool tool in this category could be the Clojure library [garden](https://github.com/noprompt/garden). It doesn't give you any styling out of the box, but it might make css easier to work with. 

You could create helper functions that either add certain classes to an element or certain styling. It would not matter much. In these helper functions you could choose to use one of these frameworks.

```
(defn +rounder-corners [attrs]
   (update attrs :class str " tailwind class "))

(defn +rounder-corners [attrs]
   (assoc attrs :style (garden/css {:text-decoration "none"})))

[:div (+rounded-corners {:on-click (fn [_} ...)}) "content here"]
```

If the helpers are given semantic names and organized properly, we might be able to benefit from better reuse as well. Maybe even change themes easily.

tangjeff0:
Themes open up a big can of worms 🤪. Even with better defaults than Roam, users will still want to customize styling, and this is especially true (and encouraged!) for open-source software. Thus, we must also consider what library or framework would be composable and flexible not just for developers, but also for end-users.

Which brings us to the can of worms: isn't styling just one application within a larger plugin architecture? Overriding CSS with a Chrome extension is simply inferior to toggling themes in, say, an IDE. That said, a plugin architecture sounds hard, even one that only handles styling. But if we do go down this route, I bet we can get inspiration from open-source VS Code and Atom :)

palashkaria:
Some general thoughts on this:

[...]

PR #54, opened 2020-05-14
Add basic styling

Smittyvb:
See #10. This uses [Garden](https://github.com/noprompt/garden) for some basic styling. Now, the body uses the browser's default sans-serif font, instead of not specifying one (some browsers use Time New Roman in this case, some use Arial, some use Tinos (a Times New Roman clone)). Also, it makes sure there is enough room for the dates on the home page.
I'm not very experienced with Clojure, but it does appear to work correctly to me.

image

### DIFF
--- a/src/cljs/athens/style.cljs
+++ b/src/cljs/athens/style.cljs
@@ -3,18 +3,29 @@
 (defn loading-css
   []
   (fn []
-    [:style (css [
-      :body {:font-family "sans-serif"}
-      :#loading-text {:font-size "1.3rem"}
-    ])]
+    [:style (css
+      [:body {
+        :font-family "sans-serif"
+        :font-size "1.3rem"
+      }]
+    )]
   )
 )
 
 (defn main-css
   []
   (fn []
-    [:style (css [
-      :body {:font-family "sans-serif"}
-    ])]
+    [:style
+      (clojure.string/join "" [
+          (css [:body {:font-family "sans-serif"}])
+          (css :.pages-table [
+            :th {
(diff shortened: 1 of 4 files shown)


src/cljs/athens/style.cljs
```diff
@@ -0,0 +1,31 @@
+(ns athens.style (:require  [garden.core :refer [css]]))
+
+(defn loading-css
+  []
+  (fn []
```
jeroenvandijk, line 5: Not sure if this nesting is needed

src/cljs/athens/style.cljs
```diff
@@ -0,0 +1,31 @@
+(ns athens.style (:require  [garden.core :refer [css]]))
```
jeroenvandijk, line 1: Normally this goes on two lines

src/cljs/athens/style.cljs
```diff
@@ -0,0 +1,31 @@
...
+            :th {
+              :font-weight "bold"
+              :min-width "11em"
+            }
+          ])
+        ]
+      )
+    ]
+  )
+)
```
jeroenvandijk, line 31: We don't do this in clojure 😎 All brackets together. Most of us use an editor that can help us keeping track of the nesting. I'm adding a tool to check the styling so you can check these things beforehand

jeroenvandijk:
Thank you @Smittyvb and congratulations on your first pull request!

See my review comments. There are some formatting issues, but we will fix those later. Probably we also want to add the css slightly different, but we are gonna go with this for now! Thank you!

PR #55, opened 2020-05-14
Add code style + unused vars checks

jeroenvandijk:
### Style check 
fails if fixes are found ([last run](https://github.com/jeroenvandijk/athens/runs/674359114#step:3:21))

### Unused vars check
fails if superfluous vars are found ([last run](https://github.com/jeroenvandijk/athens/runs/674359202#step:4:118))

### Next step
Fix the checks and/or tweaking the configuration. Probably good for beginners. Link to docs is in `script/style` and `script/carve`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,6 +75,20 @@ jobs:
       - name: Style 
         run: |
           script/style
+                 
+  carve:
+    # ubuntu 18.04 comes with lein + java8 installed
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+          submodules: 'true'
+          
+      - name: Carving unused vars 
+        run: |
+          script/carve
 
   build-app:
     needs: [test]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,8 @@ jobs:
           fetch-depth: 1
           submodules: 'true'
           
-      - name: Carving unused vars 
+      - uses: DeLaGuardo/setup-clojure@2.0
+        name: Carving unused vars 
         run: |
           script/carve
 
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -61,6 +61,37 @@ jobs:
...
+        with:
+          fetch-depth: 1
+          submodules: 'true'
+          
+      - name: Style 
+        run: |
+          script/style
+                 
+  carve:
+    # ubuntu 18.04 comes with lein + java8 installed
```
tangjeff0, line 80: I'm using `java11`. Should we try to stick to one version?

tangjeff0:
Also the `carve` script seems to fail, but not because of the check. It looks like it doesn't even run.

.github/workflows/build.yml
jeroenvandijk: It's should not matter for most Clojure feature (time API's are in new jdks'). 

I haven't reinstalled java for a long time. Still running Java8. I'm not sure if I'm unique, but I wouldn't enforce it on the dev part, unless it becomes a problem.

On CI we might want to test against multiple version indeed. Or choose a version that we specifically support. I guess java11 makes sense, although I'm still focussed on Java8 myself. I don't see big issues here

jeroenvandijk:
@tangjeff0 Regarding `carve` it's hidden in the output ```
({:filename "src/cljs/athens/core.cljs", :row 25, :col 1, :ns athens.core, :name init} {:filename "src/cljs/athens/db.cljs", :row 66, :col 1, :ns athens.db, :name help-url} {:filename "src/cljs/athens/db.cljs", :row 67, :col 1, :ns athens.db, :name ego-url} {:filename "src/cljs/athens/views.cljs", :row 50, :col 1, :ns athens.views, :name home-panel})
```

But could definitely improve the output here. Merging for now. Will fix it later

PR #57, opened 2020-05-14
Several minor UI improvements

Smittyvb:
This incorporates several minor UI changes:
- `style.cljs` is now formatted idomatically.
- Fixed two lint warnings about the order of requires.
- There is now a loading message while the JavaScript loads.
- Alternating rows in the homepage table are light gray to make it easier to see across a row.
- For some reason, some dates are stored `0` instead of an actual date. These are now displayed as "(unknown date)" in a dark gray.
- Added a gap between links in the left sidebar.

### DIFF
--- a/src/cljs/athens/views.cljs
+++ b/src/cljs/athens/views.cljs
@@ -5,8 +5,7 @@
    [athens.style :as style]
    [re-frame.core :as rf :refer [subscribe dispatch]]
    #_[reitit.frontend :as rfe]
-   [reitit.frontend.easy :as rfee]
-   ))
+   [reitit.frontend.easy :as rfee]))
 
 (defn about-panel []
   [:div [:h1 "About Panel"]])
@@ -16,6 +15,10 @@
         file (.. e -target -files (item 0))]
     (set! (.-onload fr) #(dispatch [:parse-datoms (.. % -target -result)]))
     (.readAsText fr file)))
+
+(defn- date-string [x] (if (< x 1)
+                         [:span {:class "unknown-date"} "(unknown date)"]
+                         (.toLocaleString  (js/Date. x))))
 
 (defn table
   [nodes]
@@ -34,9 +37,8 @@
       ^{:key id}
       [:tr
        [:td {:style {:height 24}} [:a {:href (rfee/href :page {:id bid})} title]]
-       [:td (.toLocaleString  (js/Date. c-time))]
-       [:td (.toLocaleString  (js/Date. e-time))]
-       ])]])
+       [:td (date-string c-time)]
+       [:td (date-string e-time)]])]])
 
 (defn pages-panel []
   (let [nodes (subscribe [:pull-nodes])]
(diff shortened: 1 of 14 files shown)


src/cljs/athens/views.cljs
```diff
@@ -88,9 +90,8 @@
       [alert]
       (if @loading
         [:div
-          [style/loading-css]
-          [:h4 {:id "loading-text"} "Loading... (at least it'll be faster than Roam)"]
-        ]
+         [style/loading-css]
+         [:h4 {:id "loading-text"} "Loading database... (at least it'll be faster than Roam)"]]
```
jeroenvandijk, line 94: Now this text is duplicated? Do we need it in index.html?

src/cljs/athens/views.cljs
```diff
@@ -61,10 +64,9 @@
        [:div [:a {:href (rfee/href :pages)} "All /pages"]]
        [:div [:span {:style {}} "Current Route: " [:b (-> @current-route :path)]]]
        [:div {:style {:border-bottom "1px solid gray" :margin "10px 0"}}]
-       [:ol {:style {:padding 0 :margin 0 :list-style-type "none"}}
+       [:ul {:class "left-sidebar"}
```
jeroenvandijk, line 67: Instead of adding classes to extract the css. Maybe we can add functions so we don't have to think about classes? E.g. we could have a function in styles +sidebar. We would reference it here as `style/+sidebar`.

If we use styles or classes it wouldn't matter. We can swap the implementation there, e.g.
```clojure
(defn +sidebar [attrs]
  (update attrs :style merge {:padding 0 :margin 0 :list-style-type "none"}))

(defn +sidebar [attrs]
  (update attrs :class (partial str " ") "left-sidebar"))
```

With `carve` in #55 we can remove unused style functions and keep styles.cljs small. What do you think?

src/cljs/athens/views.cljs
tangjeff0: Yeah, also can we remove the Roam reference in all cases? I was feeling cheeky that day. 😂

Smittyvb:
@jeroenvandijk I've removed the Roam reference, and switched to functions instead of classes.

src/cljs/athens/views.cljs
```diff
@@ -16,9 +16,13 @@
     (set! (.-onload fr) #(dispatch [:parse-datoms (.. % -target -result)]))
     (.readAsText fr file)))
 
+(defn- date-string [x] (if (< x 1)
+                         [:span (style/+unknown-date {}) "(unknown date)"]
```
jeroenvandijk, line 20: I wonder where this is coming from. Is this a problem of the import data?

jeroenvandijk: small note, I think `cljsstyle` doesn't catch this, but the `(if ...)` should be on the next line

jeroenvandijk: Athens_Research

The pages that don't have a date are pages of a day (from Daily Notes_

PR #70, opened 2020-05-16
Styling refactoring

Bardia95:
Emulated the Roam editable blocks functionality but with the HTML `:content-editable` attribute. Originally, I tried using re-frame state to toggle the editing state for each block so it would switch from being a span to a textarea. We should be careful that this method we are using now will work with persisting edits in the future.

I also made some styling refactoring, moving some styling to the `stylings.cljs`, although for the bullets and controls sections, when I moved the stylings outside of inline, the styling would break. I couldn't see to troubleshoot this so I left it as a TODO.

The final change I made was moving the on-block-click function to a global function so when we implement shift-click for the sidebar and other potential ways of clicking, we just add a conditional to the global function.

### DIFF
--- a/src/cljs/athens/page.cljs
+++ b/src/cljs/athens/page.cljs
@@ -14,7 +14,7 @@
       [:div {:class "content-block"}
        (doall
         (for [ch (:block/children @block)]
-          (let [{:block/keys [uid string open children editing] dbid :db/id} ch
+          (let [{:block/keys [uid string open children] dbid :db/id} ch
                 children? (not-empty children)]
             ^{:key uid}
             [:div
@@ -32,11 +32,7 @@
                                             :cursor         "pointer" :display "inline-block" :background-color "black"
                                             :vertical-align "middle"}
                                  :on-click #(on-block-click uid)}]]]
-              (if editing
-                [:textarea.editing {:auto-focus true
-                                    :on-blur #(dispatch [:block/toggle-editing dbid editing])
-                                    :value (parse string)}]
-                [:span.text {:on-click #(dispatch [:block/toggle-editing dbid editing])} (parse string)])]
+              [:span.text {:content-editable true} (parse string)]]
              (when open
                [:div {:style {:margin-left 20}}
                 [render-blocks uid]])])))])))
--- a/src/cljs/athens/page.cljs
+++ b/src/cljs/athens/page.cljs
@@ -4,10 +4,6 @@
             #_[reitit.frontend.easy :as rfee]
             #_[reagent.core :as reagent]))
 
-
-(defn on-block-click [uid]
-  (dispatch [:navigate :page {:id uid}]))
-
 (defn render-blocks []
   (fn [block-uid]
     (let [block (subscribe [:block/children-sorted [:block/uid block-uid]])]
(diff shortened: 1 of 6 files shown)


src/cljs/athens/page.cljs
```diff
@@ -4,6 +4,10 @@
             #_[reitit.frontend.easy :as rfee]
             #_[reagent.core :as reagent]))
 
+
+(defn on-block-click [uid]
+  (dispatch [:navigate :page {:id uid}]))
```
tangjeff0, line 9: This seems like a really helpful function. I would move it to `router.cljs` and refactor any other occurrences.

src/cljs/athens/page.cljs
```diff
@@ -68,33 +82,39 @@
...
                              {:style {:cursor "pointer"}
-                              :on-click #(dispatch [:navigate :page {:id uid}])}
+                              :on-click #(on-block-click uid)}
                              (or string title)]))
                         @parents))]
-       [:h2 {:style {:margin 0}} (str "• " (:block/string @node))]
-       [:div {:style {:margin-left 20}}
+       [:div
+        {:style {:margin 0}
+         :content-editable true}
```
tangjeff0, line 90: Sounds like content-editable is a much easier/better solution than `textarea`. Why then do you think Roam uses `textarea`?

src/cljs/athens/page.cljs
Bardia95: Ok sounds good!

Bardia95:
@tangjeff0 in regards to the styling/linting changes, what if I just kept those in this PR, changed the name to reflect what's actually happening and then submit a new pull request for the content-editable/on-block-click changes? Seems easier to do it this way than the other way around?

src/cljs/athens/page.cljs
Bardia95: There's two potential reasons, one I think is more likely than the other. The first, unlikely, reason is that textarea is a form element so using content-editable means having to manually handle submission (seems trivial using on blur). The second, more likely reason, is that content-editable preserves the styling unlike textarea, so I'm thinking maybe they tried content-editable first but were not able to implement their custom styling syntax with the way content-editable works. We won't know until we try.

tangjeff0:
@Bardia95 Yeah if you want to split it into 2 PRs that would work too.

src/cljs/athens/page.cljs
Bardia95: There's two potential reasons, one I think is more likely than the other. The first, unlikely, reason is that textarea is a form element so using content-editable means having to manually handle submission (seems trivial using on blur). The second, more likely reason, is that content-editable preserves the styling unlike textarea, so I'm thinking maybe they tried content-editable first but were not able to implement their custom styling syntax with the way content-editable works. Or they just wanted the user to see plain unformatted text when they are editing and it is a design decision rather than a technical one. We won't know until we reach that stage.

Bardia95:
@tangjeff0 let me know if its alright now! I'll open up a new PR for the editable-blocks and submit it in a couple hours

tangjeff0:
Looks good @Bardia95. I'll wait for @jeroenvandijk to merge and comment on final call regarding style guidelines.

jeroenvandijk:
@Bardia95 If I'm correct there are at least two types of changes here. Code formattting and css style refactoring? I think this should be done in two seperate commits. It's really hard for me to review this. 

Especially since we have some other pull requests waiting this is a recipe for merge conflicts. Can you please make the diffs as small as possible? We can also do the codestyle refactoring incrementally if it's too much work to put it into two seperate commits.

Bardia95:
I'm going to close this pull request and then start a new one with smaller commits.

Issue #66, opened 2020-05-15
Daily Notes


tangjeff0:
# Tasks

- [ ] calendar button to select any date
- [x] show alert when user tries to edit title
- [x] hide linked/unlinked references if there are none #191 
- ~ notional dates
- ~ jump to Timeline
- [X] scroll to create new page #212 

# Original Post

<details>
  <summary>Click to open</summary>

When a user open their Roam, they should automatically generate a notes page for the current 
date. User should be able to scroll down to previous days. 

Note: these Daily Notes pages are different from other pages in that their titles are not allowed to be changed. Attempt to do so leads to a warning.

# Warning
![Screen Shot 2020-05-15 at 7 44 29 PM](https://user-images.githubusercontent.com/8952138/82104328-7fdb2680-96e4-11ea-9de4-d0fa2cdd155f.png)

# Screenshot
![image](https://user-images.githubusercontent.com/8952138/82104275-40acd580-96e4-11ea-8be8-631c24f71d64.png)

</details

jeroenvandijk:
TASKS

- [ ] Core work: scrolling down and opening a new page (add [onscroll event listener](https://javascript.info/onscroll))
- [ ] Add warning when title of date page is changed (possibly a regexp check on the page title is enough for this. Maybe no need to check if it is a special page)
- [ ] Bonus: when you are in the `Daily Notes` view and times passes into the next day (in your timezone), a new day should be added. Could be done by triggering a re-render in X seconds via (setTimeout ...)
- [ ] Bonus: to keep the app performant endless scrolling down should also clean up day views on top. With scrollup event these views would have to be added again

Bardia95:
Working on it with @pinned5th as part of ClojureFam, looking to get a working version by end of week!

tangjeff0:
@Bardia95 that's awesome — glad you are setting a deliverable for yourself! I'm personally more curious about how you two are mapping these scopes, and what scopes will be discovered and refactored over time. Please update this issue at various checkpoints so you can teach the rest of us :)

Bardia95:
@tangjeff0 right now, the path is uncharted, my plan for today is to experiment with how we could accomplish this feature and update with any findings, and then once we've scoped out a bit will update with what we're each working on. We have planned a pair programming session once we've scoped out more granularly

tangjeff0:
> Working on it with @pinned5th as part of ClojureFam, looking to get a working version by end of week!

How are scopes progressing @Bardia95 ?

[...]

Issue #61, opened 2020-05-15
Editing page title to existing page prompts merge


tangjeff0:
# Demo

![edit-page-title-prompt-merge](https://user-images.githubusercontent.com/8952138/82101211-e9096c80-96d9-11ea-9b29-c3e3922fb31b.gif)

jelmerderonde:
I have a couple of questions before I start implementing this:

1. If I understand re-frame correctly, the flow should be something like this:
- Editing the title triggers the `:node/rename` event.
- If no page with `new-title` exists, proceed with the rename.
- Otherwise, set some state in de re-frame db to show the merge prompt. Depending on the user response trigger a new event to `:node/merge` or cancel the whole thing.
- If the user navigates away from the page before responding to the prompt, we should clear the state showing the prompt. Something to add to the `:navigated` event?
Am I on the right track with this?

2. What is a merge exactly? Should it be implemented as well or does it have its own issue?

PR #71, opened 2020-05-17
feat: add event :node/rename that also updates referencing blocks

### LINKED ISSUE #62: Editing page title renames all occurrences
Two edge cases:

- Existing page: #61
- Daily Notes: #66 

# Demo

rename-page-title

jelmerderonde:
First PR for this project. Any and all feedback is welcome. Although the requested functionality is there, you can't use it yet from the ui, because editing titles isn't possible yet. 

So I guess this PR is mainly here for feedback, I can update it when #70 is merged. If another workflow is preferred, please let me know!

- TODO: use the event when updating title is supported in the ui
- TODO: handle edge cases #61 and #66

close #62

### DIFF
--- a/src/cljs/athens/page.cljs
+++ b/src/cljs/athens/page.cljs
@@ -1,5 +1,6 @@
 (ns athens.page
   (:require [athens.parser :refer [parse]]
+            [athens.patterns :as patterns]
             [athens.router :refer [navigate-page toggle-open]]
             [re-frame.core :refer [subscribe dispatch]]
             #_[reitit.frontend.easy :as rfee]
@@ -45,18 +46,6 @@
                [:div {:style {:margin-left 20}}
                 [render-blocks uid]])])))])))
 
-; match [[title]] or #title or #[[title]]
-(defn linked-pattern [string]
-  (re-pattern (str "("
-                   "\\[{2}" string "\\]{2}"
-                   "|" "#" string
-                   "|" "#" "\\[{2}" string "\\[{2}"
-                   ")")))
-
-; also excludes [title] :(
-(defn unlinked-pattern [string]
-  (re-pattern (str "[^\\[|#]" string)))
-
 (defn block-page []
   (fn [id]
     (let [node (subscribe [:node [:block/uid id]])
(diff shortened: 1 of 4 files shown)


jeroenvandijk:
Note from Jelmer in Discord about formatting of paredit
```
Jelmer de RondeToday at 6:03 PM
@Jeroen van Dijk Just to let you know, in this case Parinfer would have changed lines 50-51 of events.cljs. The two closing parens would have moved to the end of line 50.
```

src/cljs/athens/page.cljs
```diff
@@ -45,12 +45,12 @@
...
+; e.g.: $1$3$4new-string$2$5
 (defn linked-pattern [string]
-  (re-pattern (str "("
-                   "\\[{2}" string "\\]{2}"
-                   "|" "#" string
-                   "|" "#" "\\[{2}" string "\\[{2}"
-                   ")")))
+  (re-pattern (str "(\\[{2})" string "(\\]{2})"
+                   "|" "(#)" string
+                   "|" "(#\\[{2})" string "(\\]{2})")))
```
jeroenvandijk, line 53: This looks pretty complicated and asks for some examples via tests. Maybe you can extract this functionality into a cljc file and test this behaviour with `deftest` in `test/cljc/athens/page_test.cljc`?

tomisme: Could we use https://github.com/lambdaisland/regal ?

jelmerderonde: @jeroenvandijk I've added some tests ann extracted the patterns to a new cljc file.

jelmerderonde: > Could we use https://github.com/lambdaisland/regal ?

@tomisme, maybe we could. But it might be better to actually use instaparse and the existing grammar for parsing links for this. That way there is only one "pattern" that defines how a link should be detected in Athens.

@jeroenvandijk what do you think?

jeroenvandijk: Yeah regal could be nice at some point, especially when we want to do generative testing. But we don't need it yet, so I think the current changes are fine. Since we have many new clojure people here, it is also good to not make too many steps at once, people have an easier time to understand the steps. We can upgrade the regexp's to Regal later. 

Regarding instaparse, yes we need that for the markdown parsing at least. Instaparse supports regexp's as well. Maybe there is some overlap here. Let's wait and see.

(*) I personally also don't have experience with Regal yet, but I'm sure it's good.

jeroenvandijk:
I've resolved the conflicts in #77 (my bad). Should be fine now. The other TODO's can be done in a different commit

jelmerderonde:
Sounds good, I’ll try and tackle the other todos in another PR later this
week.

PR #72, opened 2020-05-17
Feature/add test runner

wandersoncferreira:
First PR, I think I got the gist of the whole project, very interesting congrats. The content of this PR is more related to setup to execute cljs and cljc tests.

I also wrote a test to `parse-tuples`, and could not get any example from `parse-hms` function. Are you guys using this one still?

I made some changes to github build pipeline to execute all these tests as well.


src/cljs/athens/db.cljs
```diff
@@ -47,8 +47,8 @@
        (map #(map edn/read-string %))
        (map #(cons :db/add %))))
 
-(defn json-str-to-edn
-  "Convert a JSON str to EDN. May receive JSON through an HTTP request or file upload."
+(defn json-str-to-vector
+  "Convert a JSON str to a clojure vector. May receive JSON through an HTTP request or file upload."
```
jeroenvandijk, line 51: I don't think this is correct? It depends on the input if it's a vector. Can you explain why you changed this?

src/cljs/athens/db.cljs
```diff
@@ -57,10 +57,10 @@
...
   "Deserializes a JSON string into EDN and then Datoms."
   [json-str]
-  (let [edn-data (json-str-to-edn json-str)]
-    (if (coll? (first edn-data))
-      (parse-hms edn-data)
-      (parse-tuples edn-data))))
+  (let [-vector (json-str-to-vector json-str)]
+    (if (coll? (first -vector))
+      (parse-hms -vector)
+      (parse-tuples -vector))))
```
jeroenvandijk, line 63: Same as above

test/cljs/athens/db_test.cljs
```diff
@@ -0,0 +1,39 @@
...
+(deftest parse-tuples
+  (testing "The data format `sample-roam-export-data-tuple` is not a random string vector, it is a vector where every 3 itens should be grouped as (?e,?a,?v)."
+    (let [-vector (db/json-str-to-vector sample-roam-export-data-tuple)
+          ret (db/parse-tuples -vector)]
+      (testing "Notice, the first triplet is dropped because it only indicated the described structure."
+        (is (not= (count ret) (/ (count -vector) 3)))
+        (is (= (count ret) 4)))
+
+      (testing "All tuples must have a `:db/add` first item which will be the used to create the fact inside our database."
+        (is (every? some? (map first ret)))))))
```
jeroenvandijk, line 39: I'm not sure about these tests. They don't really tests any special functionality. JSON parsing is almost a standard function (via javascript). I also thing the test string could be trimmed down

If it's about testing the form of data `clojure.spec` is a better tool for this

jeroenvandijk:
@wandersoncferreira Btw, thanks for taking the time to create this pull request!

src/cljs/athens/db.cljs
tangjeff0: Agree @jeroenvandijk. I think `edn` is a more descriptive while still generic name than `vector`.

test/cljs/athens/db_test.cljs
tangjeff0: I would agree with @jeroenvandijk that a test like this, which is not 100% necessary, could ultimately be done with something like `spec` or `malli`.

src/cljs/athens/db.cljs
wandersoncferreira: Yes, I might have changed this back and forth sometimes rsrsrs not clear if was valuable or not have this argument here. Anyway, the function signature for [js->clj](https://cljs.github.io/api/cljs.core/js-GTclj) just states the conversion from js array to clojurescript vector and to be honest I was expecting the output of ":block/order" to be a keyword instead of a string from this function when I read `json-str-to-edn`. 

I don't mind changing it back again.

test/cljs/athens/db_test.cljs
wandersoncferreira: Yes, indeed a `malli` spec would be better off here. I will submit a change to verify if each tuple is a valid fact to be transacted

tangjeff0:
Closed because we setup test runners

PR #75, opened 2020-05-18
Cljstyle fix

Bardia95:


Bardia95:
@jeroenvandijk performed the fix on the whole directory

PR #76, opened 2020-05-18
Fix Carve build step

jeroenvandijk:
No more warnings about unused vars. If all goes well, this step can be enabled as a required check for PR's.

Also added caching for the retrieval of the deps. This was cluttering the output. If everything is ok tests show '()'. If not, a list of vars that are unused.

Last (successful) example run https://github.com/jeroenvandijk/athens/runs/686406577?check_suite_focus=true

### DIFF
--- a/.carve_ignore
+++ b/.carve_ignore
@@ -1,5 +1,9 @@
-; REVIEW Can some of these items be removed (later)?
+;; Init is called by shadow-cljs (https://github.com/athensresearch/athens/blob/c90f3023a1c4480b50ed61d07f973d8c29e08bf8/shadow-cljs.edn#L6)
+athens.core/init
+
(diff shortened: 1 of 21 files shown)


.carve_ignore
```diff
@@ -0,0 +1,6 @@
+; REVIEW Can some of these items be removed (later)?
+athens.db/help-url
+athens.db/ego-url
+athens.views/home-panel
+athens.core/init
+user/str-kw-mappings
```
tangjeff0, line 6: - URLs could still be used if we want to examine other DBs. especially `help-url`
- `home-panel` can be removed
- `str-kw-mappings` will be used for JSON import I believe
- not sure why `init` is not being used

PR #79, opened 2020-05-18
CSS composition

jeroenvandijk:
@tangjeff0, @Smittyvb and others what do you think about this way of composing styles in a separate style file.

@Smittyvb already made a start with some first functions. Here I've extracted some boilerblate and showed how you can compose styles and classes together.

### Benefits of this approach
* (Logic) of styles are in one place
* Unused styles are removed because of `script/carve` that looks for unused vars
* Refactoring of styles (inline to classes, different frameworks etc) doesn't affect the other view code as that part can keep using the function and doesn't care about it's implementation.

(More background here https://github.com/athensresearch/athens/issues/10)


tangjeff0:
I think this looks great @jeroenvandijk . Let's get it merged in and point to it as an example for how we should write composable styles going forward.

Issue #78, opened 2020-05-18
Setup Sentry error tracking


jeroenvandijk:
(I've applied for a Sentry opensource license 🤞)

Sentry's error tracking is useful to catch errors in production (artificacts) early. This can monitor regressions and help us determine the general health of the application.

https://sentry.io/organizations/athensresearch/

tangjeff0:
Awesome, should we look at other services like Synk?

jeroenvandijk:
Oh I didn't hear about Synk before (too much backend lately maybe). If you have good experience with that, sure.

bsvin33t:
I think Synk and Sentry go hand in hand. Synk is for vulnerability detection and Sentry is for general application error tracking.   😄

noorfathima11:
Can I get this assigned?

tangjeff0:
Sure! Can you walk me through what needs to be done? @noorfathima11

sid597:
Solved by #600

PR #77, opened 2020-05-18
feat: add event :node/rename that also updates referencing blocks (merged with master)

jeroenvandijk:
Merge PR of #71

PR #86, opened 2020-05-22
Adds title component that supports page renaming

jelmerderonde:
Next step in dealing with #62. Haven't yet dealt with #61 and #66. I've also realized that I should deal with renames that include `#hash-style-links` being renamed to `#[[title that includes whitespace]]`.

I don't really expect this to be merged yet until I deal with these issues, but would appreciate some feedback whether I'm on the right track with this.

### DIFF
--- a/src/cljs/athens/page.cljs
+++ b/src/cljs/athens/page.cljs
@@ -70,6 +70,9 @@
        [:div {:style {:margin-left 20}}
         [render-blocks (:block/uid @node)]]])))
 
+(def enter-keycode 13)
+(def esc-keycode 27)
+
 (defn title-comp [title]
   (let [s (reagent/atom {:editing false
                          :current-title title
@@ -86,8 +89,8 @@
                  :on-change #(swap! s assoc :new-title (-> % .-target .-value))
                  :on-blur save!
                  :on-key-down #(case (.-which %)
-                                 13 (save!) ; save on enter
-                                 27 (cancel!) ; cancel on esc
+                                 enter-keycode (save!) ; save on enter
+                                 esc-keycode (cancel!) ; cancel on esc
                                  nil)}]
         [:h2 {:on-click (fn [_] (swap! s #(-> %
                                               (assoc :editing true)
--- a/src/cljs/athens/page.cljs
+++ b/src/cljs/athens/page.cljs
@@ -77,8 +77,8 @@
   (let [s (reagent/atom {:editing false
                          :current-title title})
         save! (fn [new-title]
-                (dispatch [:node/rename (:current-title @s) new-title])
-                (swap! s assoc :editing false))
+                (swap! s assoc :editing false)
+                (dispatch [:node/rename (:current-title @s) new-title]))
         cancel! (fn [] (swap! s assoc :editing false))]
     (fn [title]
       (if (:editing @s)
--- a/src/cljs/athens/page.cljs
+++ b/src/cljs/athens/page.cljs
@@ -3,8 +3,8 @@
     [athens.parser :refer [parse]]
     [athens.patterns :as patterns]
     [athens.router :refer [navigate-page toggle-open]]
-    [re-frame.core :refer [subscribe]]
-    #_[reagent.core :as reagent]
+    [re-frame.core :refer [subscribe dispatch]]
+    [reagent.core :as reagent]
     #_[reitit.frontend.easy :as rfee]))
 
 
@@ -70,14 +70,37 @@
        [:div {:style {:margin-left 20}}
         [render-blocks (:block/uid @node)]]])))
 
+(defn title [title]
+  (let [s (reagent/atom {:editing false
+                         :current-title title
+                         :new-title title})
+        save! (fn [] (dispatch [:node/rename (:current-title @s) (:new-title @s)])
+                     (swap! s assoc :editing false))
+        cancel! (fn [] (swap! s #(-> %
+                                     (assoc :editing false)
+                                     (assoc :new-title (:current-title @s)))))]
+    (fn [title]
+      (if (:editing @s)
+        [:input {:value (:new-title @s)
+                 :auto-focus true
+                 :on-change #(swap! s assoc :new-title (-> % .-target .-value))
+                 :on-blur save!
+                 :on-key-down #(case (.-which %)
+                                 13 (save!) ; save on enter
+                                 27 (cancel!) ; cancel on esc
+                                 nil)}]
+        [:h2 {:on-click (fn [_] (swap! s #(-> %
+                                              (assoc :editing true)
+                                              (assoc :current-title title)
+                                              (assoc :new-title title))))}
+         title]))))
 
 (defn node-page []
   (fn [node]
     (let [linked-refs   (subscribe [:node/refs (patterns/linked   (:node/title node))])
           unlinked-refs (subscribe [:node/refs (patterns/unlinked (:node/title node))])]
       [:div
-       [:h2
-        {:content-editable true} (:node/title node)]
+       [title (:node/title node)]
        [render-blocks (:block/uid node)]
        [:div
         [:h3 "Linked References"]
--- a/src/cljs/athens/page.cljs
+++ b/src/cljs/athens/page.cljs
@@ -70,7 +70,7 @@
        [:div {:style {:margin-left 20}}
         [render-blocks (:block/uid @node)]]])))
 
-(defn title [title]
+(defn title-comp [title]
   (let [s (reagent/atom {:editing false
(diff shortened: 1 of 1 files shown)


src/cljs/athens/page.cljs
```diff
@@ -70,14 +70,37 @@
...
+                         :new-title title})
+        save! (fn [] (dispatch [:node/rename (:current-title @s) (:new-title @s)])
+                     (swap! s assoc :editing false))
+        cancel! (fn [] (swap! s #(-> %
+                                     (assoc :editing false)
+                                     (assoc :new-title (:current-title @s)))))]
+    (fn [title]
+      (if (:editing @s)
+        [:input {:value (:new-title @s)
+                 :auto-focus true
```
jeroenvandijk, line 85: I think you need to do auto-focus because otherwise you are losing focus, right? How does auto-focus work? Is it perfect?

src/cljs/athens/page.cljs
```diff
@@ -70,14 +70,37 @@
        [:div {:style {:margin-left 20}}
         [render-blocks (:block/uid @node)]]])))
 
+(defn title [title]
```
jeroenvandijk, line 73: a bit dangerous to give the argument the same name as the function (shadowing)

src/cljs/athens/page.cljs
```diff
@@ -70,14 +70,37 @@
...
+  (let [s (reagent/atom {:editing false
+                         :current-title title
+                         :new-title title})
+        save! (fn [] (dispatch [:node/rename (:current-title @s) (:new-title @s)])
+                     (swap! s assoc :editing false))
+        cancel! (fn [] (swap! s #(-> %
+                                     (assoc :editing false)
+                                     (assoc :new-title (:current-title @s)))))]
```
jeroenvandijk, line 81: I would leave the state management of the input to the browser. This gives a better user experience. You can access the current value via the event. This means you can remove the `:on-change` event
```clojure
save! (fn [event]
        (let [new-title (-> event .-target .-value)]
          (dispatch [:node/rename (:current-title @s) event-title])
          (swap! s assoc
                 :editing false
                 :current-title new-title)))
cancel! (fn [] (swap! s assoc :editing false))

File: src/cljs/athens/page.cljs [hunk 4]
```diff
@@ -70,14 +70,37 @@
...
+    (fn [title]
+      (if (:editing @s)
+        [:input {:value (:new-title @s)
+                 :auto-focus true
+                 :on-change #(swap! s assoc :new-title (-> % .-target .-value))
+                 :on-blur save!
+                 :on-key-down #(case (.-which %)
+                                 13 (save!) ; save on enter
+                                 27 (cancel!) ; cancel on esc
```
jeroenvandijk, line 90: Instead of adding the comments, might make more sense to define the codes somewhere:
```
(def enter-keycode 13)
(def esc-keycode 27)
```
 This can be reused by other parts of code and is more robust than the comment (imagine appending a 1 to the 13 in a big pull request and this is missed by the reviewers)

jeroenvandijk:
@jelmerderonde Regarding `I've also realized that I should deal with renames that include #hash-style-links being renamed to #[[title that includes whitespace]].` I think that's a different layer. You only have to care about the edit-ing logic. How this is rendered in other places is a different concern

src/cljs/athens/page.cljs
jelmerderonde: Agreed, will change this

src/cljs/athens/page.cljs
jelmerderonde: Indeed, if I don't add this attribute, the user would have to click twice on the h2 node to actually start editing. I think a limitation of auto-focus is that there can only be one element on the page with this attribute. I'm not sure what happens when there is more than one element on the page.

An alternative could be to make this a form-3 Reagent component, so I get a reference to the DOM node. Then I can call `.focus` on it.

jelmerderonde:
> @jelmerderonde Regarding `I've also realized that I should deal with renames that include #hash-style-links being renamed to #[[title that includes whitespace]].` I think that's a different layer. You only have to care about the edit-ing logic. How this is rendered in other places is a different concern

Agreed, in the sense that it doesn't belong in this PR. But I think I should have dealt with it in my previous PR #77.

src/cljs/athens/page.cljs
jeroenvandijk: Ah I see. I guess adding `:auto-focus` makes sense then. I was thinking you might have added it in relation to my suggestion of "losing focus" in my comment below.

src/cljs/athens/page.cljs
jeroenvandijk: In what scenario would edit's be ignored you think? As a long as you have focus on that field (no escape, no blur), that value is correct, right?

I might be missing something as well though :) I might try it later and show what I mean if it doesn't make sense (maybe it doesn't)

src/cljs/athens/page.cljs
```diff
@@ -70,14 +70,40 @@
...
+(defn title-comp [title]
+  (let [s (reagent/atom {:editing false
+                         :current-title title
+                         :new-title title})
+        save! (fn [] (dispatch [:node/rename (:current-title @s) (:new-title @s)])
+                     (swap! s assoc :editing false))
+        cancel! (fn [] (swap! s #(-> %
+                                     (assoc :editing false)
+                                     (assoc :new-title (:current-title @s)))))]
+    (fn [title]
```
avichalp, line 85: Small suggestion: I would usually name a function when it extends 1 or 2 lines of code. _Ex:_ `(fn my-fn [title] ...)`. The reason is while debugging the name for non-anonymous functions shows up in the trace. What do you think?

jelmerderonde: @avichalp, thanks for the suggestion! I'm not sure about this one. This actually represents a reagent render function, which is a well known pattern and I've never seen (as far as I can remember) a name used here. Curious to hear @jeroenvandijk's opinion about this.

src/cljs/athens/page.cljs
jelmerderonde: You were right, it was possuble to get rid of the `:on-change`.

jelmerderonde:
@jeroenvandijk I think I've handled all your feedback so far. Will take a shot later this weekend at the edge cases.

src/cljs/athens/page.cljs
jeroenvandijk: I'm not sure, I can imagine it could give some extra information during debugging. I have done things like this in deeply nested jvm code, for Clojurescript I haven't been there yet. Right now it feel arbitrary for me, so neutral voice here.

@avichalp does it matter if source maps are enabled or not? Or do you do this for any component fn? Or do you mean any anonymous fn?

Let's wait for some bugs and decide on this later. I'm working on integrating an error tracker #78 Might give some use cases for this proposed method of @avichalp.

jeroenvandijk:
@jelmerderonde Looks good! AFAIC we can merge this PR. If you run `cljstyle fix` the style issue will be fixed.

While testing I saw that the app can get can stuck if you change the title to something short like "a" (endless unlinked references I think). This needs to be researched later.

src/cljs/athens/page.cljs
```diff
@@ -70,14 +70,40 @@
...
 
+(def enter-keycode 13)
+(def esc-keycode 27)
+
+(defn title-comp [title]
+  (let [s (reagent/atom {:editing false
+                         :current-title title})
+        save! (fn [new-title]
+                (dispatch [:node/rename (:current-title @s) new-title])
+                (swap! s assoc :editing false))
```
jeroenvandijk, line 81: I'm wondering if we should reorder the dispatch and swap. The dispatch slows down the swap (re-rendering of the element) and might give a laggy user experience? If you try putting something too short in like "a" the whole interface stalls because of the unlinked reference feature. This is mentioned here #62 . Reversing this order here might give a more direct experience and move some of the computation in the background (of the user)

jelmerderonde:
@jeroenvandijk I've made some minor changes, should now be ready to merge. Would you like me to squash or rebase first?

jeroenvandijk:
@jelmerderonde Squashing and updating of the branch can all be done here. It ends up as one squashed commit

src/cljs/athens/page.cljs
avichalp: I think you guys are right. I had JVM in mind, actually 😅. It might be overkill—also error messages in CLJS are not that horrifying like JVM stack traces.

avichalp: Thanks @jelmerderonde for your PR 🙏

PR #87, opened 2020-05-22
Instaparse bold pattern

Bardia95:
Added the bold pattern to the parser as per Roam/Markdown **bolded text** notation.

Only works like previous patterns where the entire block must be contained within the pattern and no nesting capabilities.

TODO:
- [ ] Support partial bolded blocks
- [ ] Support nested bolded text

Example:
CleanShot 2020-05-22 at 17 37 23

### DIFF
--- a/src/cljs/athens/parser.cljs
+++ b/src/cljs/athens/parser.cljs
@@ -1,18 +1,22 @@
 (ns athens.parser
-  (:require [instaparse.core :as insta]
-            [re-frame.core :refer [subscribe]]
-            [reitit.frontend.easy :as rfee]))
+  (:require
+    [instaparse.core :as insta]
+    [re-frame.core :refer [subscribe]]
+    [reitit.frontend.easy :as rfee]))
+
 
 (declare transform parse)
 
+
 (def parser
   (insta/parser
-   "S = c | link | bref | hash
+    "S = c | link | bref | hash
     <c> = #'(\\w|\\s)+'
     link = <'[['> c <']]'>
     hash = <'#'> c | <'#'> <'[['> c <']]'>
     bref = <'(('> c <'))'>
    "))
+
 
 (defn transform
   "Transforms instaparse output to hiccup."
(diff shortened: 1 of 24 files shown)


jeroenvandijk:
@Bardia95 Looks good! Is this a proposal to merge already? If so, I would suggest to add a simple test case to catch future regressions

Bardia95:
Ok sounds good I’ll do that now!

src/cljs/athens/parser.cljs
```diff
@@ -38,7 +38,10 @@
      :bref (fn [id]
              (let [string (subscribe [:block/string [:block/uid id]])]
                [:span {:style {:font-size "0.9em" :border-bottom "1px solid gray"}}
-                [:a {:href (rfee/href :page {:id id})} (parse (:block/string @string))]]))}
+                [:a {:href (rfee/href :page {:id id})} (parse (:block/string @string))]]))
+     :bold (fn [text]
+             [:span
+              [:strong text]])}
```
roryokane, line 44: I would return just `[:strong text]` – the extra `span` wrapper doesn't add anything in this case.

roryokane:
## “Support partial bolded blocks”

Don't worry about working on this: I got partially-formatted blocks of all existing types already working locally and plan to publish my PR of it tomorrow. I tried editing my local code to include your support for bold, and the partially-bold line on [this page](http://localhost:3000/#/page/a58yqahTE) renders fine:

bold as part of a block working in my local branch

## “Support nested bolded text”

This feature makes sense and we should probably have it eventually. But it would be easier to implement after my partial bolded blocks branch is merged, so I think you can wait.

Interestingly, Roam doesn't fully implement this feature. It seems Roam doesn't allow most types formatting within bold, but it does allow italics:

bold nesting tests in Roam

## “add a simple test case to catch future regressions”

I tried setting up local tests of the parser on my branch but gave up when it seemed too much trouble. I can tell you what I did and where I ran into trouble:

1. Move `src/cljs/athens/parser.cljs` to `src/cljc/athens/parser.cljc`
1. Create `src/test/athens/parser-test.clj` and pattern it off the other tests in that folder
1. Then I noticed that `parser.cljc` was getting a parse error. This parse error was related to new code in my local branch: I used a ClojureScript-only form `:refer-macros` to import the `defparser` macro. Because of that form, the file couldn't be compiled as `cljc`. I didn't know how else to import the `defparser` macro, so I gave up on making the code `cljc` and kept it `cljs`.

Since you didn't add any ClojureScript-only code so far you probably won't have any problem after following steps 1 and 2. Then you can run `lein test` to try out your new test case.

Bardia95:
Ok great, thanks for saving me the tie @roryokane!

jeroenvandijk:
@Bardia95 You can (probably) use reader conditionals to solve your issue with refer-ing macros https://bauerspace.com/refer-a-macro-with-cljc/

PR #88, opened 2020-05-23
fix(parsing): fix red parse failures in block contents and refactor parser

roryokane:
(This code is theoretically finished but I want to double-check it after a good night's sleep before I ask for a review and merge.)

Before:

before fixing parsing

After:

after fixing parsing

The parse failures were because lines beginning with punctuation such as `{` were neither word characters (`\w`) or space characters (`\s`). I fixed that with the `any-char` rule.

The `any-char` rule matches `\w\W` instead of `.` to ensure it also matches newlines, no matter how Clojure and ClojureScript’s regexes work. This is desired in normal text, but probably not in links and hash-tags. That’s one reason for my comment about planning to move away from the `any-chars` rule.

I could have fixed it by just changing the `c` regexp to `#'(\\w|\\W)+'`, but that would lead to other problems down the line. See https://github.com/Engelberg/instaparse#regular-expressions-a-word-of-warning. The downside of doing it repetition in the parser is that I need to combine the individually-matched characters in the transformer later, with `join` and `combine-adjacent-strings`.

Other changes:

- Use clearer names for the parts of the parser
- Use `defparser` to precompile the parser for ClojureScript (following the advice in https://github.com/Engelberg/instaparse#defparser)
- Add classes to output elements to aid in future styling and debugging
- more changes from later commits I will squash

### DIFF
--- a/src/cljs/athens/parser.cljs
+++ b/src/cljs/athens/parser.cljs
@@ -1,50 +1,71 @@
 (ns athens.parser
   (:require
-    [instaparse.core :as insta]
+    [athens.parse-helper :as parse-helper]
+    [instaparse.core :as insta :refer-macros [defparser]]
     [re-frame.core :refer [subscribe]]
     [reitit.frontend.easy :as rfee]))
 
 
-(declare transform parse)
+(declare block-parser transform parse)
 
 
-(def parser
-  (insta/parser
-    "S = c | link | bref | hash
-    <c> = #'(\\w|\\s)+'
-    link = <'[['> c <']]'>
-    hash = <'#'> c | <'#'> <'[['> c <']]'>
-    bref = <'(('> c <'))'>
-   "))
+;; Instaparse docs: https://github.com/Engelberg/instaparse#readme
+
+(defparser block-parser
+  "(* This first rule is the top-level one. *)
+   block = ( syntax-in-block / any-char )*
+   (* `/` ordered alternation is used to, for example, try to interpret a string beginning with '[[' as a block-link before interpreting it as raw characters. *)
+   
+   <syntax-in-block> = (block-link | block-ref | hashtag)
+   
+   block-link = <'[['> any-chars <']]'>
+   
+   block-ref = <'(('> any-chars <'))'>
+   
+   hashtag = <'#'> any-chars | <'#'> <'[['> any-chars <']]'>
+   
+   (* It’s useful to extract this rule because its transform joins the individual characters everywhere it’s used. *)
+   (* However, I think in many cases a more specific rule can be used. So we will migrate away from uses of this rule. *)
+   any-chars = any-char+
+   
+   <any-char> = #'\\w|\\W'
+   ")
 
 
 (defn transform
-  "Transforms instaparse output to hiccup."
+  "Transforms Instaparse output to Hiccup."
   [tree]
   (insta/transform
-    {:S    (fn [x] [:span x])
-     :link (fn [title]
-             (let [id (subscribe [:block/uid [:node/title title]])]
-               [:span
-                [:span {:style {:color "gray"}} "[["]
-                [:a {:href  (rfee/href :page {:id (:block/uid @id)})
-                     :style {:text-decoration "none" :color "dodgerblue"}} title]
-                [:span {:style {:color "gray"}} "]]"]]))
-     :hash (fn [title]
-             (let [id (subscribe [:block/uid [:node/title title]])]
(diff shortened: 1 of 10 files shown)


test/athens/parse_helper_test.clj
```diff
@@ -0,0 +1,17 @@
...
+  (is (= []
+         (parse-helper/combine-adjacent-strings [])))
+  (is (= ["some text"]
+         (parse-helper/combine-adjacent-strings ["some" " " "text"])))
+  (is (= ["some text" [:link] "around a link"]
+         (parse-helper/combine-adjacent-strings ["some" " " "text" [:link] "around " "a link"])))
+  (is (= [{:something nil} "more text" [:link] "between elements" 39]
+         (parse-helper/combine-adjacent-strings [{:something nil} "more" " " "text" [:link] "between" " " "elements" 39])))
+  (is (= [{:a 1 :b 2} 3 ["leave" "intact"]]
+         (parse-helper/combine-adjacent-strings [{:a 1 :b 2} 3 ["leave" "intact"]]))))
```
jeroenvandijk, line 17: For a lot of repetition you can use `are`. E.g. something like this: 
```clojure
(deftest combine-adjacent-strings-tests
  (are [x y] (= x (parse-helper/combine-adjacent-strings y))
     [] []
     ["some text"] ["some" " " "text"]
     ["some text" [:link] "around a link"] ["some" " " "text" [:link] "around " "a link"]
     [{:something nil} "more text" [:link] "between elements" 39]
      [{:something nil} "more" " " "text" [:link] "between" " " "elements" 39]
     [{:a 1 :b 2} 3 ["leave" "intact"]]
      [{:a 1 :b 2} 3 ["leave" "intact"]]))
```

src/cljs/athens/parser.cljs
```diff
@@ -1,50 +1,75 @@
 (ns athens.parser
   (:require
-    [instaparse.core :as insta]
+    [athens.parse-helper :as parse-helper]
+    [instaparse.core :as insta :refer-macros [defparser]]
```
jeroenvandijk, line 4: Might be good to move `src/cljs/athens/parser.cljs ` to `src/cljs/athens/parser.cljc`. That way you can could also apply clj tests on the parser

test/athens/parse_helper_test.clj
roryokane: Done. That does make it a bit easier to compare the before and after values. I kept all inputs/outputs on their own line to make it easy to see where one ends and the other begins.

src/cljs/athens/parser.cljs
roryokane: Done, though only the `block-parser` code could be moved, not the `transform` or `parse` code. See the last commit message (to be pushed) for details on that.

roryokane:
Ready for review. If you approve the PR, give me a chance to squash one of the commits before you merge.

roryokane:
I mentioned in one of the commit messages that I didn’t see a good alternative to parsing one character at a time in some places and then combining those characters later in the `transform` section. Sadly this makes the parse results not that useful on their own, as you can see from the [tests in `test/athens/parser_test.clj`](https://github.com/athensresearch/athens/pull/88/files#diff-8830251d7eccb717d83a04b404df4f59R7):

~~~clj
  (is (= [:block "O" "K" "?" " " "Y" "e" "s" "."] (block-parser "OK? Yes.")))
  (is (= [:block [:block-link [:any-chars "l" "i" "n" "k"]]] (block-parser "[[link]]")))
~~~

If we wanted the parse tree to be more like a generic AST that could have more things done to it than just rendering to Hiccup, we could introduce a second transform step that just combines individual characters, leaving the other transform step completely in charge of Hiccup. This would make the parsing tests easier to write too.

I didn’t do that yet for two reasons:

- It would make parsing slower: the tree would have to be walked twice instead of once.
- I’m not sure how else the parse tree might change. I can see avenues to making the sub-rules like `block-link` parse contents as a single string, though I don’t currently see a way to avoid that for the top-level `block` rule. Because of these potential changes, this hypothetical splitting of transform steps might not be in the code long enough to be worth writing.

roryokane:
Ready for merge.

> When we sort out styling and html structure, probably good to add tests for text -> hiccup result

Yeah, some tests like that will probably be useful eventually. Note that those tests will be somewhat difficult to write: they will need to mock the re-frame/posh database since `transform` uses `re-frame.core/subscribe` to embed data into the Hiccup result.

PR #90, opened 2020-05-24
Adds merge prompt

jelmerderonde:
First work on #61. The (very ugly) prompt does appear when trying to rename to a page title that already exists. I do have some questions about completing this PR:

- Are there conventions about how re-frame's app db should be structured?
- What should happen when a user navigates away from the page without responding to the prompt? Maybe the `:navigated` event should clear it?
- Should the transaction to do the actual merge be part of this PR?

### DIFF
--- a/src/cljs/athens/events.cljs
+++ b/src/cljs/athens/events.cljs
@@ -40,7 +40,7 @@
         :start (swap! timers assoc id (js/setTimeout #(dispatch event) wait))
         :clear (do (js/clearTimeout (get @timers id))
                    (swap! timers dissoc id))))))
-                   
+
 
 (reg-event-fx
   :get-datoms
@@ -109,7 +109,11 @@
                (if (node-with-title ds new-title)
                  {:db (assoc db :merge-prompt {:active true
                                                :old-title old-title
-                                               :new-title new-title})}
+                                               :new-title new-title})
+                  :timeout {:action :start
+                            :id :merge-prompt
+                            :wait 7000
+                            :event [:node/merge-canceled]}}
                  {:transact (rename-tx ds old-title new-title)}))))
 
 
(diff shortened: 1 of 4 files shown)


jelmerderonde:
I think this is ready now. I might have messed up the git history though... Haven't figured out the correct flow for rebasing yet.

All feedback is welcome, including on how to fix the git history!

jeroenvandijk:
@jelmerderonde We squash all the commits before the merge (1 commit is the end result) so you don't have to care too much about the history if the diff is good.

However in this particular case the diff is really big (21 changed files) and you would revert things that you didn't want to touch. So you probably want to revert to the current state of master and re-apply your specific changes (I found this [gist](https://gist.github.com/riccardoscalco/129e5b630b7b00522d89)). If all fails, creating a new PR is a good option :)

jelmerderonde:
@jeroenvandijk I think I managed to find the correct git incantations to fix it.

src/cljs/athens/events.cljs
```diff
@@ -82,14 +92,78 @@
...
+         [?e :block/children ?children]]
+       ds title))
+
+
+(defn move-blocks-tx
+  [ds from-title to-title]
+  (let [block-count (count-children ds to-title)
+        block-eids (get-children-eids ds from-title)]
+    (mapcat (fn [eid]
+              (let [order (:block/order (d/pull ds [:block/order] eid))]
```
jeroenvandijk, line 143: I think you could also write this as `(:block/order (d/entity ds eid))`

src/cljs/athens/page.cljs
```diff
@@ -100,11 +100,27 @@
          title]))))
 
 
+(defn merge-prompt
+  [{:keys [old-title new-title]}]
+  [:div {:style {:background "red"
+                 :color "white"}}
```
jeroenvandijk, line 106: Eventually we want all these styles in https://github.com/athensresearch/athens/blob/master/src/cljs/athens/style.cljs

PR #91, opened 2020-05-25
sci boxes

tomisme:
An experiment in connecting mini SCI environments

Let's say you could put executable code in Athens' blocks.

Some questions:
 - In what order do we evaluate our blocks?
 - How do we pass data in and out of our blocks?
 - How do we handle async code?

Attempted approach:
 - Blocks are passed the evaluated result of their parent (`*1`)

Some other approaches:
 - Blocks inherit the environment of their parent
 - Blocks mutate a global environment
 - Blocks are babashka pods?

Fun stuff to try:
 - Pass in the datascript connection
 - `spit`/`slurp` to IPFS etc.


tomisme:
image

tomisme:
Totally understand if folks aren't sold on devcards, or think experiments like this should be in a separate project / codebase :smile:

src/cljs/athens_devcards/db.cljs
```diff
@@ -0,0 +1,7 @@
+(ns athens-devcards.db
```
jeroenvandijk, line 1: I think we can do `athens.devcards.db`?

src/cljs/athens_devcards/sci_boxes.cljs
```diff
@@ -0,0 +1,308 @@
+(ns athens-devcards.sci-boxes
```
jeroenvandijk, line 1: `athens-devcards.sci-boxes` -> `athens.devcards.sci-boxes`

src/cljs/athens_devcards/db.cljs
tomisme: Yeah sounds good :)

src/cljs/athens/devcards/sci_boxes.cljs
```diff
@@ -0,0 +1,310 @@
+(ns athens.devcards.sci-boxes
+  (:require
+    [cljsjs.react]
+    [cljsjs.react.dom]
```
tomisme, line 4: I'm not sure why these react imports are necessary but I get "React is not defined" when they're not there

tomisme:
> Really cool @tomisme! Given your questions in the PR, would you like to merge or are you still working on this?

Happy to merge, I won't be able to do any more work on it for a little while. That way others can play around with it :)

Should I squash my commits first?

tangjeff0:
It can be done on GH @tomisme. Just share the final commit here and @jeroenvandijk or I will copy it in. :)

tomisme:
> It can be done on GH @tomisme. Just share the final commit here and @jeroenvandijk or I will copy it in. :)
> 
> Or we can squash directly!

Excellent. Yeah I think that's the last of it, ready when you are :)

tangjeff0:
I'll let @jeroenvandijk finish the PR because he reviewed already and he's been working with sci more closely. To respond to your comment @tomisme , though, the reason you need to require 
```
[cljsjs.react]
[cljsjs.react.dom]
```
is because React needs to be shimmed when using `shadow-cljs` @tomisme. See https://github.com/bhauman/devcards/issues/156

tangjeff0:
`http://localhost:3001/cards.html` doesn't work for me. I have to open the file from my filesystem, e.g. `file:///athens/public/cards.html#!/athens.devcards.block`

Nothing even runs on `localhost:3001`, my build runs on `localhost:3000` still. Thoughts? @jeroenvandijk @tomisme

tomisme:
Anything at localhost:3000/cards.html?

avichalp:
For me, It is available on `localhost:3000`.  

@tangjeff0: You can also start a separate server using `lein devcards` then it will available on `3001` too.

tangjeff0:
Works now @tomisme @avichalp ! Previously I was getting the default _All Pages_ view before at `localhost:3000/cards.html` but good now!

PR #93, opened 2020-05-27
proof of concept for nesting of blocks

jeroenvandijk:
### Goals:
- make block functionality more performant
- Give example of how to make file structure of project more modular
   * Easier to work in parallel (no merge conflicts, less coupling)
   * Easier to understand
- decouple functions from database and framework (less datascript, less re-frame and subscriptions). 
   * Easier to understand, demo, experiment, iterate, integrate
- isolate features / components via devcards

TODO

- [x] Operate on atom data structure instead of db
- [ ] Operate on db (via different configuration)
- [ ] Add old styling
- [ ] Integrate back in project


jeroenvandijk:
Athens_Cards

tangjeff0:
Using `cursor`s is a great idea, but performance appears to be sufficient for now. May return to this.

Issue #95, opened 2020-05-27
Migrate from re-frame to reagent


tangjeff0:
06-07-2020 Update

- [ ] so what is the difference between posh and datascript practically? We use both interchangeably in the application

daveduthie:
@tangjeff0 is this mostly about devcards? Or are there other reasons too?

If it's the former, have you considered a dependency-injection approach to
making components devcards-friendly? This is a pattern we've used quite
succesfully in a component library I maintain. For example:

```
[some-comp {:model        *some-document
            :subscribe-fn could-be-reframe-subscribe
            :dispatch-fn  could-be-re-frame-dispatch
            :more         :args}
```

The component takes a `:model`, which could be a reagent `atom` or a re-frame
`subscription`. Components read upstream changes via `model` and send changes
downstream via `dispatch-fn` or, if that's not present, fall back to `swap!`-ing
on `model`. Richer interactions with the environment all go through
`subscribe-fn` and `dispatch-fn`.

tangjeff0:
@daveduthie part of this is definitely DevCards, and more generally, component isolation. I wasn't aware of dependency-injection in this case, but that seems like a very interesting path! Curious to hear others' opinions @jeroenvandijk @tomisme .

jeroenvandijk:
The reason I don't think re-frame is suitable is because of the lack of being able to do dependency injection properly. If we could pass the `app-db` as an argument to `subscribe` etc this would be a whole lot easier.

I think the proposal of @daveduthie, with all respect, is a start but not as good as I would like to have it. I'm used to making a whole application deterministic by having deterministic components in tests (mocking time, randomness etc). This what I would like to see with Athens as well.

tangjeff0:
As long as there is a need for global state or actions, there are still use cases for `re-frame`. In particular, routing should still be handled with re-frame.

But for datascript, we don't need to use re-posh by any means. Querying or transacting to datascript is already global. Creating re-frame subs or events are not totally useful, AFAICT. Query API is just Datalog, and Pull API is just maps and vectors. We can compose and reuse these independent from re-frame patterns.

avichalp:
> As long as there is a need for global state or actions, there are still use cases for re-frame. In particular, routing should still be handled with re-frame.

This is a good point. 

[...]

PR #101, opened 2020-05-29
chore: migrate to posh from re-posh/re-frame

tangjeff0:
- Don't know how we want to handle global subscriptions or dispatches (i.e. routing) now that we are in reagent
- Will have to have some other boot loading flow for when app or DevCards starts (and we need to pull in an example data base).


tangjeff0:
This was just a demo for how to create a component (left-sidebar) without re-frame. Continue architectural discussion in #95

PR #111, opened 2020-05-31
Design system

tangjeff0:
Creating a draft PR to continue discussion with @jeroenvandijk  on [composing styles](https://github.com/athensresearch/athens/commit/4d0b5454c8aaba5df097e6019df5204fce88b9d3#r39553790).

Basically, I think the interface to `with-style` should be very flexible and intuitive. Accept functions, maps, or nested maps. Accept any arity. 

```
;; All cases should evaluate to the the following vanilla cljs
[:div {:style {:display "flex"}}]

;; map
[:div (with-style {:display "flex"})]

;; nested map
[:div (with-style {:style {:display "flex"}})]

;; using functions
(def +flex (with-style {:display "flex"}))
[:div (+flex)]
[:div (with-style +flex)]
[:div (with-style (+flex))]


;; Now let's try composing
(def +justify-center (with-style {:justify-content "center"}))
(def +align-center (with-style {:align-center "center"}))
(def +flex-center (with-style +flex +justify-center +align-center))

[:div (with-style +flex +justify-center +align-center)]
[:div (with-style (+flex) (+justify-center) (+align-center))]
[:div (+flex-center)]


;; Mix and Match
;; 1 - function
;; 2 - called function
;; 3 - map
;; 4 - nested map
[:div (with-style +flex (+align-center) {:justify-content "center"} {:style {:text-align "center"}})]

;; Output
[:div {:style {:display "flex" :align-items "center" :justify-content "center" :text-align "center"}}]
```

### DIFF
--- a/src/cljc/athens/lib/dom/attributes.cljc
+++ b/src/cljc/athens/lib/dom/attributes.cljc
@@ -1,0 +1,88 @@
+(ns athens.lib.dom.attributes
+  (:require
+    [clojure.string]))
+
+
+(defn merge-dom-classes
+  [attrs dom-classes]
+  (let [class-str (if (string? dom-classes)
+                    dom-classes
+                    (clojure.string/join " " dom-classes))]
+    (update attrs :class (fn [s]
+                           (if s
+                             (str s " " class-str)
+                             class-str)))))
+
+
+(defn with-classes
+  [& dom-classes]
+  (fn f
+    ([] (f nil))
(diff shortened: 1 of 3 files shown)


tangjeff0:
Oh, because `(with-style)` always returns a function, if we want to use it inline, we will have to call it twice, e.g. `[:div ((with-style {:display "flex"}))]`

PR #117, opened 2020-06-03
feat: clearer browser tab titles

roryokane:
### Before
- Page with title: “Athens vs Roam Tech Stack”
- Page without title: “Athens Research”

### After
- Page with title: “Athens vs Roam Tech Stack – Athens”
- Page without title: “untitled – Athens”

### DIFF
--- a/src/cljs/athens/router.cljs
+++ b/src/cljs/athens/router.cljs
@@ -26,8 +26,10 @@
   (fn [{:keys [db]} [_ new-match]]
     (let [old-match   (:current-route db)
           controllers (rfc/apply-controllers (:controllers old-match) new-match)
-          node (subscribe [:node [:block/uid (-> new-match :path-params :id)]])] ;; TODO make the page title query work when zoomed in on a block
-      (set! (.-title js/document) (or (:node/title @node) "Athens Research")) ;; TODO make this side effect explicit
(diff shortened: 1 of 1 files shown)


src/cljs/athens/router.cljs
```diff
@@ -26,8 +26,10 @@
   (fn [{:keys [db]} [_ new-match]]
     (let [old-match   (:current-route db)
           controllers (rfc/apply-controllers (:controllers old-match) new-match)
-          node (subscribe [:node [:block/uid (-> new-match :path-params :id)]])] ;; TODO make the page title query work when zoomed in on a block
-      (set! (.-title js/document) (or (:node/title @node) "Athens Research")) ;; TODO make this side effect explicit
+          node (subscribe [:node [:block/uid (-> new-match :path-params :id)]]) ;; TODO make the page title query work when zoomed in on a block
+          node-title (:node/title @node)
```
roryokane, line 30: In this code `node-title` may be `nil`, which is why the next line has `(or node-title "untitled")`. Is there any Clojure naming convention that would let me suggest that `node-title` may be nil, like `maybe-node-title`, or is this name fine?

I know I could also merge the `node-title` line into `page-title` so there is never a possibly-nil variable, like this:

~~~clj
page-title (str (or (:node/title @node) "untitled") " – Athens")]
~~~

But I think that makes `page-title` too hard to understand at a glance.

PR #127, opened 2020-06-05
feat(parsing): parse formatting syntax within url-link-text

roryokane:
Allow writing bold-formatted text in a link with `**foo**`, or writing a literal `]` with `\]`.

As the parsing of more formatting syntax such as `~~strikethrough~~` is implemented, this parse rule might need to be updated to accept that syntax as well. Only some of the new syntax will be allowed in link text.

The backslash-escaping was taken from Markdown. Roam doesn’t support that syntax (yet?), but I think Athens should.

[Add `?w=1` to the GitHub diff URL](https://github.com/athensresearch/athens/pull/127/files?w=1) to hide the whitespace-only changes in the second hunk of `src/cljc/athens/parser.cljc`.

### DIFF
--- a/src/cljc/athens/parser.cljc
+++ b/src/cljc/athens/parser.cljc
@@ -25,7 +25,9 @@
    <hashtag-delimited> = <'#'> <'[['> #'[^\\]]+' <']]'>
    
    url-link = url-link-text url-link-url
-   <url-link-text> = <'['> any-chars <']'>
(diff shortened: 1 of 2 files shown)


roryokane:
This PR has test failures only because commit 2d79333c28dc865336e594ea97d75805d880efd4 broke the build. None of the failures are related to this PR.

roryokane:
I rebased on the latest `master` branch so that the tests would pass.

PR #142, opened 2020-06-11
refactor: use stylefy for component styling

shanberg:
Buttons, Athena, Left Sidebar, and All Pages tables refactored to use Stylefy


tangjeff0:
Please push to [GH Pages](https://github.com/athensresearch/athens/blob/master/CONTRIBUTING.md#deploying-athens-and-devcards). I remember you had an error about an argument list?

shanberg:
@tangjeff0 [Published](https://shanberg.github.io/athens/)

Haven't figured out the core issue, but I did get it published.

tangjeff0:
MIght've had something to do with the fact that you cloned and didn't fork. So `remote/origin/gh-pages` wasn't cloned as well or something. 

Will take a look tonight @shanberg

src/cljs/athens/devcards/all_pages.cljs
```diff
@@ -51,37 +55,82 @@
...
+   ::stylefy/manual [[:tbody {:vertical-align "top"}
+                      [:tr
+                       [:td {:border-top (str "1px solid " (:panel-color COLORS))}]
+                       [:&:hover {:background-color table-cell-background-color-hover
+                                  :border-radius "8px"}
+                        ;; [:td {:border-top-color "transparent"}]
+                        [:td [(selectors/& (selectors/first-child)) {:border-radius "8px 0 0 8px"
+                                                                     :box-shadow "-16px 0 hsla(30, 11.11%, 93%, 0.1)"}]]
+                        [:td [(selectors/& (selectors/last-child)) {:border-radius "0 8px 8px 0"
+                                                                    :box-shadow "16px 0 hsla(30, 11.11%, 93%, 0.1)"}]]]]]
```
tangjeff0, line 98: What do these do?

src/cljs/athens/devcards/all_pages.cljs
```diff
@@ -51,37 +55,82 @@
     (.toLocaleString  (js/Date. x))))
 
 
+(def table-cell-background-color-hover
+  (opacify (:panel-color HSL-COLORS) (first OPACITIES)))
```
tangjeff0, line 59: I'd like to come up with a cleaner way to do colors and opacities down the line. Ideally, we don't need to import too many constants, and can use more `+style` functions.

src/cljs/athens/devcards/athena.cljs
```diff
@@ -47,21 +50,125 @@
...
+                                   :display "-webkit-box"
+                                   :-webkit-line-clamp "1"
+                                   :-webkit-box-orient "vertical"
+                                   :color (opacify (:body-text-color COLORS) (nth OPACITIES 3))}
+                         :link-leader {:grid-area "icon"
+                                       :color "transparent"
+                                       :margin "auto auto"}}
+   ::stylefy/mode {:hover {:background (:link-color HSL-COLORS)
+                           :color (:app-bg-color COLORS)}}
+   ::stylefy/manual [[:&:hover [:.title :.preview :.link-leader {:color "inherit !important"}]]]})
```
tangjeff0, line 142: Why do we need to use `!important`?

src/cljs/athens/devcards/athena.cljs
```diff
@@ -178,28 +246,27 @@
...
-                   [:div (with-attributes {:class "athena-result" :key i :on-click #(navigate-page block-uid)})
-                    [:div
-                     [:h4 (highlight-match query page-title)]
-                     (when block-string
-                       [:span (highlight-match query block-string)])]
-                    [:h4 (with-styles {:margin-left "auto"}) [:> mui-icons/ChevronRight]]]))])))]])))
+                   [:div (use-style result-style {:key i :on-click #(navigate-page block-uid)})
+                    [:h4.title (use-sub-style result-style :title) (highlight-match query page-title)]
+                    (when block-string
+                      [:span.preview (use-sub-style result-style :preview) (highlight-match query block-string)])
```
tangjeff0, line 268: The UX of search results is a rabbithole, so these are just comments for the future. You've probably already thought of all this.

- I'd personally want orange highlight, not just heavier font weight, for more contrast.
- Sometimes the block-string is matched pretty far within the block, so we don't see the closest context. We only see the start of the block and the title of the page it's in.

src/cljs/athens/devcards/athena.cljs
```diff
@@ -178,28 +246,27 @@
...
+                    [:span.link-leader (use-sub-style result-style :link-leader) "->"]]))])))]])))
```
tangjeff0, line 269: No icon? 😢

src/cljs/athens/devcards/buttons.cljs
```diff
@@ -1,23 +1,115 @@
...
+                    [:span "Button"]
+                    [:> mui-icons/ChevronRight]]}]
+   [button {:disabled true :label "Button"}]
+   [button {:disabled true :label [:> mui-icons/Face]}]
+   [button {:disabled true :label [:<>
+                                   [:> mui-icons/Face]
+                                   [:span "Button"]]}]
+   [button {:disabled true :label [:<>
+                                   [:span "Button"]
+                                   [:> mui-icons/ChevronRight]]}]])
```
tangjeff0, line 93: THESE BUTTONS!!!!!!!!!!!!!!

src/cljs/athens/devcards/left_sidebar.cljs
```diff
@@ -105,9 +122,11 @@
...
-                        {:disabled true}) [:> mui-icons/TextFormat]]
-             [:button {:disabled true} [:> mui-icons/Settings]]]]])))))
+             [button {:disabled true
+                      :label [:> mui-icons/TextFormat (with-styles {:font-size "16px"})]
+                      :style {:margin-right "8px"}}]
+             [button {:disabled true
+                      :label [:> mui-icons/Settings (with-styles {:font-size "16px"})]}]]]])))))
 
 
 (defcard-rg Comments
```
tangjeff0, line 132: - Can get rid of these `Comments`.
- extraneous `div` on line 141?

src/cljs/athens/devcards/left_sidebar.cljs
```diff
@@ -43,13 +44,18 @@
...
   (with-styles +left-sidebar
-    {:flex "0 0 32px"
-     :padding "32px 0px"
+    {:flex "0 0 40px"
+     :align-items "flex-start"
+     :justify-content "flex-start"
+     :grid-template-rows "min-content"
+     :padding "32px 4px"
+     :display "grid"
+     :grid-gap "4px"
```
tangjeff0, line 58: I know it looks good in the app, but do we want to define height for collapsed sidebar? We might modify it down the line, in which full height would be nice

tangjeff0:
A bug:

Screen Shot 2020-06-11 at 5 14 36 PM

Hard to reproduce deterministically. I didn't screenshot the other errors, but I believe one of them said that we needed to wrap either a `for` loop or a `map` in a `doall`. This is because both of these return lazy sequences, so they don't get mounted to the virtual DOM, which breaks the app.

I'm not sure if it's 118 or 131 in `table.cljs`, or both. See line 16 of `page.cljs` for an example @shanberg

tangjeff0:
Also, can we replace `with-styles` and `with-attributes` with the stylefy helpers?

src/cljs/athens/devcards/all_pages.cljs
shanberg: @tangjeff0 These add a little bit of extra background to the hovered items. Tables don't allow for this kind of thing without either crazy-bad markup or a little bit of CSS trickery.

Reference (the left edge of the table ends where the border does):
Screen Shot 2020-06-11 at 9 09 05 PM

src/cljs/athens/devcards/athena.cljs
shanberg: @tangjeff0 We don't, that shouldn't have stayed in there. I'll make a cleanup PR for this little stuff.

src/cljs/athens/devcards/athena.cljs
shanberg: @tangjeff0 agreed.
- The highlight was supposed to be black according to the mockups, so a little bit darker and clearer, but I wasn't able to get them to change on row hover until just now. That's fixed in the cleanup PR. Open to discussing more contrast.
- Until we're able to trim the string, this is a tradeoff between sometimes clipping out the term and sometimes just seeing enormous block previews. Clipping to 1 line is probably overzealous, so I'll show the whole result until a decision is made or the trimming feature is added.

src/cljs/athens/devcards/athena.cljs
shanberg: @tangjeff0 Help wanted. I got a crash I couldn't solve when I tried.

src/cljs/athens/devcards/left_sidebar.cljs
shanberg: Fixed in #144

PR #143, opened 2020-06-12
Add Dockerfile for quick and easy evaluation

nightscape:
Complete Clojure noob here :wink: 
Adding a Docker build allows people to quickly get started with their own instance of Athens.
In a further step one could also consider adding an automated build on Docker Hub so that one can run the newest version by a simple
```
docker run athensresearch/athens:latest
```
The build sequence in the Dockerfile might not be optimal.
I made the assumption that JS dependencies will change less often and moved the `yarn` part above the `lein` parts (so that the cache can be reused between builds if the JS dependencies didn't change).
I also introduced a `compile` step in `project.clj` that is run in the Docker build to reduce compile time when the container actually starts. For whatever reason it still recompiles ~150 files when running the container.

Let me know what you think about it!


tangjeff0:
Awesome first PR @nightscape. I love Docker but I'm not sure it's completely necessary. There are two contexts where Docker might be used:

- development
- production

If someone is developing Athens, they will ultimately be running scripts like [linting and testing](https://github.com/athensresearch/athens/blob/master/CONTRIBUTING.md#running-ci-scripts-locally). In this case, they would need to SSH into the container to run these scripts. Is Docker that much easier than setting up a dev environment locally? Especially if they are learning Clojure, they will probably already have these dependencies locally.

For production, all we have is a frontend bundle of HTML and JS. This could be easily downloaded and opened. Furthermore, it won't be long before we have a Desktop app that is even easier to run.

If we do have a self-hosted backend, which won't be for a long while, the nice thing about Clojure is that it can be distributed as a Jar file — it's already cross-platform.

We don't have many system dependencies, we aren't leveraging cloud VMs or containers, and this is a web app. I love Docker, but for these reasons I'm not sure we need Docker now.

What are your thoughts here @nightscape ?

nightscape:
I would say it's useful for people like me who are new to Clojure development and want to quickly try out Athens without finding out how to properly install lein and yarn in their machine (which, depending on your distribution might be more or less straightforward).
For people who have a Kubernetes cluster it's also the best option to deploy Athens there, although for a real deployment the Dockerfile should be tweaked to create a smaller image (which can be done as a second step).

nightscape:
Regarding linting and testing: you don't need to SSH into your container (in fact, it's an antipattern to have an SSH server running in a Docker container).
Instead, you can
```
docker run -it athens lein do-linting-stuff-and-run-testy
```
to run lint or tests in a new container instance or
```
docker exec -it running-athens-container bash
```
to log into a running container and run everything from there.

tangjeff0:
Awesome, forgot how easy it was to execute commands in a docker process!

Next steps: would you like to add a section to CONTRIBUTING for running Athens with Docker? @nightscape

PR #154, opened 2020-06-16
refactor(blocks): don't create unnecessary spans

shanberg:
Switch to fragments for spans that don't apply styling or have semantic meaning. Makes for easier markup to navigate, and less work for the browser to do.

[Demo](https://shanberg.github.io/athens/cards.html#!/athens.devcards.blocks)

### DIFF
--- a/src/cljs/athens/devcards/blocks.cljs
+++ b/src/cljs/athens/devcards/blocks.cljs
@@ -161,7 +161,7 @@
          [:> mui-icons/KeyboardArrowDown {:style {:font-size "16px"}}]]
         [:span (use-style block-disclosure-toggle-style)])
       [:a (use-style block-indicator-style {:class (if closed? "closed" "open") :on-click #(navigate-page uid)})]
-      [:span (parse-and-render string)]]
+      [:<> (parse-and-render string)]]
      (when open?
        (for [child (:block/children block)]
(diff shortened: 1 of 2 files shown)


src/cljs/athens/devcards/blocks.cljs
```diff
@@ -161,7 +161,7 @@
          [:> mui-icons/KeyboardArrowDown {:style {:font-size "16px"}}]]
         [:span (use-style block-disclosure-toggle-style)])
       [:a (use-style block-indicator-style {:class (if closed? "closed" "open") :on-click #(navigate-page uid)})]
-      [:span (parse-and-render string)]]
+      [:<> (parse-and-render string)]]
```
tangjeff0, line 164: I think here we should just write `[parse-and-render string]`

https://github.com/reagent-project/reagent/blob/master/doc/UsingSquareBracketsInsteadOfParens.md

src/cljs/athens/parse_renderer.cljs
```diff
@@ -57,5 +57,5 @@
        {:title (pr-str (insta/get-failure result))
         :style {:color "red"}}
        string]
-      [:span
+      [:<>
```
tangjeff0, line 60: see preivous

PR #155, opened 2020-06-17
feat(pages): nicer title editing

shanberg:
A little hesitant to put this one up. [Demo](https://shanberg.github.io/athens/#/page/AVOHMK6bK)

- Simpler page title editing UX


tangjeff0:
Is it cool if we hold off on future edits to `page.cljs`? Gonna be importing the newer page and blocks into the app soon

PR #158, opened 2020-06-19
feat(blocks): first block-writing features (WIP) using event listeners

tangjeff0:
- fix block toggle for block- and node-pages
- dragging bullet point. positioning of bullet needs offset
- click to edit textarea
- real datascript transactions


tangjeff0:
merging in per #161

Issue #100, opened 2020-05-27
All Pages Table


tangjeff0:
# Downhill Tasks (good for contributors)

- [x] datetime formatter for `Modified` and `Created` columns. Library discussion: https://github.com/athensresearch/athens/issues/66#issuecomment-642039741
- [x] responsive widths. see **Collapsed** section on #99

# UI

**Don't worry about showing links within a page.**

![table](https://user-images.githubusercontent.com/8952138/83808125-40b84980-a682-11ea-9d5a-28db364703f7.png)

tangjeff0:
![Screen Shot 2020-06-02 at 6 23 34 PM](https://user-images.githubusercontent.com/8952138/83575727-2d7c7100-a4fe-11ea-983f-a676372ceee8.png)

- No breakpoints / no responsive: doesn't work at all with mobile view.

estellerostan:
I'm shafoo on Discord and I'd like to work on the task "datetime formatter for Modified and Created columns" as my ClojureFam goal to merge my first Pull Request for Athens.

One question, the linked date library is `cljs-time` while `juxt/tick` was chosen for issue #66, is it okay to work with multiple libraries in the same project or is it better to switch to tick here too?

tangjeff0:
@estellerostan I just updated my link. We should stick to one library, but not sure the state of #66. @Bardia95 ?

Bardia95:
@estellerostan I'm not sure what you mean by `cljs-time` being the linked date library, but I think its best if we stuck with just one library. Are there any benefits you see in using `cljs-time` over `juxt/tick`?

tangjeff0:
@Bardia95 I previously linked to cljs-time but have since updated it. @estellerostan I think it's safe to go ahead and try to implement this with `juxt/tick`

tangjeff0:
Can always add more columns, sorting, bulk delete/export, etc. to the table, but I think this is a really good start for v1. Closing!

PR #169, opened 2020-06-21
All pages date format

estellerostan:
This is my first PR to improve datetime format for `Created` and `Modified` columns, based on the UI showed in #100.
It uses `juxt/tick`, the chosen date library as discussed in the issue.

Some choices I made:
- The am/pm marker is simply put to lower case but this [SO answer](https://stackoverflow.com/questions/13581608/displaying-am-and-pm-in-lower-case-after-date-formatting/13581910#13581910), although not the accepted best answer, says that it is not the correct way to do it in Java with the old `SimpleDateFormat`. I went with it anyway because:
1. Athens runs in the browser so the date input is always AM/PM and thus it doesn't have the problem mentioned in the following SO comment:
> [...] some phones display AM/PM, others display a.m./p.m. 
2. I couldn't find the equivalent of `DateFormatSymbols` for Java 8 `DateTimeFormatter`

- Localization: the previous code had a JS call to Date's [`toLocaleString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) which according to the docs says that the default locale and the default time zone are en-US locale with time zone America/Los_Angeles. So localization was not implemented for Athens. Can someone correct me if I'm wrong? And it is not in the specs per se but is localization needed?

I am a Clojure beginner here to learn so comments are more than welcome!

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,9 @@
     "shadow-cljs": "2.8.83"
   },
   "dependencies": {
+    "@js-joda/core": "1.12.0",
+    "@js-joda/locale_en-us": "3.1.1",
+    "@js-joda/timezone": "2.2.0",
     "@material-ui/core": "^4.10.1",
     "@material-ui/icons": "^4.9.1",
     "create-react-class": "^15.6.3",
--- a/project.clj
+++ b/project.clj
@@ -28,7 +28,8 @@
                  [devcards "0.2.6"]
                  [borkdude/sci "0.0.13-alpha.22"]
                  [garden "1.3.10"]
-                 [stylefy "2.2.0"]]
+                 [stylefy "2.2.0"]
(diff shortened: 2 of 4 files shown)


tangjeff0:
Hi @estellerostan , wonderful job! I'm working on setting up automated deploys this week. In the meantime, could you try deploying your fork on Github pages so we could see it live? Instructions are here: https://github.com/athensresearch/athens/blob/master/CONTRIBUTING.md#deploying-athens-and-devcards

estellerostan:
Forgot to mention it, already deployed and available at https://estellerostan.github.io/athens/ !

tangjeff0:
Great job @estellerostan  :)

nthd3gr33:
Way to go @estellerostan! Congratulations 🎉

PR #177, opened 2020-06-25
docs: Add badges to README

nthd3gr33:
TODO:
- [x] Decide if we should have badges or not
- [x] Decide on badges that should be removed
- [x] Decide if other badges should be added (source: https://shields.io/)
- [ ] implement changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,21 @@ To learn more about this project, please see:
 - [Governance](GOVERNANCE.md) — BD + Core Team + Guardians + Athenians
 - [Code of Conduct](CODE_OF_CONDUCT.md) — our values and guidelines, AKA how to be an awesome Athenian
 
+<p align="center">
+  <a href="https://twitter.com/athensresearch">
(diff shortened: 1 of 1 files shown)


itsrainingmani:
I think we should definitely atleast have badges for the build status of the latest commit to `master`. That way anyone viewing the repository can see that information without having to look for it.

alaq:
I like it a lot @nthd3gr33! 

Is it the right place for them though? Usually they're right under the title (see [this random repo](https://github.com/reduxjs/redux-toolkit) for example). Is there a reason we don't have a title header on the README? It's especially useful with that many files at the root level.

tangjeff0:
+1 for `build` passing badge. +1 for title and better location.

The only badges I really want are the build badge, Twitter badge with # of followers, and Discord badge if there is one. Also, please use markdown formatting where possible for the links, like the current Twitter and Discord SVGs

I'll rename our main CI script back to `build`.

nthd3gr33:
Made the requested changes. Let me know if anything else should be done.

PR #202, opened 2020-07-02
Dropdown devcard

shanberg:
[Demo](https://shanberg.github.io/athens/cards.html#!/athens.devcards.dropdown). Adds new Dropdown component.

Screen Shot 2020-07-01 at 9 54 59 PM

### DIFF
--- a/src/cljs/athens/devcards.cljs
+++ b/src/cljs/athens/devcards.cljs
@@ -10,6 +10,7 @@
     [athens.devcards.db]
     [athens.devcards.db-boxes]
     [athens.devcards.devtool]
+    [athens.devcards.dropdown]
     [athens.devcards.filters]
     [athens.devcards.icons]
     [athens.devcards.left-sidebar]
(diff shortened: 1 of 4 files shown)


tangjeff0:
Can we use real examples instead of abstract names? This helps me imagine the components in practice.

This covers separators and items (but not sub-items!)
Page Menu

Slash commands would cover input case. It's okay if the options are arbitrary since we don't know which we actually want to support.

Still need a header use case..?

tangjeff0:
Also not sure how I feel about the `textinput` name. Also, the options for `textinput` should be kebab-case, not camelCase, if I'm not mistaken

shanberg:
@tangjeff0 Yeah, I'm working on realistic examples now. The header is for Filters, as the "FILTER" heading shouldn't strictly be present in all uses of the Filter control.

Some refactoring is needed to allow modifying the dropdown style for the case of the Filters.

tangjeff0:
hawt

PR #211, opened 2020-07-03
feat(pages): style linked references

shanberg:
[Demo](https://shanberg.github.io/athens/#/page/qPMcfiUhV)

Styled linked and unlinked references in pages.

- Linked/Unlinked section appears only if it has items to show
- Breadcrumbs appear if the reference has more than one parent

Screen Shot 2020-07-03 at 11 03 35 AM


src/cljs/athens/devcards/node_page.cljs
```diff
@@ -149,30 +196,28 @@
...
-              ;; TODO: replace with breadcrumbs?
+     (when (> (count refs) 0)
+       [:section (use-style references-style {:key linked-or-unlinked})
+        [:h4 (use-style references-heading-style)
+         [(r/adapt-react-class mui-icons/Link)]
+         [:span linked-or-unlinked]
+         [button {:label    [(r/adapt-react-class mui-icons/FilterList)]
+                  :disabled true}]]
+        [:div (use-style references-list-style)
+         (doall
```
tangjeff0, line 207: Originally my code, but `doall` should wrap the top-level `for`

src/cljs/athens/devcards/node_page.cljs
```diff
@@ -149,30 +196,28 @@
...
-        [button {:label    [(r/adapt-react-class mui-icons/FilterList)]
-                 :disabled true}]]]
-      (doall
-        (for [[group-title group] refs]
-          [:<> {:key group-title}
-           [:h4 group-title]
-           (for [{:block/keys [uid parents] :as block} group]
-             [:div {:key uid}
-              ;; TODO: replace with breadcrumbs?
+     (when (> (count refs) 0)
```
tangjeff0, line 199: Alternatives

- `seq`
- `not-empty`
- `(pos? (count refs))`

I like 2.

src/cljs/athens/devcards/node_page.cljs
```diff
@@ -149,30 +196,28 @@
...
+         [(r/adapt-react-class mui-icons/Link)]
+         [:span linked-or-unlinked]
+         [button {:label    [(r/adapt-react-class mui-icons/FilterList)]
+                  :disabled true}]]
+        [:div (use-style references-list-style)
+         (doall
+           (for [[group-title group] refs]
+             [:div (use-style references-group-style {:key group-title})
+              [:h4 (use-style references-group-title-style)
+               [:a {:on-click #(navigate-uid uid)} group-title]]
```
tangjeff0, line 211: Roam behavior is to expand the reference without navigation. May implement this down the line.

src/cljs/athens/devcards/breadcrumbs.cljs
```diff
@@ -49,42 +54,88 @@
...
+  [:li (use-style (merge breadcrumb-style style) {:title label})
+   [:a {:on-click on-click} label]])
 
 
 ;;; Devcards
 
 
 (defcard-rg Normal-Breadcrumb
-  [breadcrumbs-list
+  [breadcrumbs-list {}
```
tangjeff0, line 71: I think one way to let developer not have to pass in an empty property map is to give one- and two-arity signatures to the components.

```
(defn breadcrumbs-list
  ([& children]
   (into [:ol (use-style breadcrumbs-list-style) children]))
  ([{:keys [style]} & children]
   (into [:ol (use-style (merge breadcrumbs-list-style style)) children])))
```

A pattern I often see with these kinds of simple lists, is to pass in the list data to the main parent component. It might look like:

```
(def data ["Lorem" "Ipsum" "Athens"])
(defn component
  []
  [breadcrumbs-list (use-style {:color "blue"} {:data data})])
```

`breadcrumbs-list` is then in charge of iterating over the loop, and the API is simplified in a way. I'm not saying we should necessarily do this, just thinking about API design

tangjeff0: The way we are destructuring and then using keys is not ideal, especially in `button`. A better way might be something like

```
(defn button-primary
  [{:keys [style active class] :as props} & children]
  [:button (use-style
             (merge buttons-primary-style style)
             (merge (dissoc props [:style :active :class]) {:class (concat class (when active "active"))}))
   children])
```
Don't need to implement now, but I am requesting you do one-arity functions for these fixes.

src/cljs/athens/devcards/node_page.cljs
shanberg: @tangjeff0 What does this mean? Does it add another reference item?

tangjeff0: @shanberg https://www.loom.com/share/e396caf6ccfa47de8400f067baf6f3cb

PR #217, opened 2020-07-05
Fix,refactor(all-pages): `date-string` fn

### LINKED ISSUE #213: Fix,refactor(all-pages): `date-string` fn
Originally written by @estellerostan in #169 

# Problem

Output of `date-string` is sometimes incorrect.

```
(date-string 1593878525897)
; => "July 07, 2020 12:2pm"
```

# To Reproduce

Open Athens with `ego` db. Take a look at some of the values.

Screen Shot 2020-07-04 at 12 31 30 PM

**Steps**

- Idiomatically:
  - create a separate `def` for the `date-col-format`
  - refactor the current function to use `as->` thread macro
- Practically:
  - How is the `(< ts 1)` supposed to work? `ts` can either by a timestamp integer or `nil`, if I'm not mistaken. I'm not sure this is the best predicate.

```
(def date-col-format (t/formatter "LLLL MM, yyyy h':'ma"))

(defn date-string
  [ts]
  (if (< ts 1)
    [:span "(unknown date)"]
    (as-> (js/Date. ts) x
      (t/instant x)
      (t/date-time x)
      (t/format date-col-format x)
      (str/replace x #"AM" "am")
      (str/replace x #"PM" "pm"))))
```

estellerostan:
Fix #213 

See it [here](https://estellerostan.github.io/athens)

Thanks @tangjeff0 for spotting the bug and the well described issue :+1:


estellerostan:
Thanks! I made a mistake in the commit message, what fixed the issue is not removing `(js/Date)` so I should have called 
002ee60 refactor or perf and 3638bc5 is the only fix.
- The Java [Date and Time Patterns](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html) documentation I used for the format isn't clear about the case for the bug you raised:
_Number: For formatting, the number of pattern letters is the minimum number of digits, and shorter numbers are zero-padded to this amount. For parsing, the number of pattern letters is ignored unless it's needed to separate two adjacent fields._
so I overlooked it in #169, instead I found a side note in [this SO answer](https://stackoverflow.com/a/54806516) explaining it:
_If you don’t want the leading zero, just specify one pattern letter, that is `h` instead of `hh` or `H` instead of `HH`. It will still accept two digits when parsing, and if a number to be printed is greater than 9, two digits will still be printed._
- The js/Date call is just unnecessary because tick (or Java really) already knows how to convert a timestamp to an instant, see the SO answer [Timestamp to Instant in Java](https://stackoverflow.com/a/38365494).

tangjeff0:
@estellerostan you're really good at explaining things! 

I didn't see where they said they use the Java format. Is this just assumed because tick has Java dependencies?

estellerostan:
@tangjeff0 Yes plus you can check their documentation, [tick is built on java.time](https://www.juxt.land/tick/docs/index.html#_java_8_time) and the link I posted earlier on Date and Time Patterns is from the Java 7 documentation but it still applies to [Java 8](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html).
Finally in Tick documentation [introduction](https://www.juxt.land/tick/docs/index.html#_introduction) it is said that "Tick supports both Clojure and ClojureScript, with an identical API." so I went for it and saw it worked :smile:

Issue #209, opened 2020-07-03
fix(blocks): prevent `indent` when already a leaf block, prevent `indent` and `unindent` when already a top-level child


tangjeff0:

alaq:
Can this issue be assigned to @nthd3gr33 @itsrainingmani and myself @alaq? We are working on a solution together. Thank you!

tangjeff0:
Can you comment @itsrainingmani so I can also assign this to you?

itsrainingmani:
Hi Jeff

nthd3gr33:
Found additional unexpected behavior when making the video demo:

Pressing `tab` on the top-level child:
- expected behavior: nothing
- actual behavior: destroys the block

Tested on Dynalist and Roam to double-check that the expected behavior is the same on those platforms.

Should I create a new issue or add this to this issue (my recommendation)? @itsrainingmani @alaq

tangjeff0:
Add it to this issue @nthd3gr33. I think you can edit title as well

tangjeff0:
closed by #228

PR #228, opened 2020-07-08
fix(blocks): prevent `indent` when already a leaf block, prevent `indent` and `unindent` when already a top-level child

itsrainingmani:
This PR fixes Issue #209 .

There are 3 parts to this fix - 

* Modifying the `indent` event handler to only perform an indent if the block order for the current block is not 0
* Modifying the `unindent` event handler to extract the `id` of the current page from the `:current-route` object of `rfdb`. If the `id` of the current page is equal to the parent ID of the node being unindented, prevent the unindent to keep the node within the current context.
* Call preventDefault for tab and shift+tab key down events to prevent default browser tab behavior if an indent or unindent does not happen.

### Potential Performance Optimization

We wanted to point out that there is some potential performance gain to be had by performing the `if` check at the earliest possible moment in the re-frame event handlers and thereby skipping the evaluation of possibly complex `let` bindings.

### DIFF
--- a/src/cljs/athens/events.cljs
+++ b/src/cljs/athens/events.cljs
@@ -486,11 +486,9 @@
                        db/get-block)
         new-block {:db/id (:db/id block) :block/order (count (:block/children older-sib))}
         reindex-blocks (->> (dec-after (:db/id parent) (:block/order block)))]
-    (if (zero? (:block/order block)) ; If the block is already the top in it's level, do not indent
-      {}
-      {:transact! [[:db/retract (:db/id parent) :block/children (:db/id block)]
-                   {:db/id (:db/id older-sib) :block/children [new-block]} ;; becomes child of older sibling block — same parent but order-1
-                   {:db/id (:db/id parent) :block/children reindex-blocks}]})))
+    {:transact! [[:db/retract (:db/id parent) :block/children (:db/id block)]
+                 {:db/id (:db/id older-sib) :block/children [new-block]} ;; becomes child of older sibling block — same parent but order-1
+                 {:db/id (:db/id parent) :block/children reindex-blocks}]}))
 
 
 (reg-event-fx
@@ -500,25 +498,22 @@
 
 
 (defn unindent
-  [uid context-root]
+  [uid context-root-uid]
   (let [parent (db/get-parent [:block/uid uid])
         grandpa (db/get-parent (:db/id parent))
         new-block {:block/uid uid :block/order (inc (:block/order parent))}
         reindex-grandpa (->> (inc-after (:db/id grandpa) (:block/order parent))
                              (concat [new-block]))]
-    (if (= (:block/uid parent) context-root) ; if the parent node is the context-root, prevent unindent
-      {}
-      (when (and parent grandpa)
-        {:transact! [[:db/retract (:db/id parent) :block/children [:block/uid uid]]
-                     {:db/id (:db/id grandpa) :block/children reindex-grandpa}]}))))
+    (when-not (= (:block/uid parent) context-root-uid) ; if the parent node is the context-root, prevent unindent
+      {:transact! [[:db/retract (:db/id parent) :block/children [:block/uid uid]]
+                   {:db/id (:db/id grandpa) :block/children reindex-grandpa}]})))
 
 
 (reg-event-fx
   :unindent
-  ; Pass in the reframe db as a cofx to the :unindent event handler
   (fn [{rfdb :db} [_ uid]]
-    (let [context-root (get-in rfdb [:current-route :path-params :id])]
-      (unindent uid context-root))))
+    (let [context-root-uid (get-in rfdb [:current-route :path-params :id])]
+      (unindent uid context-root-uid))))
 
 
 (defn target-child
--- a/src/cljs/athens/events.cljs
+++ b/src/cljs/athens/events.cljs
@@ -499,21 +499,25 @@
 
 ;; TODO: no-op when user tries to unindent to a child out of current context
 (defn unindent
-  [uid]
+  [uid context-root]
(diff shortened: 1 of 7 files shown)


src/cljs/athens/events.cljs
```diff
@@ -497,23 +499,26 @@
...
         reindex-grandpa (->> (inc-after (:db/id grandpa) (:block/order parent))
                              (concat [new-block]))]
-    (when (and parent grandpa)
-      {:transact! [[:db/retract (:db/id parent) :block/children [:block/uid uid]]
-                   {:db/id (:db/id grandpa) :block/children reindex-grandpa}]})))
+    (if (= (:block/uid parent) context-root) ; if the parent node is the context-root, prevent unindent
+      {}
+      (when (and parent grandpa)
+        {:transact! [[:db/retract (:db/id parent) :block/children [:block/uid uid]]
+                     {:db/id (:db/id grandpa) :block/children reindex-grandpa}]}))))
```
tangjeff0, line 513: Flatten/refactor this nested conditional expression.

I'm not sure we even need `when (and parent grandpa)` anymore since `context-root` is a stricter check.

src/cljs/athens/events.cljs
```diff
@@ -486,9 +486,11 @@
                        db/get-block)
         new-block {:db/id (:db/id block) :block/order (count (:block/children older-sib))}
         reindex-blocks (->> (dec-after (:db/id parent) (:block/order block)))]
-    {:transact! [[:db/retract (:db/id parent) :block/children (:db/id block)]
-                 {:db/id (:db/id older-sib) :block/children [new-block]} ;; becomes child of older sibling block — same parent but order-1
-                 {:db/id (:db/id parent) :block/children reindex-blocks}]}))
+    (if (zero? (:block/order block)) ; If the block is already the top in it's level, do not indent
```
tangjeff0, line 489: Since this condition is context-less — it doesn't require any re-frame or datascript subscriptions — I would add it directly to the giant `on-key-down` handler.

src/cljs/athens/events.cljs
```diff
@@ -497,23 +499,26 @@
...
+                     {:db/id (:db/id grandpa) :block/children reindex-grandpa}]}))))
 
 
 (reg-event-fx
   :unindent
-  (fn [_ [_ uid]]
-    (unindent uid)))
+  ; Pass in the reframe db as a cofx to the :unindent event handler
+  (fn [{rfdb :db} [_ uid]]
+    (let [context-root (get-in rfdb [:current-route :path-params :id])]
```
tangjeff0, line 520: I slightly prefer `context-root-uid`.

src/cljs/athens/events.cljs
```diff
@@ -497,23 +499,26 @@
...
+      (when (and parent grandpa)
+        {:transact! [[:db/retract (:db/id parent) :block/children [:block/uid uid]]
+                     {:db/id (:db/id grandpa) :block/children reindex-grandpa}]}))))
 
 
 (reg-event-fx
   :unindent
-  (fn [_ [_ uid]]
-    (unindent uid)))
+  ; Pass in the reframe db as a cofx to the :unindent event handler
```
tangjeff0, line 518: Don't need this comment, this is a very common practice.

src/cljs/athens/events.cljs
```diff
@@ -486,9 +486,11 @@
                        db/get-block)
         new-block {:db/id (:db/id block) :block/order (count (:block/children older-sib))}
         reindex-blocks (->> (dec-after (:db/id parent) (:block/order block)))]
-    {:transact! [[:db/retract (:db/id parent) :block/children (:db/id block)]
-                 {:db/id (:db/id older-sib) :block/children [new-block]} ;; becomes child of older sibling block — same parent but order-1
-                 {:db/id (:db/id parent) :block/children reindex-blocks}]}))
+    (if (zero? (:block/order block)) ; If the block is already the top in it's level, do not indent
+      {}
```
tangjeff0, line 490: Even if it's not necessary anymore, not sure how I feel about empty `{}` vs `nil` in general for return value. Thoughts?

itsrainingmani: The `{}` was to indicate explicitly that `no-op` was taking place. I think it has some semantic clarity.

src/cljs/athens/events.cljs
itsrainingmani: Moving the condition to the `on-key-down` handler has the side effect of not reaching the `preventDefault` code - 

```clojure
(and (= key-code KeyCodes.TAB) (not block-zero?)) (do (.. e preventDefault)
                                                            (dispatch [:indent uid]))
```

src/cljs/athens/events.cljs
itsrainingmani: Yeah it doesn't seem like we do. `context-root-uid` seems to handle the case just fine.

nthd3gr33: > I'm not sure we even need when (and parent grandpa) anymore since context-root is a stricter check.

Good point Jeff 🧠

src/cljs/athens/events.cljs
nthd3gr33: I specifically read somewhere yesterday in one of our libraries' official docs that this was an acceptable practice. Maybe it was in the `re-frame` docs? I will see if I can find it... 😅

tangjeff0: Lemme know if you find it @nthd3gr33!

tangjeff0:
Just did a big refactor of blocks, can you resolve conflicts @itsrainingmani ?

itsrainingmani:
Yeah no problem

itsrainingmani:
@tangjeff0 Should be able to merge now.

itsrainingmani:
Thanks !!

PR #236, opened 2020-07-09
fix(block): fix unindent on date pages

tangjeff0:
small fix, cc @nthd3gr33 @itsrainingmani @alaq


tangjeff0:
I was wrong that we didn't need the check https://github.com/athensresearch/athens/pull/228#discussion_r451472926. Need the check for date pages!

PR #237, opened 2020-07-09
feat(buttons): add active state

shanberg:

- Buttons now have 'active' state, currently used in left sidebar to indicate the current nav item
- Refactor of button styles to connect them to theme values


src/cljs/athens/views.cljs
```diff
@@ -111,7 +111,7 @@
          (if @loading
            [initial-spinner-component]
            [:div (use-style app-wrapper-style)
-            [left-sidebar]
+            [left-sidebar route-name]
```
tangjeff0, line 114: Let `left-sidebar` subscribe to `current-route` directly, see below

src/cljs/athens/views/left_sidebar.cljs
```diff
@@ -112,7 +112,7 @@
 
 
 (defn left-sidebar
-  []
+  [route-name]
   (let [open? (subscribe [:left-sidebar/open])
         ;; current-route (subscribe [:current-route]) ;; TODO: disabled primary button if current route == navigation button
```
tangjeff0, line 117: Subscribe here instead of passing `route-name` as a parameter. Both work functionally, but left-sidebar is already making a `subscribe` call, we should maintain its independence from needing parameters passed in.

tangjeff0: See comment on line 117

src/cljs/athens/views/buttons.cljs
```diff
@@ -23,41 +38,49 @@
...
-                      [(selectors/& (selectors/not (selectors/first-child))) {:margin-inline-start "0.251em"}]
-                      [(selectors/& ((selectors/first-child (selectors/last-child)))) {:margin-inline-start "-4px"
-                                                                                       :margin-inline-end "-4px"}]]
+   :transition       "all 0.075s ease"
+   ::stylefy/manual [[:&:hover {:background (darken (color :panel-color :opacity-low) 10)}]
+                     [:&:active
+                      :&:hover:active
+                      :&.active {:color (color :body-text-color)
+                                 :background-color  (darken (color :panel-color :opacity-med) 10)}]
+                     [:&:disabled {:color (color :body-text-color 0.3)
```
tangjeff0, line 49: Do we want to set `cursor: none` if disabled?

src/cljs/athens/views/buttons.cljs
```diff
@@ -23,41 +38,49 @@
...
-                                        [:disabled {:color "rgba(0, 0, 0, 0.3)"
-                                                    :background-color "#EFEDEB"
-                                                    :cursor "default"}]]}))
+  (merge buttons-style {:color (color :link-color)
+                        :background-color (color :link-color :opacity-lower)
+                        ::stylefy/manual [[:&:hover {:background (color :link-color :opacity-low)}]
+                                          [:&:active
+                                           :&:hover:active
+                                           :&.active {:color "white"
+                                                      :background-color (color :link-color)}]
```
tangjeff0, line 67: We need to have `&:active` and `&.active` pseudo and class selectors?

shanberg: One is for when the browser sees the user is clicking, one is for toggling the 'pressed' state of the button, such as when it’s indicating the active nav location. Renamed to `.is-active` for consistency.

src/cljs/athens/views/buttons.cljs
shanberg: I don't usually see special cursors on disabled buttons, usually it’s just the default arrow, rather than the pointing finger.

src/cljs/athens/views/buttons.cljs
```diff
@@ -12,9 +12,24 @@
 ;;; Styles
 
 
+(def button-icons-style
+  {:margin-block-start "-0.0835em"
+   :margin-block-end "-0.0835em"})
```
tangjeff0, line 17: This number is too precise

PR #239, opened 2020-07-10
Fix incorrect date time format

itsrainingmani:
`date-col-format` in `all_pages.cljs` had the format - `"LLLL MM, yyyy h':'mma"` which would yield July 07, 2020 for all dates in July.

This format has been fixed to be `"LLLL dd, yyyy h':'mma"`.


tangjeff0:
Please review @estellerostan

Issue #126, opened 2020-06-04
Athena


tangjeff0:
# Tasks
- [ ] typing bug https://github.com/athensresearch/athens/issues/126#issuecomment-660161119
- [x] `up` and `down` to go through list
   - arrow keys work, but now needs the list needs to scroll
- [x] two items can be highlighted at once: one from mouse hover, one from arrow keys
- [X] `enter` - create new page or go to page
- [X] `shift-enter` - open in right sidebar
- [X] save and show recent searches #208 
- [x] [debouncing](https://www.martinklepsch.org/posts/simple-debouncing-in-clojurescript.html)
- [x] detect click outside of component to close
- [X] create keybinding event listener that opens Athena from anywhere (e.g. `cmd-k`)
- [X] UI part
- [X] search part #20

# Design
![Stuarts](https://cdn.discordapp.com/attachments/710946956754550844/718291377926176768/Screen_Shot_2020-06-04_at_10.33.17_PM.png)

prabhath6:
Is anyone working on fixing the `save and show recent searches` task from the above list? If not I would like to pick it up.
Below is the gif from my local changes.

![recent items](https://user-images.githubusercontent.com/10682370/86271086-ea133200-bb91-11ea-8ed5-8cb711719161.gif)

I have a few questions regarding the design of this feature that I would like to discuss before pushing the changes.

tangjeff0:
hi @prabhath6 thanks for this. I actually was just about to merge in some changes. I removed the cache, and added key down functionality (up, down, enter, shift-enter).  This will likely affect how you implemented `recent` searches, but I'm happy to look at your fork. You can see the code changes here: https://github.com/athensresearch/athens/compare/master...tangjeff0:athena-new-page?expand=1

prabhath6:
Hi @tangjeff0, you can find my changes below. We have modified the same functions so i expect merge conflicts.
I am not sure where to store the recent items, so for now i am saving them to the `*recent-items` atom. The content of the atom is reset when the page is refreshed.

* Current code changes: https://github.com/athensresearch/athens/compare/master...prabhath6:athena-recent-items?expand=1

tangjeff0:
Awesome code @prabhath6 ! I just merged in #201, which removes the `*cache` and merges all state into one atom. It also puts search results from pages and blocks into one vector. It was easier to `concat` all those results together at query time, but we should probably still have a separate vector for recent items. In fact, it should probably be in the re-frame db as your comment asks.

[...]

PR #247, opened 2020-07-12
refactor: recent and results into single component

prabhath6:
Extracted the list drop down builder into a function that is reused to display both reuslts and recent items.
Related to: #126

### DIFF
--- a/src/cljs/athens/views/athena.cljs
+++ b/src/cljs/athens/views/athena.cljs
@@ -196,6 +196,22 @@
       :else nil)))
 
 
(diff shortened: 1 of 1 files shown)


tangjeff0:
Hi @prabhath6, thanks for this. I think you're right that this adds more complexity by perhaps trying to force code re-use. Let me think about this longer, I may fork off of your branch.

tangjeff0:
Closing for now, thank you for your work @prabhath6, not a priority ATM.

PR #253, opened 2020-07-14
feat(blocks): styled and scrolly search menu

shanberg:
[Demo](https://shanberg.github.io/athens)

- Style tweaks to block search dropdown menu
- Block search dropdown menu now scrolls the active item into view


src/cljc/athens/util.cljc
```diff
@@ -10,3 +10,14 @@
   []
   (-> (js/Date.) .getTime))
 
+
+(defn scrollIfNeeded
+  [element container]
+  (if (< (.-offsetTop element) (.-scrollTop container))
+    (set! (.-scrollTop container) (.-offsetTop element))
```
tangjeff0, line 17: I like to use `(.. container -scrollTop)` for JS interop, even when I don't necessarily need to chain JS methods/properties. Doesn't matter too much whether we use `..` syntax sugar or not, but I'd like to be somewhat consistent

src/cljc/athens/util.cljc
```diff
@@ -10,3 +10,14 @@
...
+    (let [offsetBottom (+ (.-offsetTop element) (.-offsetHeight element))
+          scrollBottom (+ (.-scrollTop container) (.-offsetHeight container))]
+      (when (> offsetBottom scrollBottom)
```
tangjeff0, line 20: Would prefer `(< scrollBottom offsetBottom)`just because you use a `<` earlier.

src/cljc/athens/util.cljc
```diff
@@ -10,3 +10,14 @@
   []
   (-> (js/Date.) .getTime))
 
+
+(defn scrollIfNeeded
+  [element container]
```
tangjeff0, line 15: Can you add a comment and/or link to OP explaining the arithmetic?

src/cljs/athens/keybindings.cljs
```diff
@@ -78,12 +79,16 @@
                                          (.. e preventDefault)
                                          (if (= index 0)
                                            (swap! state assoc :search/index (dec (count results)))
-                                           (swap! state update :search/index dec)))
+                                           (swap! state update :search/index dec))
+                                         (scrollIfNeeded (.getElementById js/document (str "result-" (:search/index @state)))
```
tangjeff0, line 83: Will probably want to define these HTML ids soon. They are becoming a little unwieldy

src/cljc/athens/util.cljc
```diff
@@ -10,3 +10,14 @@
   []
   (-> (js/Date.) .getTime))
 
+
+(defn scrollIfNeeded
```
tangjeff0, line 14: Can use kebab-case now. Not necessary for JS props, but makes sense for the function name

tangjeff0:
@itsrainingmani @alaq @nthd3gr33 this is one way to do scroll for a dropdown list with arrow-keys. We can probably turn off the on-hover as well, and just use arrow-keys to change `index` of list.

Issue #248, opened 2020-07-12
Add docs for connecting to REPL


prabhath6:
Update the contributing docs with details on how to connect to REPL.

- [x] [Cursive](https://cursive-ide.com/)
- [x] [CIDER](https://github.com/clojure-emacs/cider)
- [x] [Calva](https://marketplace.visualstudio.com/items?itemName=betterthantomorrow.calva)
- [ ] VIM plugins
   - [ ] [vim-iced](https://github.com/liquidz/vim-iced)
   - [ ] [conjure](https://github.com/Olical/conjure)
   - [ ] [fireplace](https://github.com/tpope/vim-fireplace)
- [ ] A video explaining on how/when to load namespaces and evaluate forms to iterate quickly while developing using REPL.

Include name and version number of the operating system, editor and any other plugins involved used by the editor for the development.

tangjeff0:
@juniusfree did a mini tutorial on Discord for Calva: https://discord.com/channels/708122962422792194/708375112537342025/729621272987959396

itsrainingmani:
CIDER - 

Editor - GNU Emacs 26.3 (build 1, x86_64-apple-darwin19.3.0, Carbon Version 162 AppKit 1894.3) of 2020-04-27
OS - MacOS Catalina v10.15.5

Process - 

1. Navigate to any file within your local `athens` folder.
2. `M-x cider-jack-in-cljs`

<img width="421" alt="Screen Shot 2020-07-13 at 8 26 55 PM" src="https://user-images.githubusercontent.com/10191300/87369921-43db0b00-c547-11ea-8ed0-eaf3ddfe6252.png">

3. Choose `shadow-cljs`

<img width="568" alt="Screen Shot 2020-07-13 at 8 28 15 PM" src="https://user-images.githubusercontent.com/10191300/87369990-72f17c80-c547-11ea-9406-5f0dc98bec75.png">
You should see something like - 
<img width="568" alt="Screen Shot 2020-07-13 at 8 28 57 PM" src="https://user-images.githubusercontent.com/10191300/87370024-8866a680-c547-11ea-90fd-93b9fb3c9c7d.png">

4. Choose `shadow` and then you should be able to choose which `shadow-cljs` build to run
<img width="608" alt="Screen Shot 2020-07-13 at 8 35 12 PM" src="https://user-images.githubusercontent.com/10191300/87370382-615ca480-c548-11ea-85e6-853279ca3ec3.png">

[...]

PR #257, opened 2020-07-14
Docs to connect to repl.

prabhath6:
Added docs for Cursive, Calva and Cider.
Resolves issues #248


tangjeff0:
Revised the instructions a little bit. The next step would be training on how/when to load namespaces and evaluate forms to iterate quickly while developing! But that's a little harder with text. A video might be better!

prabhath6:
I am not familiar with the VIM plugins, hoping someone would help out with it. I agree a video explaining REPL driven development will be nice.

PR #262, opened 2020-07-15
feat(blocks): esc to close menus and stop editing

shanberg:

- Pressing escape closes slash menus, block/page search menus, and ends block editing, in that order.


tangjeff0:
Should I ignore #261 ?

shanberg:
> Should I ignore #261 ?

Yep. I gitted that one up pretty bad. This just has the significant changes.

PR #265, opened 2020-07-16
refactor(style): prefer rem units

shanberg:

Default to rem units for lengths, with some exceptions:
- _smallest length above zero_ = px units
- sizes that scale with context (block typography): em units


tangjeff0:
Please resolve conflicts @shanberg

PR #266, opened 2020-07-16
feat(page): block in page menu

shanberg:
[Demo](https://shanberg.github.io/athens)

Screen Shot 2020-07-16 at 10 32 00 PM

Adds dropdown menu with some placeholder page controls: add to bookmarks and delete.

### DIFF
--- a/src/cljs/athens/views/node_page.cljs
+++ b/src/cljs/athens/views/node_page.cljs
@@ -9,6 +9,7 @@
     [athens.views.blocks :refer [block-el]]
     [athens.views.breadcrumbs :refer [breadcrumbs-list breadcrumb]]
     [athens.views.buttons :refer [button]]
+    [athens.views.dropdown :refer [page-menu-component]]
     [cljsjs.react]
     [cljsjs.react.dom]
     [clojure.string :as string]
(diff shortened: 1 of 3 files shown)


src/cljs/athens/views/node_page.cljs
```diff
@@ -244,9 +263,11 @@
   [ident]
   (let [{:keys [block/uid node/title] :as node} (db/get-node-document ident)
         editing-uid @(subscribe [:editing/uid])
-        timeline-page? (is-timeline-page uid)]
+        timeline-page? (is-timeline-page uid)
+        show-page-menu? (r/atom false)
+        page-menu-position (r/atom {:x 0 :y 0})]
```
tangjeff0, line 268: Your `atom` skills have leveled 🆙  I would probably merge the atoms into atom. Clojurians typically like to co-locate all stateful (read: bad) things in one place. This is good from a design perspective.

Another reason to keep state together is that it's just one less variable/parameter we have to keep track of. For example, it would not be great if we had an individual atom for each key-value in `block-component`.

Something that I haven't explained properly, and I don't think I fully understand it yet either, is why sometimes we create the atom in `x-el` and sometimes in `x-component`.

src/cljs/athens/views/node_page.cljs
```diff
@@ -201,6 +212,14 @@
         :class      (when (= editing-uid uid) "is-editing")
         :auto-focus true
         :on-change  (fn [e] (db-handler (.. e -target -value) uid))}])
+    [button {:on-click-fn (fn [e]
+                            (doall (swap! show-page-menu? not)
+                                   (reset! page-menu-position {:x (.. e -target getBoundingClientRect -left) :y (.. e -target getBoundingClientRect -bottom)})))
```
tangjeff0, line 217: I would make a `let` binding for bounding client rect. This makes it easier to acces `-left` and `-bottom`.

I don't think the `doall` is necessary.

Issue #240, opened 2020-07-10
Figure out why loading big pages is so slow


tangjeff0:
My guess is that it is `any-char` in the parser. Please see the file attached.

I timed three different functions against one another.

1. `identity` - 0.14-0.295 msecs
1. `parse-and-render` - 12681-14405 msecs
1. `parser/block-parser` 14813-15369 msecs

I ran each function 5 times. The first four times have no print logs. The last iteration prints the output of the function.

`cmd-f` "Elapsed time".

[tmp.log](https://github.com/athensresearch/athens/files/4922329/tmp.log)

```

(defn time-me
  [data f]
  (doall
    (for [x data]
      (f x)))
  nil)


(map (fn [f]
       (dotimes [i 5]
         (if (= i 4)
           (time (time-me data (comp prn f)))
           (time (time-me data f)))))
  [identity parse-and-render parser/block-parser])
```

roryokane:
It makes sense that `any-char` would be slow, so that’s probably a good guess.

The parse rule introduced [in commit `9433e79`](https://github.com/athensresearch/athens/commit/9433e79d2c30f5e7ab2ccd9feb953132b0e30b96#diff-881af7a0c0560f5d6837f19e442a8132R20) demonstrates a more efficient way to parse “any chars”:

~~~
<any-non-page-link-chars> = #'[^\\[\\]]*'
~~~

We can probably replace other uses of `any-chars` with similar constructions.

I didn’t write the parser with regexes like that to start with because such regexes slightly increase the difficulty of adding new syntax that may be nested. For example, if page links could contain `**bold**` text, the above regex would have to be `#'([^*\\[\\]]|\\*[^*])*'` (or you could use a shorter regex paired with another parse rule). The information that the syntax for bold text is `**` would be duplicated both in the `bold` rule and in regexes of rules that can contain bold text.

nthd3gr33:
Thank you for looking into this issue @roryokane 🙏

HaojiXu:
I read the issues section of instaparse, and here's their advice: 
* <https://github.com/Engelberg/instaparse/issues/131>
* <https://github.com/Engelberg/instaparse/blob/master/docs/Performance.md>

Because people over there mentioned avoid regex combinations & use LL(1)-like rules as much as possible, so I made a experimental parser with their advices here: https://github.com/athensresearch/athens/compare/master...HaojiXu:master

[...]

PR #268, opened 2020-07-17
(experimental) fix(parser): performance fix for parser using LL(1)-like parsing

HaojiXu:
See also: https://github.com/athensresearch/athens/issues/240#issuecomment-659948413

<s>Couldn't pass the test suite because</s> the unicode bare tags wouldn't be recognized in this config. Really curious if anyone knows how to address this problem. _Edit: used temporary solution to most of the use cases. Still curious to see what's the best practice though!_

Edit: Demo here: <https://haojixu.github.io/athens>

Thanks :)

### DIFF
--- a/src/cljc/athens/parser.cljc
+++ b/src/cljc/athens/parser.cljc
@@ -58,13 +58,13 @@
    (* -- However, I think in many cases a more specific rule can be used. So we will migrate away from uses of this rule. *)
    
    (* Here are a list of 'stop characters' we implemented, to get the LL(1) performance. *)
-   (* The current reserved characters are:  ->  ( [ * < ` {  # <- *)
+   (* The current reserved characters are:  ->  ( [ * < ` {  # ! <- *)
    (* Note that since our grammar is a left-recursive one, we only use the opening chars in the pair. *)
    (* IMPORTANT: if you are adding new reserved characters to the list, remember to change them all in the following regex & update the list above! *)
    (* Regex could be a thinker at times, but you can use this tool https://regex101.com/ for a visual debugging experience. *)
-   <non-reserved-char> = #'[^\\(\\[\\*\\<\\`\\{\\#]'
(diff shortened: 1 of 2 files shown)


src/cljc/athens/parser.cljc
```diff
@@ -38,13 +50,24 @@
...
+   <non-bold-chars> = #'[^\\*]*'
+   bold = <'**'> non-bold-chars <'**'>
    
-   (* It’s useful to extract this rule because its transform joins the individual characters everywhere it’s used. *)
-   (* However, I think in many cases a more specific rule can be used. So we will migrate away from uses of this rule. *)
-   any-chars = any-char+
+   (* -- It’s useful to extract this rule because its transform joins the individual characters everywhere it’s used. *)
+   (* -- However, I think in many cases a more specific rule can be used. So we will migrate away from uses of this rule. *)
    
+   (* Here are a list of 'stop characters' we implemented, to get the LL(1) performance. *)
```
tangjeff0, line 60: Can you elaborate a bit on the difference between comments that begin with `--`?

roryokane:
@HaojiXu I don’t understand exactly what problem you were running into with the Unicode character classes in the hashtag parsing regexp, but I wondered if it was related to browser support for that feature:

- Unicode property escapes support table [on Can I use…][caniuse-support]
- Unicode property escapes support table [on MDN][mdn-support]

From those tables it seems that `\p{…}` in regexps is supported in all modern browsers as well as Node since v10. However, Firefox only gained this support in the last month with the release of version 78. Were you testing on Firefox 77? If so, that could be the problem.

[caniuse-support]: https://caniuse.com/#feat=mdn-javascript_builtins_regexp_property_escapes
[mdn-support]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Browser_compatibility

PR #274, opened 2020-07-18
feat(blocks): components support (WIP)

HaojiXu:
This is a work in progress - see also the discord discussions!


HaojiXu:
# Discussion:

1. Where to parse? My current idea (implemented in this PR) based on performance concerns is: treat everything into one big string as long as they are inside the {{component}} double brackets. Then, we use Regexes to go through every component (stock or custom) and determine if there is a match.
    * If there are complex syntaxes still (e.g. {{query}}, we should build a seperate parser in the component file to ensure maximum compatibility (custom components can have complex syntaxes too!) and performance (if there aren't any block with this complex syntax, it won't be processed at all)
    * Cons: maybe this would bring code fragmentation? Also might be harder to maintain.

Demo: <https://haojixu.github.io/athens>

src/cljs/athens/components/website_embeds.cljs
```diff
@@ -0,0 +1,30 @@
+(ns athens.components.website-embeds)
+
+
+(def component-youtube-embed
+  {:match  #"\[\[youtube\]\]\:.*"
+   :render (fn [content uid]
+             [:span {:id (str content uid)}
```
tangjeff0, line 7: May want to create a generic span around all components that calls `(.. e stopPropagation)` in order to not drop into the textarea on click. Maybe `components-utils.default-component`

src/cljc/athens/components_util.cljc
```diff
@@ -0,0 +1,22 @@
+(ns athens.components-util
```
tangjeff0, line 1: I would actually put all your component-related work into one namespace for now: `src/cljs/athens/components.cljs`

- components_util
- website_embeds
- todo

src/cljs/athens/components/website_embeds.cljs
tangjeff0, line 7: Remove id

src/cljs/athens/components/website_embeds.cljs
```diff
@@ -0,0 +1,30 @@
...
+              [:iframe {:width       640
+                        :height      360
+                        :src         (str "https://www.youtube.com/embed/" (get (re-find #".*v=([a-zA-Z0-9_\-]+)" content) 1))
+                        :allow       "accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"}]])})
+
+
+(defn find-weblink
+  [content]
+
```
tangjeff0, line 16: Delete empty line

src/cljc/athens/components_util.cljc
```diff
@@ -0,0 +1,22 @@
...
+  [content uid]
+  [:span {:id uid} [:button {:class "component-default"} content]])
+
+;; TODO: use metaprogramming to achieve dynamic rendering with both basic components and custom components
+(defn render-component
+  "Renders a component using its parse tree & its uid."
+  [content uid]
+  (let [components (concat todo/components website-embeds/components)
+        rendered   (some (fn [comp]
+                           (if (re-matches (get comp :match) content)
```
tangjeff0, line 17: Keyword's naturally index maps, don't need `get`. e.g.

Both `(:match comp)` or `(comp :match)` would work. I prefer the former.

tangjeff0, line 17: `if` statement that returns `nil` in the else clause should just use `when`

src/cljc/athens/components_util.cljc
```diff
@@ -0,0 +1,22 @@
...
+                             (get comp :render)
+                             nil)) components)]
+    (if rendered
```
tangjeff0, line 20: `render` makes more sense than `rendered` to me. The render hasn't happened yet.

src/cljc/athens/components_util.cljc
```diff
@@ -0,0 +1,22 @@
...
+      (rendered          content uid)
+      (default-component content uid))))
```
tangjeff0, line 22: Both of these return hiccup. You should use bracket notation: https://github.com/reagent-project/reagent/blob/master/doc/UsingSquareBracketsInsteadOfParens.md

src/cljc/athens/components_util.cljc
```diff
@@ -0,0 +1,22 @@
+(ns athens.components-util
+  (:require
+    [athens.components.todo :as todo]
+    [athens.components.website-embeds :as website-embeds]))
+
+
+(defn default-component
+  [content uid]
+  [:span {:id uid} [:button {:class "component-default"} content]])
```
tangjeff0, line 9: Don't use `id` and `class` here

HaojiXu: Regarding `id` and `class` here (and [here](https://github.com/athensresearch/athens/pull/274#discussion_r456835720) and [here](https://github.com/athensresearch/athens/pull/274#discussion_r456835405)), I'd say that it's more for the sake for future themability (like how Roam can make the buttons/component elements look different based on different `roam/css`). If we're doing this though, we should document them for future theme-makers to reference. Or we can do it at some later time.

tangjeff0: Have had a few early discussions with @shanberg about custom theming. Main point: we think we can do better than `roam/css`. Main reason I don't want to not-yet-used classes/ids yet is because our current classes/ids are kind of a mess and will require a refactor soon.

HaojiXu: Makes sense! I'll just hold off adding them so we can talk later

src/cljc/athens/components_util.cljc
HaojiXu: Yes, this would've made more senes. Done!

src/cljs/athens/components/todo.cljs
```diff
@@ -0,0 +1,41 @@
...
+
+
+(def component-done
+  {:match #"\[\[DONE\]\]"
+   :render (fn [content uid]
+             [:span [:input {:type     "checkbox"
+                             :z-index  5
+                             :class    "component-todo"
+                             :checked  "true"
+                             :on-click #(todo-on-click uid #"\{\{\[\[DONE\]\]\}\}" "{{[[TODO]]}}")
```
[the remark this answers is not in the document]
HaojiXu: seems like that the string `{{[[TODO]]}}` is not always at the beginning of the block in texts not generated by `cmd-enter` - so in my opinion the two functions are still doing different (but similar) things

src/cljs/athens/components/components_util.cljs
```diff
@@ -0,0 +1,25 @@
+(ns athens.components.components-util
```
tangjeff0, line 1: Not only did I mean to use `cljs` instead `cljc`, I meant that all 3 files can be in one components file. I prefer monolithic files rather than fragmenting across several files, until it's absolutely necessary. See how big `events` and `db` are!

https://github.com/athensresearch/athens/pull/274#discussion_r456837520

src/cljs/athens/components/components_util.cljs
```diff
@@ -0,0 +1,25 @@
...
+  (:require
+    [athens.components.todo :as todo]
+    [athens.components.website-embeds :as website-embeds]))
+
+
+(defn default-component
+  [content uid]
+  ((constantly nil) uid)
```
tangjeff0, line 9: I see we're using `constantly` in many places. I don't understand what it's doing.

src/cljc/athens/parser.cljc
```diff
@@ -109,7 +109,9 @@
      :url-link-url-parts     (fn [& chars]
                                (string/join chars))
      :any-chars              (fn [& chars]
-                               (string/join chars))}
+                               (string/join chars))
+     :component              (fn [raw-content-string]
+                               (into [] (concat [:component raw-content-string] (into [] (rest (block-parser raw-content-string))))))}
```
tangjeff0, line 114: Really bad code smell using 2 `into`s and a `concat`. Without jumping into a REPL, I'm pretty confident you only need one of these vector creation functions, not 3.

src/cljs/athens/parse_renderer.cljs
```diff
@@ -83,8 +83,8 @@
   (insta/transform
     {:block         (fn [& contents]
                       (concat [:span {:class "block" :style {:white-space "pre-line"}}] contents))
-     :component     (fn [contents]
-                      (components-util/render-component contents uid))
+     :component     (fn [& contents]
+                      (components-util/render-component (first contents) uid))
```
tangjeff0, line 87: Why are we making `& contents` into a list and then calling `first` on it?

Are arguments variadic? If they are, we need comments explaining why we're dropping the rest.

tangjeff0: This goes for `page-link` as well. Why do we sometimes use `&` and sometimes not?

src/cljs/athens/components/components_util.cljs
```diff
@@ -0,0 +1,25 @@
...
+  [:button content])
+
+;; TODO: use metaprogramming to achieve dynamic rendering with both basic components and custom components
+(defn render-component
+  "Renders a component using its parse tree & its uid."
+  [content uid]
+  (let [components (concat todo/components website-embeds/components)
```
tangjeff0, line 16: Because you've defined `components` across several namespaces, you need to join them at some point. That's not inherently a bad thing, but you are joining them every single time you call `render-component`, when the computed collection never actually changes.

src/cljs/athens/components/components_util.cljs
HaojiXu: I'd say that putting components in different files could be more helpful, because the component system should be designed as extensible as possible, especially since we plan to add very complex ones later (such as `query` & `kanban`).

src/cljs/athens/components/components_util.cljs
HaojiXu: Makes sense - should've concated them only once

src/cljs/athens/parse_renderer.cljs
HaojiXu: Arguments are variadic because the contents list's first element is the unparsed string (for more efficient matching across many regexes), while the rest are parsed tree for things like dynamic references & auto page creation. Will add documentation

src/cljs/athens/components/components_util.cljs
HaojiXu: Sorry for my Clojure noob-ness - I didn't know `_` existed. Fixed

src/cljc/athens/parser.cljc
HaojiXu: Solved!

src/cljs/athens/parse_renderer.cljs
HaojiXu: Docs added

src/cljs/athens/components/components_util.cljs
tangjeff0: Whether the files are distributed or in one file doesn't affect their extensibility. I don't think we will necessarily copy verbatim how Roam implements `{{kanban}}` and `{{query}}` either. I like to build up then break out. Will still be a while until these reach the level of complexity of say `blocks`. Please couple the components for now.

src/cljs/athens/parse_renderer.cljs
tangjeff0: Thank you

HaojiXu: `page-link` might also need a bit more clarification - will add docs in a separate commit since it's unrelated to this.

src/cljs/athens/components/components_util.cljs
HaojiXu: I don't mind if the files are together or seperate personally, but putting them all together would make other contributors spend more effort when trying to add their own component (for example, I had to read the entire `blocks` before adding the auto page-link feature), which might discourage collaboration. For now, I can combine them and add separate docs to indicate which part of the file does what, which could hopefully mitigate this issue. `components_util` should remain separate though, because I'm looking into metaprogramming and would soon experiment with ways users could add their own components, and adding them into components declaration would only confuse people.

tangjeff0: I think if a contributor wanted to add a component, they would look through the previous components to see how it's done. How else would they know how to make a component? If you hadn't read the blocks file, would you have known how to make the auto-page link feature? If someone hasn't read the relevant parts of the codebase, then yes I am would discourage them from contributing yet, if they are somehow even able to complete the feature without prerequisite knowledge.

Maybe a big file is intimidating, but it's tractable. Having to look through more files in a new codebase is worse because 1) you have to know the file exists and 2) you may need to read it anyways, we wouldn't know ahead of time.

I don't know what you intend to do with metaprogramming so I don't know if it would confuse people or not having it multiple files. Please elaborate before you try to use that as support for what you want to do. Like kanban, query, and backends (on Discord), I believe you're assuming too much.

This convo sounds nitpicky. But part of what I'm saying comes from my experience doing DevOps, where I created and ran our company's CI builds. In one project I implemented monorepos with [lerna](https://github.com/lerna/lerna) and yarn workspaces. Scattering logic across multiple repos (or files in this case), increases the burden on the developer. This is especially true when the developer _needs_ to see related files to know how to figure out what to do.

There aren't unanimous best practices for separation of files. Mono vs many is still up for debate. PurelyFunctional disagrees with separating `subs` and `events` into separate files ([source](https://purelyfunctional.tv/guide/database-structure-in-re-frame/#events-and-subscriptions)). Some Clojurians have argued separating namespaces into `clj` `cljs` and `cljc` is bad practice. Here, I don't believe your points are well-formed enough to convince me otherwise.

doc/components.md
```diff
@@ -0,0 +1,32 @@
+# Athens Components Documentation
+
+Components are special syntaxes in blocks that allow complex interactions and data display from an end-user perspective. Athens has an open attitude for components: we believe that it is an important part of end-user programming. Thus, users can not only contribute to the components directory through this repository (following the general components syntax), but can also add custom components directly in the user-end, then share it with the entire Athens community (coming soon).
```
tangjeff0, line 3: There has never even been so much as a conversation on this. Assumes too much.

doc/components.md
```diff
@@ -0,0 +1,32 @@
...
+
+This documentation provides a technical overview regarding how Athens components are processed in the frontend. If you have any ideas or suggestions, feel free to open an issue!
+
+## How Components are Parsed
+
+The Athens [parser](./parser.md) considers everything in double curly brackets (`{{}}`) a component. You can have all kinds of syntaxes in a component, as the everything in the component is matched using regular expressions first and could be further parsed using instaparse if it has a more complex syntax.
+
+After the parsing process, the `:component` list would be a variadic list in which the first element is the unparsed string for more efficient pattern matching while rest of the elements are parsed tree for things like dynamic references & auto page creation. Will add documentation.
+
+Relavant code:
```
tangjeff0, line 13: Relevant

src/cljs/athens/components/default_components.cljs
```diff
@@ -0,0 +1,62 @@
...
+  {:match  #"\[\[youtube\]\]\:.*"
+   :render (fn [content _]
+             [:iframe {:width       640
+                       :height      360
+                       :src         (str "https://www.youtube.com/embed/" (get (re-find #".*v=([a-zA-Z0-9_\-]+)" content) 1))
+                       :allow       "accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"}])})
+
+
+(def component-generic-embed
+  {:match  #"iframe\:.*"
```
tangjeff0, line 53: Is it on purpose that youtube is `{{[[youtube: ]]}}` but iframe is `{{iframe: }}`? Should we be consistent?

src/cljs/athens/components/components_util.cljs
HaojiXu: Thanks for your contribution on this issue - I love how you explained your coding style from your experience and was able to relate to other people from that experience. Now I see your point in full light, I wholeheartedly agree.

On another note, my mindset gave too much weight on other incoming "occasional & inexperienced" contributors over existing contributors. Ideally we'd like something with a literate programming paradigm so that end-user programming could be more powerful and integrated, but for now we can stick with what we have.

Regarding my upcoming contributions, I'll look into clojure metaprogramming first and try to implement some user-programmable components (after discussing the specs with you guys), then we could have more conversations regarding the backend. Kanbans & queries aren't a priority on my list because they're relatively independent and could make a great training ground for newly-graduated learners :)

src/cljs/athens/components/default_components.cljs
HaojiXu: This is kind of a legacy problem from Roam - their component is exactly like how I implemented. I do believe we might need a revamp on the syntax & data structure of Roam for consistency, but in the current method's defense, YouTube is a more specific website, which makes having backlinks to YouTube make more sense than just backlinking every iframe embeds.

src/cljs/athens/components/components_util.cljs
tangjeff0: Kanbans are not a priority because I don't think anyone would miss them from Roam right now. Queries are in fact quite powerful, but might require more discussion from both eng. and design.

To your point on literate programming, mani and I briefly thought about this idea: [backup](https://backhub.co/) our github to an athens db and then begin using # and [[links]] in code comments. Link between issues, code, and docs seamlessly :)

PR #275, opened 2020-07-18
feat(athena): replace default close button

shanberg:

Screen Shot 2020-07-17 at 8 46 12 PM

Removes default blue text-input close button and replaces it with a custom close button.


shanberg:
Closed and replaced with #276

PR #277, opened 2020-07-18
feat(blocks): polished selected blocks

shanberg:

Restyled block selections to better separate selected siblings.

Screen Shot 2020-07-17 at 10 49 17 PM

### DIFF
--- a/src/cljs/athens/views/blocks.cljs
+++ b/src/cljs/athens/views/blocks.cljs
@@ -36,6 +36,7 @@
   {:display "flex"
    :line-height "2em"
    :position "relative"
+   :border-radius "0.125rem"
    :justify-content "flex-start"
    :flex-direction "column"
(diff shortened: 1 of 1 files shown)


src/cljs/athens/views/blocks.cljs
```diff
@@ -45,7 +46,22 @@
                                                      :top "2em"
                                                      :bottom "0"
                                                      :transform "translateX(50%)"
-                                                     :background (color :border-color)}]]})
+                                                     :background (color :border-color)}]
+                     [:&:after {:content "''"
```
tangjeff0, line 50: I've seen this pattern a few times `:before` or :`after` and then `:content "''"`. What does it do?

Is it being used the same way on lines 42, 50, and 63?

tangjeff0: @shanberg

PR #280, opened 2020-07-19
feat(athena): scroll search results with arrow keys

### LINKED ISSUE #271: Athena Result list does not scroll when navigating with arrow keys

**Problem**

When using arrow keys to navigate up and down through the list of search results, the list does not automatically scroll as you move past the first few items. The user is unable to see which search result is currently selected.

**Expected Behavior**

Like Roam, when you select a search result is that is initially beyond the visible items, the search result list should scroll to bring the selection into view.

**To Reproduce**

https://athensresearch.github.io/athens/ then load test data, search for a term in Athena and use the arrow keys to go down the result list.

**Screenshots**

https://www.loom.com/share/90714b294cdb4c77a0d2f69d1fe56396

itsrainingmani:
This PR closes #271 and checks off part of #126 

Athena can now scroll through the search results with arrow keys. The implementation also cycles through the search result list.

### DIFF
--- a/src/cljs/athens/views/athena.cljs
+++ b/src/cljs/athens/views/athena.cljs
@@ -176,7 +176,6 @@
         shift (.. e -shiftKey)
         {:keys [index query results]} @state
         item (get results index)]
-
     (cond
       ;; FIXME: why does this only work in Devcards?
       (= key KeyCodes.ESC)
@@ -205,10 +204,30 @@
       ;; TODO: change scroll as user reaches top or bottom
       ;; TODO: what happens when user goes to -1? or past end of list?
       (= key KeyCodes.UP)
-      (swap! state update :index dec)
+      (when (> index 0)
+        (do
+          (swap! state update :index dec)
+          (let [select-el (first (array-seq (. js/document getElementsByClassName "selected")))
+                next-el (.. select-el -previousElementSibling)
+                athena-el (first (array-seq (. js/document getElementsByClassName "athena")))
+                result-el (nth (array-seq (.. athena-el -children)) 2)
+                result-box (.. result-el getBoundingClientRect)
+                next-box (.. next-el getBoundingClientRect)]
+            (if (< (.. next-box -top) (.. result-box -top))
+              (.. next-el (scrollIntoView true {:behavior "auto"}))))))
 
       (= key KeyCodes.DOWN)
-      (swap! state update :index inc)
+      (when (< index (dec (count results)))
+        (do
+          (swap! state update :index inc)
+          (let [select-el (first (array-seq (. js/document getElementsByClassName "selected")))
+                next-el (.. select-el -nextElementSibling)
+                athena-el (first (array-seq (. js/document getElementsByClassName "athena")))
+                result-el (nth (array-seq (.. athena-el -children)) 2)
+                result-box (.. result-el getBoundingClientRect)
+                next-box (.. next-el getBoundingClientRect)]
+            (if (> (.. next-box -bottom) (.. result-box -bottom))
+              (.. next-el (scrollIntoView false {:behavior "auto"}))))))
 
       :else nil)))
 
--- a/src/cljs/athens/views/athena.cljs
+++ b/src/cljs/athens/views/athena.cljs
@@ -169,14 +169,15 @@
                                            (take 20 (search-in-block-content query)))
                                    vec)}))))
 
+
 (defn is-beyond-rect?
   "Checks if any part of the element is above or below the container's bounding rect"
   [element container]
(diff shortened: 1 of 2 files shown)


src/cljs/athens/views/athena.cljs
```diff
@@ -170,13 +170,22 @@
...
 
 
+(defn is-beyond-rect?
+  "Checks if any part of the element is above or below the container's bounding rect"
+  [element container]
+  (let [el-box (.. element getBoundingClientRect)
+        cont-box (.. container getBoundingClientRect)]
+    (or
+      (> (.. el-box -bottom) (.. cont-box -bottom))
+      (< (.. el-box -top) (.. cont-box -top)))))
```
tangjeff0, line 180: Move to `util`. This can be re-used by inline search and slash commands.

src/cljs/athens/views/athena.cljs
```diff
@@ -170,13 +170,22 @@
...
+
 (defn key-down-handler
   [e state]
   (let [key (.. e -keyCode)
         shift (.. e -shiftKey)
         {:keys [index query results]} @state
         item (get results index)]
-
     (cond
       ;; FIXME: why does this only work in Devcards?
```
tangjeff0, line 190: Can delete this comment now XD.

src/cljs/athens/views/athena.cljs
```diff
@@ -205,10 +214,24 @@
...
       (= key KeyCodes.UP)
-      (swap! state update :index dec)
+      (do
+        (swap! state update :index #(dec (if (zero? index) (count results) index)))
+        (let [cur-index (:index @state)
+              input-el (.. e -target)
+              result-el (.. (. input-el closest "div.athena") -lastElementChild)
+              next-el (nth (array-seq (.. result-el -children)) cur-index)]
+          (when (is-beyond-rect? next-el result-el)
+            (.. next-el (scrollIntoView (not= cur-index (dec (count results))) {:behavior "auto"})))))
```
tangjeff0, line 224: After a long day, trying to read this is requiring a lot of effort, so I think there's too much going on.

src/cljs/athens/views/athena.cljs
```diff
@@ -205,10 +214,24 @@
       ;; TODO: change scroll as user reaches top or bottom
       ;; TODO: what happens when user goes to -1? or past end of list?
```
tangjeff0, line 215: delete

src/cljs/athens/views/athena.cljs
```diff
@@ -205,10 +214,24 @@
...
       (= key KeyCodes.UP)
-      (swap! state update :index dec)
+      (do
+        (swap! state update :index #(dec (if (zero? index) (count results) index)))
+        (let [cur-index (:index @state)
+              input-el (.. e -target)
+              result-el (.. (. input-el closest "div.athena") -lastElementChild)
```
tangjeff0, line 221: What does this do? add comment?

tangjeff0: also you are already using `..` syntax sugar, so it should just be
```(.. input-el (closest "div.athena") -lastElementChild)```
recommend http://app.klipse.tech/

Comment (itsrainingmani) [reply src/cljs/athens/views/athena.cljs:221]: > What does this do? add comment?

Yeah looking at it now, it's not obvious why I'm doing that. I'll add some comments 👍

Comment (itsrainingmani) [reply src/cljs/athens/views/athena.cljs:221]: > also you are already using `..` syntax sugar, so it should just be
> `(.. input-el (closest "div.athena") -lastElementChild)`
> recommend http://app.klipse.tech/

This is awesome! Thanks for the link.

**Comment by itsrainingmani** [comment]
Hi @tangjeff0 . I've made the required changes. The only thing I'm a little unsure of is how much commenting the event handler needed. So let me know if it's too much

File: src/cljs/athens/views/athena.cljs [hunk 6]
```diff
@@ -202,13 +200,37 @@
...
+      (do
+        (.. e preventDefault)
+        (swap! state update :index #(dec (if (zero? %) (count results) %)))
+        (let [cur-index (:index @state)
+
+              ;; Search input box
+              input-el (.. e -target)
+
+              ;; Get the result list container which is the last element child
+              ;; of the whole athena component
```
tangjeff0, line 213: This makes a lot more sense to me now. I'm pretty sure I wrote `[:div.athena]`, but that should probably be an HTML id. `.athena` also doesn't do anything functionally, so maybe not that useful.

This could also be simplified if we just gave an HTML id/class to the results list directly, but not important.

Issue #292, opened 2020-07-21
Pages - Team Seneca


tangjeff0:
#96 

- [x] deletion
- [ ] merging
- [ ] link filters

nthd3gr33:
Can you try to assign yourselves @itsrainingmani @alaq?

itsrainingmani:
Just did @nthd3gr33

alaq:
For reference, for whoever picks this up later (us or someone else):

- The code we currently have lives [here](https://github.com/itsrainingmani/athens/tree/link-filters).
- We have made the popup show  when clicking on the filter icon next to linked references. See https://github.com/itsrainingmani/athens/commit/243060004f7611f183d5db9e17abcb68ea23084d
- We have positioned the popup correctly (stays put when scrolling for instance -- but still very rough). See https://github.com/itsrainingmani/athens/commit/7c27739bd738b51d17b53cd237c3cd4cc67b3e11

Issue #96, opened 2020-05-27
Pages


tangjeff0:
# Tasks

- [x] delete page
- [x] add to shortcuts
- [x] merge pages if editing a non-date page. show confirmation alert #61 #62 
- [x] hide linked/unlinked refs if in Daily Pages view #66 
- [x] prevent title editing if date page. show warning alert
- [ ] button to create a new page? can already create pages in Athena
- [ ] filters #65

nthd3gr33:
Please comment @itsrainingmani @alaq so I can assign our team to this issue.

itsrainingmani:
Let's go @nthd3gr33 !

alaq:
👋

nthd3gr33:
We started looking into the shortcut part of this issue, but we have discovered that quite a bit was already done. So we will likely move on to the deleting pages part of the issue.

We have questions about implementation of this feature:
What is the expected UX for it? In roam, you check pages in the all pages list and then you can bulk delete
Plus what happens to empty pages that have no links to them
do we keep them? eagerly delete them?

nthd3gr33:
(Fat fingers, sorry 😅 )

We started looking into the shortcut part of this issue, but we have discovered that quite a bit was already done. So we will likely move on to the deleting pages part of the issue.

We have questions about implementation of this feature, specifically in regards to expected UX for deleting pages:  
1. In Roam, you can check pages in the All Pages list and then you can bulk delete. Do we also want to implement that here? Or are bulk operations out of the scope of this issue?
1. What happens to empty pages that have no links to them. Do we keep them? Eagerly delete them?
1. If you delete a page, should all corresponding links to it be deleted? (this is current Roam behavior)

tangjeff0:
The main part remaining for shortcuts is decrementing all the indices after the index that was removed, which is NBD.

1. Bulk operations are part of All Pages, out of scope.
1. Can you give me an example?
1. Yes, and we have all the `:block/uid`s in linked references.

Made #292 for you three. I also think you can assign yourselves to the issue now. I had to add the athens repo to the devs team.

alaq:
> 2. Can you give me an example?

We're thinking of pages that were created by a link, but the link was later removed. Here is what is happening in Roam:
1. Create new block with link to [[new page]]
1. `new page` is visible in All Pages
1. remove link to `new page` in new block
1. `new page` is no longer visible in All Pages (it has been deleted)

tangjeff0:
[...]

PR #295, opened 2020-07-23
feat(page): delete page and child blocks recursively

alaq:
This PR checks off part of #292 (and by extension #96).

It implements the following:
- It lets a user delete a page on click in the dropdown
- It navigates away from soon to be deleted page (navigates to `All Pages`, but could be to something else?)
- Adds helper function to get id for a given `uid`
- It deletes children blocks when deleting page

In `node_page.cljs`, in the `on-click` we are redirecting first and deleting afterwards to minimize the time that the user would see the 404 page. https://github.com/athensresearch/athens/pull/295/files#diff-38e22d4e464b19fd15f43b782729eb20R249

We realize this is not ideal. What is the best way to solve this issue?

~~We're working on stripping blocks of the link to the now deleted page.~~ This is now taken care of by #297

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -249,6 +249,24 @@
           (recur (get ch (dec n))))))))
 
 
+(defn get-id
+  [uid]
+  (-> (d/q '[:find ?id
+             :in $ ?uid
+             :where [?id :block/uid ?uid]]
+           @dsdb
+           uid)
+      ffirst))
+
+
+(defn get-children-recursively
+  "Get list of children UIDs for given block ID (including the root block's UID)"
+  [uid]
+  (let [document (->> @(pull dsdb '[:block/order :block/uid {:block/children ...}] (get-id uid)))]
+    (->> (tree-seq :block/children :block/children document)
+         (map :block/uid))))
+
+
 (defn re-case-insensitive
   "More options here https://clojuredocs.org/clojure.core/re-pattern"
   [query]
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -200,7 +200,7 @@
 
 (defn get-node-document
   [id]
-  (->> @(pull dsdb '[:db/id :node/title :block/uid :block/string :block/open :block/order {:block/children ...} :edit/time] id)
+  (->> @(pull dsdb '[:db/id :node/title :block/uid :block/string :block/open :block/order :page/sidebar {:block/children ...} :edit/time] id)
        sort-block-children))
 
 
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -249,24 +249,6 @@
           (recur (get ch (dec n))))))))
 
 
-(defn get-id
-  [uid]
-  (-> (d/q '[:find ?id
-             :in $ ?uid
-             :where [?id :block/uid ?uid]]
-           @dsdb
-           uid)
-      ffirst))
-
-
-(defn get-children-recursively
-  "Get list of children UIDs for given block ID (including the root block's UID)"
-  [uid]
-  (let [document (->> @(pull dsdb '[:block/order :block/uid {:block/children ...}] (get-id uid)))]
-    (->> (tree-seq :block/children :block/children document)
-         (map :block/uid))))
-
-
 (defn re-case-insensitive
   "More options here https://clojuredocs.org/clojure.core/re-pattern"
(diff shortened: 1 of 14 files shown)


itsrainingmani:
🍾 🚀 🥇

alaq:
@itsrainingmani just stumbled upon https://docs.datomic.com/on-prem/transactions.html#dbfn-retractentity which does a lot of the work that we are doing manually in this PR (like `get-children-recursively` -- which is not needed anymore). We're updating this PR.

itsrainingmani:
> @itsrainingmani just stumbled upon https://docs.datomic.com/on-prem/transactions.html#dbfn-retractentity which does a lot of the work that we are doing manually in this PR (like `get-children-recursively` -- which is not needed anymore). We're updating this PR.

So after some additional testing, while `:db/retractEntity` does retract an entity, it doesn't seem to actually recursively retract all the child entities.
* Updated the branch to include the `get-id`, `get-children-recursively` functions again.
* `:page/delete` event handler now uses `:db/retractEntity` on all the children to retract them in place of `:db/retract`.

src/cljs/athens/db.cljs
```diff
@@ -249,6 +249,24 @@
...
 
 
+(defn get-id
+  [uid]
+  (-> (d/q '[:find ?id
+             :in $ ?uid
+             :where [?id :block/uid ?uid]]
+           @dsdb
+           uid)
+      ffirst))
```
tangjeff0, line 259: Should be `(d/q '[:find ?id .`

Mind the period.

https://docs.datomic.com/on-prem/query.html#find-specifications

tangjeff0: I would actually just add this below `db/e-by-av`. Direct indexing is faster.

```
(defn v-by-ea
  [e a]
  (-> (d/datoms @dsdb :eavt e a) first :v))
```

src/cljs/athens/db.cljs
```diff
@@ -249,6 +249,24 @@
...
+
+
+(defn get-children-recursively
+  "Get list of children UIDs for given block ID (including the root block's UID)"
+  [uid]
+  (let [document (->> @(pull dsdb '[:block/order :block/uid {:block/children ...}] (get-id uid)))]
+    (->> (tree-seq :block/children :block/children document)
```
tangjeff0, line 266: Nice `tree-seq`. Two `->>` seems a little unnecessary.

src/cljs/athens/events.cljs
```diff
@@ -284,6 +284,12 @@
        :dispatch [:editing/uid child-uid]})))
 
 
+(reg-event-fx
+  :page/delete
+  (fn [_ [_ uid]]
+    {:transact! (vec (map (fn [uid] [:db/retractEntity [:block/uid uid]]) (get-children-recursively uid)))}))
```
tangjeff0, line 290: `mapv` instead of `(vec (map ...))`

src/cljs/athens/db.cljs
itsrainingmani: We don't need the `get-id` function at all now since all it does is get an id given a `uid`. Which the `e-by-av` function does perfectly.

alaq: It doesn't look like we need `v-by-ea` for this, and we can use `e-by-av`. Should we add it for future use?

tangjeff0: Oh good point @itsrainingmani. Yeah, add `v-by-ea` and add it to `.carve_ignore`

src/cljs/athens/db.cljs
```diff
@@ -249,6 +254,13 @@
           (recur (get ch (dec n))))))))
 
 
+(defn get-children-recursively
+  "Get list of children UIDs for given block ID (including the root block's UID)"
+  [uid]
+  (let [document (->> @(pull dsdb '[:block/order :block/uid {:block/children ...}] (e-by-av :block/uid uid)))]
+    (map :block/uid (tree-seq :block/children :block/children document))))
```
tangjeff0, line 261: I think what you want is

```
(defn get-children-recursively
  "Get list of children UIDs for given block ID (including the root block's UID)"
  [uid]
  (->> @(pull dsdb '[:block/order :block/uid {:block/children ...}] (e-by-av :block/uid uid))
    (tree-seq :block/children :block/children)
    (map :block/uid)))
```

alaq: Thank you, just pushed the changes!

PR #297, opened 2020-07-23
feat(blocks): add hashtag page auto-create and auto-delete

HaojiXu:
<s>Currently waiting for page deletion from <https://github.com/athensresearch/athens/pull/295>.</s>

Update: integrated #295 and added full page deletion.

### DIFF
--- a/src/cljs/athens/util.cljs
+++ b/src/cljs/athens/util.cljs
@@ -1,6 +1,9 @@
 (ns athens.util
   (:require
+    [athens.db :as db]
+    [athens.patterns :as patterns]
     [clojure.string :as string]
+    [posh.reagent :refer [#_pull q]]
     [tick.alpha.api :as t]
     [tick.locale-en-us]))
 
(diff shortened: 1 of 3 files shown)


HaojiXu:
@tangjeff0 I've added #295 and now auto page addition/deletion is 100% working on my test. Demo isn't available because `devcards/dropdown` is still broken, but this branch is now ready for code review!

src/cljs/athens/util.cljs
```diff
@@ -92,3 +95,71 @@
       (string/replace x #"PM" "pm"))))
 
 
+;; -- Linked & Unlinked References ----------
```
tangjeff0, line 98: I think `db` is a slightly better place to put this than `util`, but neither are great.

src/cljs/athens/views/blocks.cljs
```diff
@@ -249,23 +249,49 @@
...
                                      (let [now (now-ts)
                                            uid (gen-block-uid)]
-                                       (dispatch [:transact [{:node/title  inner-title
-                                                              :block/uid   uid
-                                                              :edit/time   now
-                                                              :create/time now}]])))
-                                   (str "[[" inner-title "]]")))} (parser/parse-to-ast value)))
+                                       (db-fn inner-title now uid)))
+                                   (str "[[" inner-title "]]")))
+                    :hashtag   (fn [title]
```
tangjeff0, line 264: Does this also include `#[[]]` and `#[[nested [[links]]]]`?

src/cljs/athens/util.cljs
HaojiXu: I agree! We can just put it here for now, and potentially create a new references.cljs when we get started on graph previews.

src/cljs/athens/util.cljs
```diff
@@ -92,3 +95,71 @@
...
+
+
+(defn get-data
+  [pattern]
+  (-> pattern get-ref-ids merge-parents-and-block group-by-parent seq))
+
+
+(defn get-data-by-block
+  [pattern]
+  (-> pattern get-ref-ids merge-parents-and-block seq))
```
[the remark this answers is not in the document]
HaojiXu: Basically get data by block doesn't group references by their source page. Since this could also be useful later (for various counting, graph, etc), I kept it a separate function.

src/cljs/athens/util.cljs
```diff
@@ -92,3 +95,71 @@
...
+  [title]
+  (-> title patterns/linked get-data-by-block))
+
+
+(defn get-unlinked-references
+  [title]
+  (-> title patterns/unlinked get-data))
+
+
+(defn count-linked-references-excl-uid
```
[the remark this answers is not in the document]
HaojiXu: If I reasoned correctly, because dispatch is async, simply decreasing count by one could result in a race condition that might or might not return one less reference than expected.

src/cljs/athens/views/blocks.cljs
HaojiXu: I think it will work because the parsing is bottom-up, but didn't know exactly. I'll test it

HaojiXu: Tested - it works with `#[[]]` but not with `#[[nested [[links]]]]`. Upon further inspection, the latter is a parser issue. Instead of fixing it here, I think it might be better to start a separate PR where I fix this issue and add some unit tests to prevent these problems from happening in the future.

tangjeff0: sounds good, thanks @HaojiXu

PR #303, opened 2020-07-24
fix(athena): escape special chars in search input and title

### LINKED ISSUE #301: Athena Regex issues
Athena/Athens has some issues when you enter any unterminated regex special characters `[`, `(` into the text input.

**Problem**

The main issue is when you try to navigate to a page that has an unterminated `[`, `(` in the title. The entire page freezes and doesn't render anything.
Entering special characters like `[`, `(` in Athena and trying to create a new page also has some weird behavior. Depending on the combination of characters, Athena doesn't display the page creation button, or the page is created but with some characters missing.

**Expected Behavior**

@nthd3gr33 tested the behavior on Roam and there are no issues with navigating to pages with unterminated `[`, `(` characters in the title.
* Navigating to a page that has `[`, `(` in the title should not freeze the whole page.
* Entering these characters into Athena should not cause Athena to freeze and/or should create the page with all the input characters.

**Dependencies**

- Athens: Latest Master
- browser: Firefox, Edge, Chrome

**To Reproduce**

For Page Creation issue - 
1. Open Athena
2. Enter search text that has unterminated `[`, `(` characters. For example - `Athens [(`. And try to create the page

For Page Freezing issue - 
1. Load the Test Data
2. Open Athena
3. Search for "Athens" and click on the result "Athens [ATH" in the search results
4. Page should freeze

**Screenshots**

https://www.loom.com/share/725a748a7d90433dae55faa10b92af28

image

itsrainingmani:
This PR closes #301 

* create util function for escaping all regexp special characters in a given string

* escape the title in Linked and Unlinked references.

* escape the query in re-case-insensitive

Co-authored-by: nthd3gr33

### DIFF
--- a/src/cljs/athens/util.cljs
+++ b/src/cljs/athens/util.cljs
@@ -92,3 +92,16 @@
       (string/replace x #"PM" "pm"))))
 
 
+;; -- Regex -----------------------------------------------------------
+
+;; https://stackoverflow.com/a/11672480
+(def regex-esc-char-map
+  (let [esc-chars "()*&^%$#![]"]
+    (zipmap esc-chars
+            (map #(str "\\" %) esc-chars))))
+
+
+(defn escape-str
+  "Take a string and escape all regex special characters in it"
+  [str]
+  (string/escape str regex-esc-char-map))
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -1,5 +1,6 @@
 (ns athens.db
   (:require
+    [athens.util :refer [escape-str]]
     [clojure.edn :as edn]
     [datascript.core :as d]
     [posh.reagent :refer [posh! pull]]))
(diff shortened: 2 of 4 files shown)


src/cljs/athens/util.cljs
```diff
@@ -92,3 +92,16 @@
...
+(def regex-esc-char-map
+  (let [esc-chars "()*&^%$#![]"]
+    (zipmap esc-chars
+            (map #(str "\\" %) esc-chars))))
+
+
+(defn escape-str
+  "Take a string and escape all regex special characters in it"
+  [str]
+  (string/escape str regex-esc-char-map))
```
tangjeff0, line 107: Just a FYI, Clj is both nice and not nice because you can shadow/override clojure.core symbols. In this closure, you are shadowing `str` with an input parameter, so if you tried to use `str` the clojure.core function like `(str "hello" "world")` you would get a function invocation error.

tangjeff0: I had some difficulty tracking some bugs in `keybindings` because I bound a variable with the symbol `key`, which is also a clojure.core function!

PR #307, opened 2020-07-29
rfct: simpler code for count function

fnumatic:
+ code is more simple
- sligthly more complex computation

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -489,9 +489,6 @@
 
 (defn count-linked-references-excl-uid
   [title uid]
-  (reduce (fn [current-count ref]
-            (if (= (:block/uid ref) uid)
-              current-count
-              (inc current-count)))
-          0
-          (get-linked-references-by-block title)))
+  (->> (get-linked-references-by-block title)
+       (remove #(= (:block/uid %) uid))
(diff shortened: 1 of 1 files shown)


src/cljs/athens/db.cljs
```diff
@@ -489,9 +489,6 @@
...
   [title uid]
-  (reduce (fn [current-count ref]
-            (if (= (:block/uid ref) uid)
-              current-count
-              (inc current-count)))
-          0
-          (get-linked-references-by-block title)))
+  (->> (get-linked-references-by-block title)
+       (remove #(= (:block/uid %) uid))
+       count))
```
HaojiXu, line 494: Thank you! This definitely is much more readable and intuitive. Just a small styling suggestion: maybe move line 494 up to 493 so that the `remove` function is together?

fnumatic: I don't understand, could you be more specific.
The form in L494 is independent from the one in L493.
remove -> filtered items
count -> number of filtered items

HaojiXu: Oh yeah NVM, didn't notice the `->>` operation :)

tangjeff0:
Nice first PR!

PR #310, opened 2020-07-30
rfct: use goog dom function

fnumatic:


fnumatic:
> Should we do this globally? Many files use `getElementById`

I'm on it

tangjeff0:
Okay, could just update this PR, don't need to close :)

fnumatic:
yeah know. family interference

fnumatic:
curious, on athena.cljs L296 there is a call of set! with one param ???

tangjeff0:
That clears the input text @fnumatic

PR #316, opened 2020-08-02
Inline search fixes

### LINKED ISSUE #315: Errors prevent you from using inline search when entering links
**Problem**

I am unable to use inline search when entering page and block links. Whenever I try to enter the first character inside the double brackets (square or round), an error is thrown and no text is entered, so that I can't type anything.

20200802_athens_link_error

**Expected Behavior**

I should be able to type some characters and see a list of search results inline.

**Dependencies**

- Athens: commit 98deccc15022c78a3ed6c48fee6a8582b718a18c.
- browser: Chrome 84.0.4147.105
- JVM: Java 1.8.0_252 OpenJDK 64-Bit Server VM
- node.js: v14.4.0

**To Reproduce**

1. Create a new block.
2. Enter two square brackets or parentheses.
3. Enter any character.
4. See that nothing happens and errors are thrown in the console.

**Screenshots**

See above

bonobo-d:
(This PR is mostly for learning. If you think it is out of place/unnecessary, I don't mind closing it. I learned a lot anyway)

1. Fix error when trying to type any character within `[[]]` or `(())` (closes #315 )
2. Even after fixing the above, the block search function hangs the whole app when entering the first character within `(())` (perhaps because there are too many results?) This is a deeper issue to be investigated, but as a temporary workaround I have turned off `db/search-in-block-content`'s behavior for queries shorter than 4 characters. This way we can at least test the other behavior.
    - ROdZcWLH2F
3. Refactored query/results updating into its own function to make it a bit DRYer.
4. Since hitting backspace was not updating the inline results list, I fixed that too. Now backspace works much better.
    - Before:
    - AKiqrnAUIQ
    - After:
    - GVeQaHPPTI
5. Improved the Enter behavior when inside `[[]]` or `(())`, to move the caret directly to after the last bracket.
    - Before:
    - NEZwuKkhEv
    - After:
    - voIZ40i5Ix

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -501,7 +501,12 @@
 ;; Block state atom helpers
 
 (defn update-query
-  [state new-query query-fn]
-  (let [results (query-fn new-query)]
+  [state head key type]
+  (let [link-start (count (re-find #".*\[\[" head))
+        new-query (str (subs head link-start) key)
+        query-fn (cond
+                   (= type :block) search-in-block-content
+                   (= type :page) search-in-node-title)
+        results (query-fn new-query)]
     (swap! state assoc :search/query new-query)
     (swap! state assoc :search/results results)))
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -501,7 +501,8 @@
 ;; Block state atom helpers
 
 (defn update-query
-  [state head key type]
+  ([state head type] (update-query state head "" type))
+  ([state head key type]
   (let [query-fn (cond
                    (= type :block) search-in-block-content
                    (= type :page)  search-in-node-title)
@@ -511,4 +512,4 @@
         new-query (str (subs head link-start) key)
         results (query-fn new-query)]
     (swap! state assoc :search/query new-query)
-    (swap! state assoc :search/results results)))
+    (swap! state assoc :search/results results))))
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -502,11 +502,13 @@
 
 (defn update-query
   [state head key type]
-  (let [link-start (count (re-find #".*\[\[" head))
-        new-query (str (subs head link-start) key)
-        query-fn (cond
+  (let [query-fn (cond
                    (= type :block) search-in-block-content
-                   (= type :page) search-in-node-title)
+                   (= type :page)  search-in-node-title)
+        link-start (cond
+                     (= type :block) (count (re-find #".*\(\(" head))
+                     (= type :page)  (count (re-find #".*\[\[" head)))
+        new-query (str (subs head link-start) key)
         results (query-fn new-query)]
     (swap! state assoc :search/query new-query)
     (swap! state assoc :search/results results)))
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -503,13 +503,13 @@
 (defn update-query
   ([state head type] (update-query state head "" type))
   ([state head key type]
-  (let [query-fn (cond
-                   (= type :block) search-in-block-content
-                   (= type :page)  search-in-node-title)
-        link-start (cond
-                     (= type :block) (count (re-find #".*\(\(" head))
-                     (= type :page)  (count (re-find #".*\[\[" head)))
-        new-query (str (subs head link-start) key)
-        results (query-fn new-query)]
-    (swap! state assoc :search/query new-query)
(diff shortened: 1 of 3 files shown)


itsrainingmani:
Hi Marco,

Awesome first PR! 
I had one comment about 5. Roam moves the cursor to the end of the the `]]` when you hit enter regardless of the cursor position within the block.
I think that's currently not handled by your code. Do you think you'd be able to add that functionality in?

src/cljs/athens/keybindings.cljs
```diff
@@ -326,6 +327,14 @@
     ;;(= key-code KeyCodes.CLOSE_SQUARE_BRACKET)
 
 
+;; This should probably reside in another file as it has nothing to do
```
itsrainingmani, line 330: I think you could move this function `db.cljs` since that's where we're storing a lot of common functionality.

bonobo-d: I was wondering where would be the best place. This function is about updating the block's atom, not the datascript db, but if that's not a problem I'll move it to db.cljs.

bonobo-d:
@itsrainingmani
Thanks Mani, good point about pressing enter while in the middle of the link's string. Actually, even adding characters there will give unexpected behavior now.

If the objective is to reproduce Roam's behavior fully there, I think a deeper rewrite will be necessary. I'll think about it a little.

bonobo-d:
Made a few updates based on the feedback.

@itsrainingmani 
Just to confirm, is replicating the exact behavior of Roam a strict objective here? I don't have a Roam subscription to test now, but Obsidian for instance doesn't handle the Enter key in that way. When I start typing in the middle of the link string it will ignore anything that comes after it, and hitting Enter closes the `]]` and leaves whatever was after the caret there. Not great, but it might be acceptable depending on our target level of refinement.

If we want to do the most flawless behavior possible, I'll think about how to sophisticate the query handling.

src/cljs/athens/keybindings.cljs
tangjeff0: I think it's fine to refactor the logic into a function, but I would keep it within `keybindings` for now and not `db` for the reasons you state @bonobo-d

tangjeff0:
@bonobo-d we don't need to replicate Roam at this level of precision. If this is approximately what Obsidian does, it's probably good enough for now! Agree that the better UX would require a more thorough rewrite.

I'm hoping that a lot the complex keyboard write logic will be simplified upon using https://github.com/gadfly361/re-pressed or our own keyboard framework. Main idea is to use rules and centralize logic.

bonobo-d:
@tangjeff0 yeah I remember reading that the keybinding approach may change, so maybe it's better not to go too deep until that happens.

Right now it's still not exactly acting like Obsidian either, and typing in the middle of the string will give erratic behavior. I'll fix that in this PR.

bonobo-d:
@itsrainingmani @tangjeff0 I've rewritten how the query string is computed in a more flexible way, and also what pressing enter does when in the middle of the link. It now behaves like Obsidian does in this case (not deleting anything ahead of the caret).

EdrPgRDcuN

And the overall behavior should be much more robust now.

There are still quirks (e.g. characters not getting registered when typing fast) but I've already gone well past the scope I set for myself in this PR, so I'll leave them for now.

src/cljs/athens/db.cljs
```diff
@@ -496,3 +496,20 @@
   (->> (get-linked-references-by-block title)
        (remove #(= (:block/uid %) uid))
        count))
+
+
+;; Block state atom helpers
+
+(defn update-query
```
tangjeff0, line 503: Move to `keybinding.cljs`

src/cljs/athens/keybindings.cljs
```diff
@@ -204,29 +204,47 @@
...
+        tail-pattern (cond
+                       (= type :block) #"(\)\))?(.*)"
+                       (= type :page)  #"(\]\])?(.*)")
+        new-head (cond
+                   (= type :block) "$1(("
+                   (= type :page)  "$1[[")
+        closing-str (cond
+                      (= type :block) "))"
+                      (= type :page)  "]]")
+        new-str (clojure.string/replace-first head head-pattern (str new-head completed-str closing-str))
```
tangjeff0, line 223: require `clojure.string`

src/cljs/athens/db.cljs
bonobo-d: Done.

src/cljs/athens/keybindings.cljs
bonobo-d: Done.

PR #318, opened 2020-08-02
docs: update CONTRIBUTING with better starting instructions. update community roles

tangjeff0:
Are these instructions for getting started clearer? Will probably delete [README column of Project Board](https://github.com/athensresearch/athens/projects/2#column-9464291), then link to CONTRIBUTING from Top of "Good First Issues" column.

This doesn't include the resources for developers. It assumes that link to ClojureFam is a good entrypoint for learning.

@bonobo-d

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@
 
 ## Developers
 
-- Join our [Discord](https://discord.gg/GCJaV3V) 👾 and check out the `#engineering` and `#engineers` channels.
+- Join our [Discord](https://discord.gg/GCJaV3V) 👾 and check out the `#engineering` and `#engineers` channels. The former is for anything engineering-related, and the latter is for [contributors](./GOVERNANCE.md#contributors) who have already made non-trivial code contributions.
   - Post work updates in the `#build-in-public`. This keeps the team on the same page! Let's avoid stepping on each others toes, minimize blockers / dependencies, and cheer each other on!
 - Watch the repo and bookmark our [Project Board](https://github.com/athensresearch/athens/projects/2). This is the ultimate source of truth for product roadmapping.
 - To start working on your PR, you have a few ways to get started:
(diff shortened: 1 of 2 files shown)


bonobo-d:
@tangjeff0 looks great, I'm sure it will save some people some time having those instructions in that file. By the way, are there any branching/git workflow (apart from the commit category thing) conventions we're following? If so, it may be a good idea to put them there.

tangjeff0:
@bonobo-d we don't have any tips on branching workflows, but that would be really useful, especially if collaborators are working off each others' branches.

itsrainingmani:
@tangjeff0 @bonobo-d This is something that @alaq and I could add to the ClojureFam onboarding document and/or the Contributing doc since it seems like the information would be very useful. We had to work with each other's forks a couple of times during our Team Seneca PRs.

CONTRIBUTING.md
```diff
@@ -25,10 +28,29 @@
...
+## Developers
 
-- The best place to reach us is our [Discord](https://discord.gg/GCJaV3V)! 👾
+- Join our [Discord](https://discord.gg/GCJaV3V) 👾 and check out the `#engineering` and `#engineers` channels.
+  - Post work updates in the `#build-in-public`. This keeps the team on the same page! Let's avoid stepping on each others toes, minimize blockers / dependencies, and cheer each other on!
+- Watch the repo and bookmark our [Project Board](https://github.com/athensresearch/athens/projects/2). This is the ultimate source of truth for product roadmapping.
+- To start working on your PR, you have a few ways to get started:
+    1. ask a question in our `#engineering` [Discord](https://discord.gg/GCJaV3V) channel
+    1. comment on one of the existing top-level issues on the project board
+    1. create a PR draft or issue, then assign yourself (prefer drafts over new issues)
```
alaq, line 39: Are we trying to get people to open empty draft PRs? I feel like when you're ready to open a (not empty) draft, you're almost 50% of the way there, and some time may have past already.

CONTRIBUTING.md
```diff
@@ -25,10 +28,29 @@
 
 # Contributing to Athens
 
-Whether you are a designer, developer, or have other superpowers, please see our [v1 Project Board](https://github.com/athensresearch/athens/projects/2) to see what we're working on.
+## Developers
 
-- The best place to reach us is our [Discord](https://discord.gg/GCJaV3V)! 👾
+- Join our [Discord](https://discord.gg/GCJaV3V) 👾 and check out the `#engineering` and `#engineers` channels.
```
alaq, line 33: Do we want to add a little note on how the two channels are different from one another?

CONTRIBUTING.md
tangjeff0: I'm pretty okay with empty-ish PR drafts. That's the purpose of the line right below this. Some people might have a really clear intention of what they want to work on.

Also, creating a PR draft directly skips the need to open an issue as well. Perhaps slightly lighter weight!

alaq: Fine by me then! :)

PR #319, opened 2020-08-03
refactor: search functions

### LINKED ISSUE #269: Application becomes unresponsive after entering a short query in Athena
**Problem**

Entering a short query (one or two characters) in Athena will freeze the application. The only way out of it is to close the tab. You need to have data for this.

**Expected Behavior**

The application should remain responsive and show a list of results in Athena.

**Dependencies**

- Athens: latest master
- browser: Chromium, Firefox

**To Reproduce**

https://athensresearch.github.io/athens/ then load test data, input one or two characters in Athena. The application will freeze.

itsrainingmani:
Refactor the `search-in-node-title` and `search-in-block-content` functions to return only 20 items by default. This should fix #269.

* These functions also now take an optional parameter to specify max number of results

* Remove (take 20) from create-search-handler since those functions default to returning 20 results

Advantage of this refactor is that we can exert more control over these functions by specifying the maximum number of results returned by these functions.

There is still some latency when the search query is 1 or 2 characters long but the search functions do not hang anymore.

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -280,16 +280,18 @@
 
 
 (defn search-in-node-title
-  [query]
-  (d/q '[:find [(pull ?node [:db/id :node/title :block/uid]) ...]
-         :in $ ?query-pattern ?query
-         :where
-         [?node :node/title ?title]
-         [(re-find ?query-pattern ?title)]
-         [(not= ?title ?query)]] ;; ignore exact match to avoid duplicate
-       @dsdb
-       (re-case-insensitive query)
-       query))
+  ([query] (search-in-node-title query 20))
+  ([query n]
+   (->> (d/q '[:find [(pull ?node [:db/id :node/title :block/uid]) ...]
+               :in $ ?query-pattern ?query
+               :where
+               [?node :node/title ?title]
+               [(re-find ?query-pattern ?title)]
+               [(not= ?title ?query)]] ;; ignore exact match to avoid duplicate
+             @dsdb
+             (re-case-insensitive query)
+             query)
+        (take n))))
 
 
 (defn get-root-parent-node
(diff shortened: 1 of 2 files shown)


alaq:
This is great!
What's the logic for selecting these 20 items? Are they going to be the items the user will want to see?

src/cljs/athens/db.cljs
```diff
@@ -280,16 +280,18 @@
...
-  (d/q '[:find [(pull ?node [:db/id :node/title :block/uid]) ...]
-         :in $ ?query-pattern ?query
-         :where
-         [?node :node/title ?title]
-         [(re-find ?query-pattern ?title)]
-         [(not= ?title ?query)]] ;; ignore exact match to avoid duplicate
-       @dsdb
-       (re-case-insensitive query)
-       query))
+  ([query] (search-in-node-title query 20))
```
HaojiXu, line 283: Do we need another overload to find all results (e.g. for custom components/for toggling between top-20 and all?)

itsrainingmani:
> This is great!
> What's the logic for selecting these 20 items? Are they going to be the items the user will want to see?

@alaq It's whatever the first 20 items returned by the datascript query are. The call to `search-in-block-content` in `athena.cljs` does pretty much the same thing during `(take 20 (search-in-block-content query))`. I think the question of how to return the best or most relevant search results might be out of the scope of this PR.

src/cljs/athens/db.cljs
itsrainingmani: > I think this refactor makes sense in terms of performance - but I think another question concerning this UX-wise is whether we show only the top 20 results or show everything in the search dropdown from `keybinding.cljs`. What do you think @shanberg @tangjeff0 ?

@HaojiXu I think it makes sense to enforce a default cap on search results for any dropdown because depending on the search query, there could be hundreds of results and we probably don't want to display all of them to the user. And if we require more than 20 results, we can specify a higher max number of results.

HaojiXu: Definitely. I do believe that it might be a bit tricky to determine which results should be in the top 20, but just as you said, this is a question we can talk about later.

PR #320, opened 2020-08-03
chore(deps): update re-frame to 1.0.0

tangjeff0:
https://day8.github.io/re-frame/releases/2020/


itsrainingmani:
Should we also upgrade `reagent`, `shadow-cljs` and `ClojureScript` to the versions mentioned in the `re-frame` changelog?

tangjeff0:
Yeah! @itsrainingmani good catch. Hopefully this doesn't break anything :)

PR #322, opened 2020-08-05
refactor "scroll if needed" function

fnumatic:
- separate logic and effect
- scroll-top! can be registered as re-frame effect handler


fnumatic:
sth went wrong :(

PR #348, opened 2020-08-22
Proposed fix to on-change conflict

### LINKED ISSUE #340: New on-change behavior overwrites keydown results
**Problem**

Auto-pairing of braces and inline search isn't doing anything any more. I believe the problem is the following.

Commit 15de32ebf4df4233a379df5d184bb5fe22df53b5 introduces a function, `block-on-change`, which is fired with the textarea's `on-change` event and does this:

```cljs
(swap! state assoc :atom-string (.. e -target -value))
```

However, since this is happening *after* the effects of  `on-key-down`, it is overwriting any change to `:atom-string` that occurred there, including auto-pairing (and probably enter/tab behavior as well).

**Expected Behavior**

When typing `[` or `(` in a block, the corresponding closing brace should be created automatically.

**Dependencies**

- Athens: ab71dc0d474ee4e41567c7729eaf4e5077fd13d9
- browser: Chrome 84.0.4147.125
- JVM: Leiningen 2.9.3 on Java 1.8.0_252 OpenJDK 64-Bit Server VM
- node.js: v14.4.0

**To Reproduce**

1. Type `[` anywhere in a block.
2. No auto-pairing!

bonobo-d:
I'm proposing a way to fix #340. Now all functions that auto-generate text (auto-pairing, auto-complete, enter, etc.) in keybindings.cljs write to a new atom parameter `:generated-str` instead of directly to `:atom-string`. The job of writing to `:atom-string` is left to `blocks/block-on-change` instead. `blocks/block-on-change` now has some logic to switch between auto-generated text and event text from `on-change`.

This seems to fix all behavior I've tested (auto-completion, auto-pairing, copy&paste), but it does feel a bit hacky. Without a deeper understanding of the direction we want to take I can't come up with a better solution right now.

### DIFF
--- a/src/cljs/athens/keybindings.cljs
+++ b/src/cljs/athens/keybindings.cljs
@@ -217,7 +217,7 @@
         new-str     (str replace-str expand)]
     (swap! state merge {:search/index 0
                         :search/type nil
-                        :atom-string  new-str})))
+                        :generated-str  new-str})))
 
 
 (defn auto-complete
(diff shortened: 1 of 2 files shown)


tangjeff0:
After coding today, the only thing I noticed that doesn't work as expected is the default behavior for `ctrl-backspace`: delete back a word or to the closest prior punctuation. I'm on Ubuntu, but another thing I think might be `fn-backspace` which is `delete` on Mac.

bonobo-d:
@tangjeff0 both `ctrl-backspace` and `ctrl-delete` work well for me on Windows. Is it something you have already fixed, or a platform-dependent problem?

tangjeff0:
Fixed in #355 @bonobo-d !

PR #358, opened 2020-09-01
fix(inline): slash and search

tangjeff0:
- inline search auto-complete without results still moves caret
- typing after "/" without results hides slash menu
- `nth` over `get`
- proper expansion for "/" - remove query characters
- fix backspace bug where bksp with all text highlighted deleted block
- show lower opacity hint when no search results for inline
- [ ] modify `auto-complete-inline` and `auto-complete-slash` so they work on click, not just with enter
- [ ] there's a bug where after expanding, the first keypress right after doesn't change the textarea

### DIFF
--- a/src/cljs/athens/keybindings.cljs
+++ b/src/cljs/athens/keybindings.cljs
@@ -119,12 +119,12 @@
 (defn auto-complete-slash
   [state e]
   (let [{:keys [string/local] :search/keys [index results]} @state
-        {:keys [head]} (destruct-event e)
+        {:keys [head tail]} (destruct-event e)
         [_ _ expansion _] (nth results index)
         expand (if (fn? expansion) (expansion) expansion)
         start-idx (dec (count (re-find #".*/" head)))
         new-head (subs local 0 start-idx)
-        new-str     (str new-head expand)]
+        new-str (str new-head expand tail)]
     (swap! state assoc
            :search/type nil
            :string/generated new-str)))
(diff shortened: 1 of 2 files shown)


src/cljs/athens/keybindings.cljs
```diff
@@ -106,31 +106,39 @@
...
-        [_ _ expansion _] (nth slash-options index)
+  [state e]
+  (let [{:keys [string/local] :search/keys [index results]} @state
+        {:keys [head]} (destruct-event e)
+        [_ _ expansion _] (nth results index)
         expand (if (fn? expansion) (expansion) expansion)
-        replace-str (subs local 0 (dec (count local)))
-        new-str     (str replace-str expand)]
+        start-idx (dec (count (re-find #".*/" head)))
+        new-head (subs local 0 start-idx)
```
tangjeff0, line 126: Need `tail` as well.

src/cljs/athens/keybindings.cljs
```diff
@@ -106,31 +106,39 @@
...
+        new-str     (str new-head expand)]
```
tangjeff0, line 127: whitespace

src/cljs/athens/keybindings.cljs
```diff
@@ -106,31 +106,39 @@
...
-  (let [{:keys [start head tail target]} (destruct-event e)
-        {:search/keys [query type]} @state
+  [state e]
+  (let [{:search/keys [query type index results]} @state
+        {:keys [node/title block/uid]} (nth results index)
+        {:keys [start head tail target]} (destruct-event e)
+        completed-str (or title uid)
         block? (= type :block)
         page? (= type :page)
+        ;; rewrite this more cleanly
```
tangjeff0, line 141: Could this be rewritten with `condp->`?

PR #365, opened 2020-09-21
Split focused block into children on enter

theianjones:
Adds this functionality:

Screen Recording 2020-09-21 at 05 50 PM

### DIFF
--- a/src/cljs/athens/events.cljs
+++ b/src/cljs/athens/events.cljs
@@ -490,6 +490,31 @@
                                   :block/children reindex}]]]
           [:dispatch [:editing/uid new-uid]]]}))
 
+(defn split-block-to-children
+  "Takes a block uid, its value, and the index to split the value string.
+  It sets the value of the block to the head of (subs val 0 index)
+  It then creates a new child block with the tail of the string set as its value and sets editing to that block."
+  [uid val index]
+  (let [block (db/get-block [:block/uid uid])
+        head (subs val 0 index)
+        tail (subs val index)
+        new-uid (gen-block-uid)
+        new-block {:db/id        -1
+                   :block/order  (inc (:block/order block))
+                   :block/uid    new-uid
+                   :block/open   true
+                   :block/string tail}
+        reindex (->> (inc-after (:db/id block) (:block/order block))
+                     (concat [new-block]))]
+    {:fx [[:dispatch [:transact [{:db/id (:db/id block) :block/string head :edit/time (now-ts)}
+                                 {:db/id (:db/id block)
+                                  :block/children reindex}]]]
+          [:dispatch [:editing/uid new-uid]]]}))
+
+(reg-event-fx
+ :split-block-to-children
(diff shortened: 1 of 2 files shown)


src/cljs/athens/events.cljs
```diff
@@ -490,6 +490,31 @@
...
+                     (concat [new-block]))]
+    {:fx [[:dispatch [:transact [{:db/id (:db/id block) :block/string head :edit/time (now-ts)}
+                                 {:db/id (:db/id block)
+                                  :block/children reindex}]]]
+          [:dispatch [:editing/uid new-uid]]]}))
+
+(reg-event-fx
+ :split-block-to-children
+ (fn [_ [_ uid val index]]
+   (split-block-to-children uid val index)))
```
theianjones, line 517: I'm not sure if we want this to be registered as the other events in here. It just seemed more idiomatic to me than defining ad hoc events in the view file.

src/cljs/athens/events.cljs
```diff
@@ -490,6 +490,31 @@
...
+                   :block/order  (inc (:block/order block))
+                   :block/uid    new-uid
+                   :block/open   true
+                   :block/string tail}
+        reindex (->> (inc-after (:db/id block) (:block/order block))
+                     (concat [new-block]))]
+    {:fx [[:dispatch [:transact [{:db/id (:db/id block) :block/string head :edit/time (now-ts)}
+                                 {:db/id (:db/id block)
+                                  :block/children reindex}]]]
+          [:dispatch [:editing/uid new-uid]]]}))
```
theianjones, line 512: This does exactly the same thing as `split-block` except it sets the `new-block` as a child to `block`.

tangjeff0: I actually just implemented adding a new child in case that is more similar than `split-block`, see: https://github.com/athensresearch/athens/pull/366/files#diff-00d35e0f4787abcb3c17f4ebc0ce56b8R527-R537

Some orders are off:
- `new-block`'s should be `0`
- `(inc-after (:db/id block) (:block/order block))` should be `(inc-after (:db/id block) -1)`

src/cljs/athens/events.cljs
tangjeff0: Makes sense to me -- `indent` and `enter` are examples where there is 1 event that calls several use cases, but even then it might make more sense to give a unique event to each case. It makes tracking things in re-frame easier :)

tangjeff0 (changes requested):
Great job! Make sure to check orders by hovering over the bullet points :)

theianjones:
@tangjeff0 this is ready. Im not sure why the linter is complaining about `destruct-event` it is being pulled in from `athens.keybindings`

tangjeff0:
@theianjones I'm not sure either, that's really weird. Linter has no warnings when run locally.

tangjeff0:
Superceded by #378. Great work @theianjones!!!

PR #370, opened 2020-09-24
fix: get rid of `:string/generated` everywhere. use `nth` with not-found

tangjeff0:
When using `:string/generated` instead of `:string/local` for custom keypress behavior, this made it so that user had to press any key one time before it would change the textarea value again. Probably had to do with `:generated-str-listener`.

- `nth` takes an additional parameter when not found. very useful
- `re-matches` returns a vector instead of one value if capture groups are used&found

### DIFF
--- a/src/cljs/athens/keybindings.cljs
+++ b/src/cljs/athens/keybindings.cljs
@@ -276,7 +276,7 @@
 
 (defn handle-enter
   [e uid state]
-  (let [{:keys [shift ctrl start head tail value target]} (destruct-event e)
+  (let [{:keys [shift ctrl start head tail value]} (destruct-event e)
         {:search/keys [type]} @state]
     (.. e preventDefault)
     (cond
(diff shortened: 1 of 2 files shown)


src/cljs/athens/keybindings.cljs
```diff
@@ -416,16 +422,19 @@
   [e _ state]
   (let [{:keys [head key]} (destruct-event e)
         slash-key? (= key "/")
+        hashtag-key? (= key "#")
```
tangjeff0, line 425: Do hashtag stuff in a separate PR

src/cljs/athens/keybindings.cljs
```diff
@@ -416,16 +422,19 @@
...
         {:search/keys [type]} @state]
     (cond
       slash-key? (swap! state assoc
                         :search/index 0
                         :search/query ""
                         :search/type :slash
                         :search/results slash-options)
+      hashtag-key? (swap! state assoc :search/type :page :search/query "")
```
tangjeff0, line 433: see above

src/cljs/athens/keybindings.cljs
```diff
@@ -392,10 +396,12 @@
...
                                                      (swap! state assoc
                                                             :search/type nil
-                                                            :string/generated new-str)
+                                                            :string/local new-str)
                                                      (js/setTimeout #(setCursorPosition target (dec start)) 10))
       ;; slash: close dropdown
-      (= "/" (last value)) (swap! state assoc :search/type nil)
+      (= "/" look-behind-char) (swap! state assoc :search/type nil)
+      ;; hashtag: close inline-search
+      (= "#" look-behind-char) (swap! state assoc :search/type nil)
```
tangjeff0, line 404: see below

PR #373, opened 2020-09-24
feat(electron): @DeLaGuardo/setup-clojure works for windows now!

tangjeff0:
- source code was updated but waiting for official release


tangjeff0:
Doesn't actually work until `tool-deps` is implemented as well. Closing for now https://github.com/DeLaGuardo/setup-clojure/blob/0ba662eae20a54f4b47d4094c3f23165b383c6a5/src/setup-clojure.ts#L34-L39

PR #404, opened 2020-10-05
feat(parser): italics and raw urls

tangjeff0:
- italics fails tests because links can have `_`, e.g. `#a_hashtag` or `[[a_link]]`
- raw url fails if there are characters before the start of the link.
  - `https://google.com foo` works
  - `foo https://google.com` does not work
- using regex from https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url


tangjeff0:
stalled

PR #452, opened 2020-10-21
aria: ui suggestions, keyboard navigation, auto-complete

tangjeff0:
For @AriaMinaei:

- `alt-left` and `alt-right` to navigate (Chrome shortcuts)
- toolbar is flush with top
- hide filter icon because feature not implemented yet
- hashtag auto-complete doesn't delete result
- inline search allows for exact match `#L1` `#L2`
- TODO: links that span multiple lines make a weird box
- previous PR fixes paste being out of order


AriaMinaei:
This is EPIC! Can't believe how fast these suggestions landed in code 🤯

tangjeff0:
Should see the updates in next release, probably coming later today!

PR #453, opened 2020-10-21
WIP: add filters for linked refs

tangjeff0:


tangjeff0:
Closing - filters are quite intense and require more discussion/design. Consider sorting as well - this is essentially a list of data. Think Notion/Airtable tables.

PR #475, opened 2020-11-01
feat: hide electron title bar

shanberg:
- Uses 'hidden inset' titlebar style for more integrated appearance.
- Makes app header a window drag handle
- Fixes missing background color in appheader controls

Screen Shot 2020-11-01 at 11 27 59 AM


tangjeff0:
This is what it looks like on Ubuntu :) 

Wonder what it looks like on Windows.

image

shanberg:
Yeah... seems like it definitely needs to account for operating system. Probably want to solve that with a class on the page body.

PR #489, opened 2020-11-15
docs: Added documentation for vim-fireplace

zekzekus:
Added cider-nrepl as a development dependency


tangjeff0:
Great job @zekzekus !

Issue #447, opened 2020-10-19
fix(ui): Blank line ends up inline with next block/subblock


seltzered:
**Problem**

For a given block, when creating a second blank line it will appears inline with a block or sub-block below it until you write a non-whitespace character.
(this also occurs when having a null page title)

**Expected Behavior**

A new blank line should create a new line / visual spacing for it.

**Dependencies**

- Athens: 1.0.0-beta15
- browser: macOS electron build
- JVM:
- node.js: 

**To Reproduce**

shift+enter on a block to create a new line, particularly one where you have some content underneath. Before typing on the second line observe that it's inline until you write text.

**Screenshots**

![2020-10 athens hidden newline](https://user-images.githubusercontent.com/297435/96401875-69e8fa80-1189-11eb-8ea5-4a49c5b04a9d.gif)

tangjeff0:
https://www.loom.com/share/124aff147a7649b68e3b3779e66ecbbb

Issue #460, opened 2020-10-24
fix(ui): textarea and underlying html element have different sizes


tangjeff0:
At the block page:
https://www.loom.com/share/40faea89648f4e8aa15967637a83b0f5

At the node page:
https://www.loom.com/share/b44a57da61ff4a718a8197966eae2160

At the block level, which is most likely the same issue as #447 
https://www.loom.com/share/7962e880a2c048729fed146384f7745f

PR #491, opened 2020-11-15
refactor(blocks): better positioning method for block textareas

shanberg:
Should fix a few cases where textareas broke out of their containers and overlapped sibling blocks.

Before:
- The `block-content` container creates a new [positioning context](https://alistapart.com/article/css-positioning-101/).
- Rendered content is naturally placed in that context, and defines the size of its container.
- The textarea is absolutely positioned to fit the space created by the rendered content.
This means only the rendered/parsed content affects the height of the block. If the textarea content is larger than the rendered content (as is often the case, given the syntax for adding markdown links, etc.), the textarea will be cut off or overflow onto following blocks.

After:
- The `block-content` container creates a new grid area, with a single grid cell.
- Rendered content is placed within that grid cell, enlarging the size of the cell to its size.
- The textarea is also placed within the same grid cell, enlarging the size of the cell to its size, if its size is larger than the rendered content.
This means both the rendered content _and_ the textarea can enlarge the size of the block, and the block will always be the size of the larger of the two, _and_ the textarea and the rendered content will also stretch larger to fit the size of the block, if necessary.


tangjeff0:
> Should fix a few cases where textareas broke out of their containers and overlapped sibling blocks.

Which cases? Doesn't seem like this fixes the cases mentioned in #460 and #447.

shanberg:
@tangjeff0 Initially this just addressed regular blocks, but I just updated the title blocks too.

My testing indicates this fixes #447 and #460.

tangjeff0:
#536 implements just the changes to modal and blocks, not titles. Wasn't able to fix the flicker, but wanted to implemented block changes. Thank you @shanberg!

PR #499, opened 2020-11-18
fix: missing cursor issue #464

### LINKED ISSUE #464: Cursor gets lost after using uparrow on root/topmost node of a page.

**Problem**

See screenshot. When hitting the up arrow on the root block (going from a non-zero character position within the same block), the cursor disappears. Sometimes you can just hit down arrow to recover, in some cases, it seems to not recover until mousing over. 

Behavior may change depending on whether focused on a single page vs. on the 'daily notes' view.

**Expected Behavior**

Should just go to the starting position of the block, then possibly the page title if up arrow hit again.

**Dependencies**

[ Please specify your environment and software dependencies, with versions. ]

- Athens: 1.0.0-beta19
- OS: macOS 10.15.17

**To Reproduce**
Mouseover to text somewhere in root block, hit up arrow.

**Screenshots**
2020-10 athens cursor

juniusfree:
fix #464 

I found other issues that might be related to this but I'm not sure if they should be included here.
- arrow down is not working when the cursor is on the top node
- the cursor position is inconsistent when moving up and down the blocks using the arrow key

### DIFF
--- a/src/cljs/athens/views/node_page.cljs
+++ b/src/cljs/athens/views/node_page.cljs
@@ -385,6 +385,7 @@
           (when-not timeline-page?
             [autosize/textarea
              {:value         (:title/local @state)
+              :id            (str "editable-uid-" uid)
               :class         (when (= editing-uid uid) "is-editing")
               :on-blur       (fn [_] (handle-blur node state ref-groups))
               :on-key-down   (fn [e] (handle-key-down e state))


juniusfree:
It looks like I misunderstood the issue. I'll update this.

tangjeff0:
@juniusfree it looks good to me, just figuring out why the tests are failing (not your fault).

juniusfree:
@tangjeff0 I just want to clarify your comment at https://github.com/athensresearch/athens/blob/74d132262dc3de3101c71d573a99ec667c4038a6/src/cljs/athens/views/node_page.cljs#L384

Should the user be able to edit the title page?

tangjeff0:
@juniusfree I see the confusion - that comment is unclear. "Prevent editing" only applies to date pages, which can show up in multiple places.

Issue #501, opened 2020-11-21
fix: the arrow down key is not working when the cursor is on the page title.


juniusfree:
*If you aren't sure about your bug, you can always join our [Discord](https://discord.gg/HNmxvpm) and ask in #agora or #engineering!*

**Problem**

The arrow down key is not working when the cursor is on the node/block page title.

**Expected Behavior**

Given the cursor is on the node/block page title, when the user presses the arrow down key, then the cursor should go to the first child block.

**Dependencies**

- Athens: version=1.0.0-beta.21
- browser: electron
- JVM: openjdk 11.0.9.1 2020-11-04
- node.js: v14.4.0

**To Reproduce**

1. Click on the node block / page title
2. Press the arrow down key

**Screenshots**

![title](https://user-images.githubusercontent.com/8164667/99862611-6526c480-2bd5-11eb-9c20-e6e29aca65b6.gif)

tangjeff0:
@juniusfree also make sure this works for block pages too!

juniusfree:
@tangjeff0 noted

PR #505, opened 2020-11-29
fix: arrow down key function in page title and block page title #501

juniusfree:
Fix for #501

Page Title
page-title

Block Page Title
block-page-title

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -463,8 +463,9 @@
   [uid]
   (loop [uid uid]
     (let [sib    (nth-sibling uid +1)
-          parent (get-parent [:block/uid uid])]
-      (if (or sib (:node/title parent))
+          parent (get-parent [:block/uid uid])
+          {node :node/title}   (get-block [:block/uid uid])]
+      (if (or sib (:node/title parent) node)
         sib
         (recur (:block/uid parent))))))
 
@@ -478,10 +479,10 @@
   ([uid]
    (let [block                (->> (get-block [:block/uid uid])
                                    sort-block-children)
-         {:block/keys [children open]} block
+         {:block/keys [children open] node :node/title } block
          next-block-recursive (next-sibling-recursively uid)]
      (cond
-       (and open children) (:block/uid (first children))
+       (and (or open node) children) (:block/uid (first children))
        next-block-recursive (:block/uid next-block-recursive))))
   ([uid selection?]
    (if selection?
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -479,7 +479,7 @@
   ([uid]
    (let [block                (->> (get-block [:block/uid uid])
                                    sort-block-children)
-         {:block/keys [children open] node :node/title } block
+         {:block/keys [children open] node :node/title} block
          next-block-recursive (next-sibling-recursively uid)]
      (cond
        (and (or open node) children) (:block/uid (first children))
--- a/src/cljs/athens/views/block_page.cljs
+++ b/src/cljs/athens/views/block_page.cljs
@@ -2,7 +2,7 @@
   (:require
     ["@material-ui/icons" :as mui-icons]
     [athens.db :as db]
-    [athens.keybindings :refer [destruct-key-down]]
+    [athens.keybindings :refer [destruct-key-down textarea-key-down]]
     [athens.parse-renderer :as parse-renderer]
     [athens.router :refer [navigate-uid]]
     [athens.style :refer [color]]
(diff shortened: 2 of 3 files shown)


juniusfree:
@tangjeff0 

1. I'm not entirely sure with this particularly with the use of `textarea-key-down` so let me know your thoughts on that.
2. Shift+Enter broke on this. Should I create a separate PR for that or should I include it here?

Thanks.

tangjeff0:
Wow, I'm honestly so impressed with the progress you're making!

The code in `db.cljs` looks promising.

I see you replaced `handle-key-down` with `textarea-key-down`. This is a good intuition, but basically we just need to borrow parts of `textarea-key-down` into `handle-key-down` rather than using it entirely. The reasoning here is that we only need partial arrow-key functionality, but `textarea-key-down` does _a lot_ of stuff for blocks that we don't need.

- Copy a lot of the code from `handle-arrow-key`. https://github.com/athensresearch/athens/blob/00a6f3a2e997fd3389dcd96238fb53bcbdad319e/src/cljs/athens/keybindings.cljs#L299-L308 https://github.com/athensresearch/athens/blob/00a6f3a2e997fd3389dcd96238fb53bcbdad319e/src/cljs/athens/keybindings.cljs#L336-L341
- Copy the `get-caret-position` logic. You're using atoms and state here, which might be new to you. https://github.com/athensresearch/athens/blob/00a6f3a2e997fd3389dcd96238fb53bcbdad319e/src/cljs/athens/keybindings.cljs#L595-L596

Let me know if you need help or have questions. Great work @juniusfree .

juniusfree:
@tangjeff0 Thanks for the detailed feedback. I'm learning a lot from this.

juniusfree:
@tangjeff0 Code updated. Let me know if I get it right or not. Thanks.

tangjeff0:
Great job @juniusfree. At this point, I would actually just import `handle-key-down` on `node-page.cljs` into `block-page.cljs`, then delete the redundant functions on `block-page.cljs`. This is a more direct example of reusing the same function in multiple places.

Two things:
- Why did you modify CSS?
- Make sure the GitHub Actions tests pass next. See https://github.com/athensresearch/athens/blob/master/CONTRIBUTING.md#running-ci-scripts-locally

juniusfree:
Noted on the tests and `handle-key-down`. I'm still trying to figure out when to reuse a function or when to write a new one. Any tips on that?

>     * Why did you modify CSS?

`span.offsetTop` for the title is 0...
https://github.com/athensresearch/athens/blob/a1e7a304d47610ea4e5ceca45990cd770b23c6b0/resources/public/textarea.js#L135

...which resulted in the `:down` event not being called
https://github.com/athensresearch/athens/blob/00a6f3a2e997fd3389dcd96238fb53bcbdad319e/src/cljs/athens/keybindings.cljs#L339-L341

The problem why `span.offsetTop` is 0 is because the `textarea` element has the same height as the `h1` element. So I added a line-height for `h1` to make it taller than the `textarea`. I adjusted the margin as well so that the total height (content+padding+margin) of the `h1` does not change.

juniusfree:
@tangjeff0 Code updated. Thanks for your patience and feedback. 😅

1. I used the `handle-key-down` in block_page.cljs
2. The `handle-enter` needs to be moved to node_page.cljs because `handle-key-down` is using it. 
Previously, pressing the enter key on the page title doesn't do anything. Now, it creates a new child block because of `handle-enter`. Is it the correct behavior for that?

tangjeff0:
> I'm still trying to figure out when to reuse a function or when to write a new one. Any tips on that?

@juniusfree this is an excellent question, was hoping you would ask that. In this case, I thought the code would be exactly the same, so we could use the same function. In the previous case, the code was similar, but not enough so to use the same function.

Eventually, you build an intuition for it, but for where you're at it's just good to ask :)

> Previously, pressing the enter key on the page title doesn't do anything. Now, it creates a new child block because of handle-enter. Is it the correct behavior for that?

This is incorrect behavior. I forgot that node-page and block-page are slightly different for `enter` behavior.

juniusfree:
@tangjeff0 I'll try to figure out your update. 😁 Do I need to do something on my end? And as always, thanks for the feedback.

tangjeff0:
@juniusfree You can accept the PR haha. Or you can try to copy it locally and play around with it.

juniusfree:
@tangjeff0 #501 approved. 😅 Closing this one.

tangjeff0:
@juniusfree I think if you approved it, it would just update this branch, so I don't think closing was necessary

juniusfree:
@tangjeff0 I was looking at the wrong page. PR merged.

tangjeff0 (approved):

tangjeff0:
Great job on PR 2 @juniusfree :muscle:!

juniusfree:
Thanks @tangjeff0!

PR #508, opened 2020-11-30
feat(auto-complete): pressing [[ or (( on selected text opens dropdown

tangjeff0:
`update-query` has specified behavior for backspace, but this doesn't work if user does ctrl-backspace, which deletes an entire word, or if user backspaces on selected text, which deletes all the text.

Also need to not close out of dropdown if user presses left or right and is still within the brackets/parens.


tangjeff0:
Can return to this later as part of a larger rework on keybindings.

PR #520, opened 2020-12-28
Add ctrl-down and ctrl-up shortcuts to open/close blocks

visika:
The desired event is triggered, however it is not perfect:
as the block if opened and closed, the editing pointer moves around.
The movement of the pointer should be freezed as an improvement.

### DIFF
--- a/src/cljs/athens/keybindings.cljs
+++ b/src/cljs/athens/keybindings.cljs
@@ -292,7 +292,7 @@
 
 (defn handle-arrow-key
   [e uid state]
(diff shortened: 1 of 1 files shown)


tangjeff0:
Great job @visika. I refactored the `cond`, then added `e preventDefault` to stop the cursor from moving unnecessarily. Thank you and congrats on your first contribution!

visika:
Thanks for helping me with this. I noticed that is ctrl + left/right crashes the app unfortunately. I should remedy to this soon.

PR #522, opened 2020-12-29
fix(#520): ctrl-left and right need to do nothing

visika:
Ctrl+left/right does not crash the app anymore.


tangjeff0 (approved):

tangjeff0:
Thanks for this @visika . When we use a real keybinding library, specifying nil behavior won't be necessary!

PR #539, opened 2021-01-09
fix: width & height of slash cmd and inline search

### LINKED ISSUE #459: fix(ui): too tall and too wide when opened near right side of block
**slash commands**
image

**inline search**
image

---

**should be**
image

juniusfree:
fix #459

image

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,29 +7,8 @@ assignees: ''
 
 ---
 
-*If you aren't sure about your bug, you can always join our [Discord](https://discord.gg/HNmxvpm) and ask in #agora or #engineering!*
-
 **Problem**
 
-[ Please provide a short and concise description of the bug. ]
+**Screenshots/Demo**
 
-**Expected Behavior**
-
-[ What is the behavior you expected to see? ]
-
-**Dependencies**
-
-[ Please specify your environment and software dependencies, with versions. ]
-
-- Athens:
-- browser: 
-- JVM:
-- node.js: 
-
-**To Reproduce**
-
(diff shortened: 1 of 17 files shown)


tangjeff0:
Hey @juniusfree, this isn't exactly a great fix. The reason is that the `min-width` makes the width too wide if the text within the dropdown is less than `16em`:
image


The problem with the current implementation in `master` seems to stem from the fact that right side of the dropdown can't extend beyond the block container. Does that sound right @shanberg?

tangjeff0:
Don't get too stuck on this @juniusfree. CSS can be very finicky :)

juniusfree:
@tangjeff0 Yes. CSS is confusing sometimes. 😅

> The problem with the current implementation in master seems to stem from the fact that right side of the dropdown can't extend beyond the block container.

Do we have the design for this so I can see what it looks like? Thanks.

tangjeff0:
To be honest, I don't really know why this happens, and my CSS-fu isn't great. But do you know notice how the width of the dropdown is different once the cursor position gets closer to the right edge?

image

image


image

juniusfree:
OK. This is how it looks with 16em. 

image

But as you mentioned, the problem with min-width is that it won't look good for short text.

image

Maybe I'll try to apply a different min-width for each dropdown.

tangjeff0:
@shanberg can you help us understand what is going on with the width as the dropdown approaches the right edge of the page? I'm guessing it has something to do with nested `position`.

In https://github.com/athensresearch/athens/issues/459#issuecomment-720116279, you mentioned React portal as a possible solution. I implemented portal in some components, but it was problematic for some reasons that I forget, so I reverted it #538. Maybe we can all jam on this sometime?

tangjeff0:
Closing for now!

PR #288, opened 2020-07-20
feat(import): roam import first pass

tangjeff0:
Datascript to datascript import, using Roam's `window.roamAlphaAPI.q`.

Next steps:

- implement export from Roam using roam-toolkit, see #31 
- create desktop app that can access filesystem. user flow:
  - upload one file as the db for Athens (for now)
  - remember the file path
  - when db is updated, write file to filesystem
  - if file is updated (e.g. via Dropbox), update application db

PR #561, opened 2021-01-21
feat(import): Roam import

tangjeff0:
Picking up from #288 

- had to change roam dates to athens dates, e.g. January **1st**, 2021 to January **1**, 2021, for links and titles
- had to fix linked refs to use query links by :block/_refs rather than with regex


src/cljs/athens/events.cljs
```diff
@@ -39,6 +39,133 @@
...
+                                          string (assoc :block/string (athens.patterns/replace-roam-date string))
+                                          title (assoc :node/title (athens.patterns/replace-roam-date title))))
+                                date-concat)]
+    ;;tx-data))
+    (d/db-with db tx-data)))
+
+;;(/ 3736 3842) 97% clean
+;;(-> (- 1056 2)
+;;    (+ (- 3088 406))))
+;;(defonce ROAM-DB (atom nil))
```
avichalp, line 153: 🤔   This line of comment might not be intentional?

Screen Shot 2021-01-29 at 5 54 46 PM

tangjeff0: Yes not intentional

avichalp, line 153:
```suggestion
(defonce ROAM-DB (atom nil))
```

avichalp: Great, the roam import worked for me after uncommenting the line.

tangjeff0: How's performance? What is your `index.transit` size now? @avichalp

avichalp:
```yaml
index.transit: 4.1M
roam export EDN file`: 7.2M
```

There is a high latency in loading search results and in loading pages with a high number of linked blocks (~ 400+). From my perception, latency is as high as 10-11 seconds sometimes. Would like to know how is the performance for other people. If we were to optimize performance, the first step would be to collect some data, I guess.

tangjeff0: My guess is that this stems from inefficient posh queries. Roam and LogSeq teams both told me that Posh wasn't super performant, so they both ended up writing their own reactive lightweight wrappers around Datascript.

I also remember that @jeroenvandijk was working on performance optimizations a while ago, using Reagent cursors. I wonder if that is part of the solution. https://github.com/athensresearch/athens/pull/93/files#diff-3c7f15f69987f2ac41d3dfa65d60e8dfae0778e494b2207f64721d20baad680cR43-R46


Current performance issue: #570

src/cljc/athens/patterns.cljc
```diff
@@ -16,3 +16,26 @@
    Lookarounds don't consume characters https://stackoverflow.com/questions/27179991/regex-matching-multiple-negative-lookahead "
   [string]
   (re-pattern (str "(?i)(?<!#)(?<!\\[\\[)" string "(?!\\]\\])")))
+
+
+;; Positive Lookbehind: between 1 and 2 digits
+;; One of an oridinal suffix, e.g. -st, -nd, -rd, -th, see https://en.wikipedia.org/wiki/Ordinal_indicator
```
agentydragon, line 22: typo: "oridinal"

tangjeff0:
Depends on #665

src/cljs/athens/db.cljs
```diff
@@ -234,18 +234,32 @@
       (conj :node/title :page/sidebar)))
 
 
+(def roam-node-document-pull-vector
+  '[:node/title :block/uid :block/string :block/open :block/order {:block/children ...}])
```
baristuncay, line 238:
```suggestion
  '[:node/title :create/time :edit/time :block/uid :block/string :block/open :block/order {:block/children ...}])
```

Create and edit time was not included in import

PR #574, opened 2021-01-27
feat(electron): Use settings.json for separate build channels

tangjeff0:


tangjeff0:
Closing - going to turn off auto-update and then let users download previous versions.

PR #596, opened 2021-02-02
feat(electron): eliminate common race condition, create n backups

tangjeff0:
Based off of [recommendation by @pithyless](https://github.com/athensresearch/athens/discussions/590#discussioncomment-329333), but uses `copyFile` instead of `rename` to persist a backup.


lambduhh (approved):
lgtm

PR #601, opened 2021-02-03
fix(events): Check for stale parent when deleting block (#567)

### LINKED ISSUE #567: error when fastly deleting many single characters
**Problem**

When trying to delete many characters at a time, error message appeared.

**Expected Behavior**

Deleting the characers (maybe even speed up when staying on key for a while...

**Dependencies**

[ Please specify your environment and software dependencies, with versions. ]

- Athens: beta 1.00 0.30
- local electron app
- JVM:
- node.js:  14.15.4
- OS: Linux Pop_OS

**To Reproduce**

Deleting many single characters (staying on delete-key or hitting delete-key multiple times in a fast order...)

**Screenshots**
 
image

pithyless:
Holding down backspace when deleting text can enqueue multiple `:backspace` reg-event-fxes. The first will correctly delete and merge the parent blocks, but the next events in the queue will still be processed.

This is a surgical fix that just checks for an invalid parent and fixes #567, but a more heavy-handed approach may be appropriate to handle other unreported edge-cases of event queueing.

### DIFF
--- a/src/cljs/athens/events.cljs
+++ b/src/cljs/athens/events.cljs
@@ -502,6 +502,7 @@
 
 (defn backspace
   "If root and 0th child, 1) if value, no-op, 2) if blank value, delete only block.
+  No-op if parent is missing.
   No-op if parent is prev-block and block has children.
   No-op if prev-sibling-block has children.
   Otherwise delete block and join with previous block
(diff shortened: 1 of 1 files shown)


tangjeff0:
Thanks @pithyless! A similar problem also existed for `:enter`, where if a user held on to enter or pressed it many times, the new block orders would get messed up. I specifically throttled this as a temporary fix:

https://github.com/athensresearch/athens/blob/92d482e3e2b1e091ac94e0b56f7551a4a8ae3633/src/cljs/athens/keybindings.cljs#L380-L404

PR #617, opened 2021-02-08
fix: open web links using default browser

### LINKED ISSUE #548: Open web links using default browser

**Problem**

Currently opening a web link (i.e. starting with `http` or `https`) from athens causes it to open in a new electron window, instead of my default browser.

**Expected Behavior**

Web links are opened in my default browser.

**Dependencies**

Arch linux.

- Athens: 1.0.0-beta.25
- browser: Firefox 84.0.2
- JVM: unsure, please let me know if this is relevant and I'll look into it
- node.js: unsure, please let me know if this is relevant and I'll look into it

**To Reproduce**

- Add a web link to a block (e.g. `[athens](https://github.com/athensresearch/athens/)`)
- Go to a new block, such that the link renders as something you can click
- Click the link

**Screenshots**

Please let me know if a screenshot is needed.

juniusfree:
fix #548

works in complete url (e.g. "https://www.google.com" but not "google.com")


tangjeff0:
Is it possible to do this without dependencies? e.g. https://github.com/electron/electron/issues/1344

juniusfree:
@tangjeff0

For this one, I copied these:
https://github.com/electron/electron/issues/1344#issuecomment-150031171
https://github.com/electron/electron/issues/1344#issuecomment-171516261

I'll take a look at the other solution.

juniusfree:
@tangjeff0 Updated. The dependency is removed.

tangjeff0:
I fixed some syntax errors:

1. `..` at beginning
2. `..` at end
3. paren around `open`

Curious, were you able to build this locally?

This is also a good resource for converting cljs to js: http://app.klipse.tech/

juniusfree:
> Curious, were you able to build this locally?

@tangjeff0 I thought it's a valid syntax. I had the `..` too before I updated it to `.` 

https://cljs.github.io/api/syntax/#dot
https://lwhorton.github.io/2018/10/20/clojurescript-interop-with-javascript.html#function-invocation

But upon rechecking, the renderer build has no error with `.` that's why it works locally but the main build has an issue with that syntax.

> This is also a good resource for converting cljs to js: http://app.klipse.tech/

🤦‍♂️

tangjeff0:
@juniusfree so I believe the issue is that the single `.` dot invocation only works when there is just one level of indirection, e.g. `(. js/window fn)` works but `(. js/window -webContents fn)` wouldn't work. In the second case, you need double dot, e.g. `(.. js/window -webContents fn)`

juniusfree:
@tangjeff0 OK noted on that one.

I'm running those commands in Calva but I didn't check the logs. I'll be always checking that next time.

Thanks for the feedback.

PR #624, opened 2021-02-09
WIP: real-time collaboration

tangjeff0:
Some conversation on Discord: https://discord.com/channels/708122962422792194/714804822909517935/808779175413678171

Value:
1. Requested by our potential first enterprise customer
2. Enables Athens team to use a db together to share knowledge

Specs
- uses filesystem to communicate real-time changes
	- therefore requires a filesync service like Dropbox or Syncthing
- on startup, read `db/path/index.transit`
- on change, update in-memory db and write a log to `db/path/tx-logs/{timestamp}-{os-username}.log`
	- edge case: same os-username, different computers/clients
- watch `db/path/tx-logs/`. on new log from another username, update in-memory db
- when no new logs detected for 60 seconds, write `{timestamp}-index.transit.bkp` file and copy to `index.transit`, same as current saving flow
	- but when the new db is written to client A and then db is synced, client B believes that `index.transit` is wrong, and overwrites the changes that just happened

### DIFF
--- a/src/cljs/athens/effects.cljs
+++ b/src/cljs/athens/effects.cljs
@@ -224,11 +224,16 @@
         (pprint more-tx-data)
         (prn "TX FINAL INPUTS")                             ;; parsing block/string (and node/title) to derive asserted or retracted titles and block refs
         (pprint final-tx-data)
-        (let [outputs (:tx-data (transact! db/dsdb final-tx-data))]
-          (ph-link-created! outputs)
-          (prn "TX OUTPUTS")
-          (pprint outputs))))
-
+        (let [outputs (:tx-data (d/with @db/dsdb final-tx-data))]
+           (ph-link-created! outputs)
+           (prn "TX OUTPUTS")
+           (pprint outputs)
+           (reset! a outputs)
+           (dispatch [:fs/write-log outputs]))
+        #_(let [outputs (:tx-data (transact! db/dsdb final-tx-data))]
+            (ph-link-created! outputs)
+            (prn "TX OUTPUTS")
+            (pprint outputs))))
     (catch js/Error e
       (js/alert (str e))
       (prn "EXCEPTION" e))))
--- a/src/cljs/athens/effects.cljs
+++ b/src/cljs/athens/effects.cljs
@@ -228,7 +228,7 @@
            (ph-link-created! outputs)
            (prn "TX OUTPUTS")
            (pprint outputs)
-           (reset! a outputs)
+           ;;(reset! a outputs)
            (dispatch [:fs/write-log outputs]))
         #_(let [outputs (:tx-data (transact! db/dsdb final-tx-data))]
             (ph-link-created! outputs)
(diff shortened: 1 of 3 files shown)


jeroenvandijk:
> edge case: same os-username, different computers/clients

What about creating a unique id (uuid) the first time Athens starts and use that to identify the client? The username could be added for debugging purposes. If the string is too long you could hash the uuid e.g. `jeroen-923flsaf`.

tangjeff0:
After working on this implementation deeply for a week, I've concluded that it's too difficult at this time to resolve conflicts with atomic append-only tx logs over a synced filesystem.

- Each client can have differing tx logs, which might not be valid transactions on another db.
- Each client can also write to `index.transit`, which is currently the final source of truth
- But each client can also have it's own in-memory db state, which can differ from filesystem db state.

This is `n*3` different states. There are edge cases where each state could be considered "authoritative": the tx log, the in-memory db, or the fs db.

That's not to say tx logs are impossible as an eventual solution, but that deeper design is needed to implement a solution a feel is safe for users. Roam has notably has the worst data loss and availability issues it's had in over a year, in the past 2 weeks.

I do believe stability and longevity should be prioritized when it comes to user data.

jeroenvandijk:
I can imagine this is a hard problem to tackle. You mention:


> Each client can have differing tx logs, which might not be valid transactions on another db.

Do you have examples of invalid transactions?


> There are edge cases where each state could be considered "authoritative"

I'm curious to learn more about these edge cases. 

I'm also wondering if `index.transit` could be treated as a cached materialized view fo the tx-logs and only be used as a cache during the initialisation of Athens. It would then be updated every X transactions

tangjeff0:
> Do you have examples of invalid transactions?

The way it was written in this branch, we save the final raw output datoms, then transact them to other clients on read. However, when raw datoms are directly transacted, [they are "trusted" and not validated](https://cljdoc.org/d/datascript/datascript/1.0.3/api/datascript.core#init-db). Therefore if an entity id exists on one db but not the others, an error is thrown. Conor told me to always use `:block/uid` for transactions over websockets, which implies entity ids can get out of sync.

Example of a transaction:

```
[#datascript/Datom [100 :block/order 1 536893313 true]
 #datascript/Datom [100 :block/uid "0d0891ed2" 536893313 true]
 #datascript/Datom [100 :block/open true 536893313 true]
 #datascript/Datom [100 :block/string "" 536893313 true]
 #datascript/Datom [100 :block/children 38041 536893313 true]]
```

> I'm curious to learn more about these edge cases.

1. Filesystem is correct store: on startup
2. Append-log is correct store: when a client reads an append log from another client
3. In-memory  is correct store: when a user writes to their application

Finally, you can run two instances of Athens on two different computers with the branch `tangjeff0:append-log`, and try making multiple changes at the same time to test some of this behavior out. @jelmerderonde

tangjeff0:
@tomisme mentioned https://braid.news/ here

PR #627, opened 2021-02-10
fix(block-page):577 Add a DS tranaction to header when onblur

lambduhh:
If user clicks out of header area (on-blur) after making changes to text, a datascript transaction will log and record the change. 

Steps to replicate:
- Open header
- Type changes
- Navigate/click elsewhere and view console


src/cljs/athens/views/block_page.cljs
```diff
@@ -87,20 +96,23 @@
...
           [breadcrumbs-list {:style {:font-size "1.2rem"}}
            (doall
-             (for [{:keys [node/title block/uid block/string]} parents]
-               ^{:key uid}
-               [breadcrumb {:key (str "breadcrumb-" uid) :on-click #(navigate-uid uid)}
-                (or title string)]))]]
+            (for [{:keys [node/title block/uid block/string]} parents]
+              ^{:key uid}
+              [breadcrumb {:key (str "breadcrumb-" uid) :on-click #(navigate-uid uid)}
+               (or title string)]))]]
```
tangjeff0, line 102: Make sure you run `./script/lint` and `./script/style` when your PR's are ready for review.

src/cljs/athens/views/block_page.cljs
```diff
@@ -87,20 +96,23 @@
...
          [:h1 (use-style title-style {:data-uid uid :class "block-header"})
-          [autosize/textarea
-           {:id          (str "editable-uid-" uid)
-            :value       (:string/local @state)
-            :class       (when (= editing-uid uid) "is-editing")
-            :auto-focus  true
-            :on-key-down (fn [e] (node-page/handle-key-down e uid state nil))
-            :on-change   (fn [e] (block-page-change e uid state))}]
+          (let [diff? (when (not= (:string/local @state)
+                                  (:string/previous @state)) true)]
```
tangjeff0, line 107: Does this variable need to be bound outside the textarea, or can it be done from within `transact-string`? This value only needs to be computed on `blur` at the moment, but is instead computed on any change to the component including each key press. What if we put this logic, including `when diff?` inside `transact-string`.

Also, most event handlers in the Athens codebase have the event name as part of their name. It might make more sense to call `transact-string` something like `handle-blur` or `handle-textarea-blur` or `handle-header-blur`.

tangjeff0: Separate point: `when...true` seems like not a great practice too. `not=` will return a boolean true or false. Thoughts?

tangjeff0 (changes requested):

src/cljs/athens/views/block_page.cljs
lambduhh: 👍🏼 Yeah these whitespace changes from `cider-fomat-buffer` TODO: adopt athens style.. do we use zprint ? May be worth looking into https://github.com/kkinnear/zprint/blob/master/doc/using/team.md

src/cljs/athens/views/block_page.cljs
lambduhh: Thought I responded to this, but must have run before hitting enter.
1) Good point. Thoughts initially were to build `transact-string` to be as abstract as possible so could possibly be used elsewhere. Current fix should be better. 
1a) IMHO naming things is one of the hardest things to do in software 😂 My understanding of best practice is to try to avoid clunking up the names of functions with unnecessary or redundant verbs.. The re-frame documentation touches on it [here](https://day8.github.io/re-frame/EffectfulHandlers/#pattern-structure) . Most of my experience with this comes from having to be disciplined about avoiding the temptation to call every component `render-the-thing` . However, I see the benefit of having consistency among the codebase so I'll leave it up to you. 
2) Apologies, bad habit- now you know my secret!

tangjeff0 (approved):

PR #623, opened 2021-02-09
fix(ci): ignore windows Sentry variable

tangjeff0:
Curious if you know how to interpolate variables from the Windows command line @juniusfree 

In linux/mac it is:

`lein run -m shadow.cljs.devtools.cli --npm compile main renderer --config-merge "{:closure-defines {athens.core/SENTRY_DSN \"${SENTRY_DSN}\"}}"`

For Windows, I tried unsuccessfully to do:

`lein run -m shadow.cljs.devtools.cli --npm compile main renderer --config-merge "{:closure-defines {athens.core/SENTRY_DSN \"%SENTRY_DSN%\"}}"`

PR #628, opened 2021-02-10
fix: windows sentry variable

juniusfree:
fix windows sentry variable https://github.com/athensresearch/athens/pull/623

https://github.com/juniusfree/athens/runs/1868201203?check_suite_focus=true

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -210,10 +210,9 @@ jobs:
          env:
            SENTRY_DSN: ${{ secrets.sentry_dsn }}
 
-       # unable to interpolate Windows environment variables properly. Ignore for Windows for now
        - name: Compile JS Assets (Windows)
          if: matrix.os == 'windows-latest'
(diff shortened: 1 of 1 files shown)


tangjeff0:
It looks like the latest build, `d18ba1e`, failed: https://github.com/juniusfree/athens/actions/runs/556869285

Single quotes or double quotes?

juniusfree:
The first command that you made is correct. This is what I get on my local and the JS build is working.
image

I'm checking why the other test failed.

juniusfree:
@tangjeff0 All builds passed now https://github.com/juniusfree/athens/runs/1879603628?check_suite_focus=true

For future reference: Windows build is using Powershell, not Cmd.exe.

To escape special characters in Powershell: https://www.rlmueller.net/PowerShellEscape.htm

tangjeff0:
> For future reference: Windows build is using Powershell, not Cmd.exe.

> To escape special characters in Powershell: rlmueller.net/PowerShellEscape.htm


Very useful information. Thanks @juniusfree

PR #632, opened 2021-02-11
feat: added  linux "category" office

westofer:
this warning is recived when compiling under linux
```
application Linux category is set to default "Utility"  reason=linux.category is not set and cannot map from macOS docs=https://www.electron.build/configuration/linux
```
Adding it as "Office" > https://specifications.freedesktop.org/menu-spec/latest/apa.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "linux": {
       "target": [
         "AppImage"
-      ]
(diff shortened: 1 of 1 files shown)


tangjeff0:
Great job on your first contribution!

PR #634, opened 2021-02-12
fix: linked reference should show the formatting

### LINKED ISSUE #594: Linked References should show correct formatting.  Not markdown
When linked references are shown in the sidebar, they could be rendered accordingly as their markdown dicatates.  It would add better visual experience if the below screenshot showed the below in **BOLD** instead of wrapped in ** **

image

juniusfree:
fix #594

before:
image

fixed:
image

### DIFF
--- a/src/cljs/athens/parse_renderer.cljs
+++ b/src/cljs/athens/parse_renderer.cljs
@@ -90,7 +90,7 @@
   [tree uid]
   (insta/transform
     {:block         (fn [& contents]
-                      (concat [:span {:class "block" :style {:white-space "pre-line"}}] contents))
+                      (concat [:span {:class "block"}] contents))
      ;; for more information regarding how custom components are parsed, see `doc/components.md`
      :component     (fn [& contents]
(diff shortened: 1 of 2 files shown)


lambduhh (approved):
lgtm

tangjeff0 (commented):
This was more complex than it seemed, but accepting the changes for now, with my own additional change. See https://www.loom.com/share/a9034ce06a544203b6a1bb9a7fc84896

juniusfree:
@tangjeff0 Thanks for the feedback! I missed the issue on `white-space` because I focused on the right sidebar only. I should have checked the other components that use the linked reference component.

And for the parsing of block/page title, I also observed that one. I'll create a separate PR for that.

tangjeff0:
I believe `white-space` would've had the same issue on right sidebar. I don't think it's the location (nested child block, linked refs, or right sidebar), but it's wherever there are breadcrumbs and you could use `parse-and-render`.

PR #635, opened 2021-02-12
fix(router)598 shift clicking tag opens right hand sidebar

### LINKED ISSUE #598: SHIFT - LEFT CLICK of tags not opening in right-hand side bar
**Problem**

When `shift-clicking` on a tag, e.g. `#youtube`, it does not open in the right-hand sidebar.
When `shift-clicking` on a page, e.g. `[[youtube]]`, it does open in the right-hand sidebar. 

I would have expected that tags and pages would open in the right-hand sidebar when SHIFT- LEFT CLICKING. Not sure it is a bug or it is by design.

**Athens Version**
1.0.0-BETA.33 (MacOS, M1)

lambduhh:
-Clears re-frame console error that occurred when `right-sidebar/open-item` was called and sidebar was already open
-Fixes inconsistent behavior between user `shift-clicks` on tags vs pages. 

Close #598


src/cljs/athens/router.cljs
```diff
@@ -119,7 +119,7 @@
   ([uid]
    (let [current-route-uid @(subscribe [:current-route/uid])]
      (when (not= current-route-uid uid)
-       (dispatch [:navigate :page {:id uid}]))))
+       (dispatch [:right-sidebar/open-item uid]))))
```
tangjeff0, line 122: Incorrect, watch: https://www.loom.com/share/65967a4df0814245bb61fd67292a30e3

tangjeff0 (changes requested):

tangjeff0:
Accidentally merged this in, correct fix provided by #637

Issue #528, opened 2021-01-01
Allow either user-deletion or hiding of Daily Note entries


seltzered:
As of Athens 1.0.0-beta.25, Daily Note entries get automatically created when scrolling, which is fine, but if they get accidentally created there's no way to delete them. 

Propose for either:

- Allow user to delete Daily Note entries. This is probably easier to implement.

- Hiding null Daily Note entries from the 'All Pages' list and Search. 
I haven't dived into the codebase yet but this *might* make more sense from a backend perspective / consideration of interactions. Particularly when a user is trying to scroll through multiple prior days to intentionally create a Daily Note entry that's from several days ago. In that case with the first option, the user then has to manually delete the other Daily Note entries created.

itsrainingmani:
Allowing users to delete a Daily Note would be pretty easy since the `Delete Page` button is just hidden behind a condition check.

Regarding #2, would it make sense to have a checkbox in the Settings page to the effect of - "Hide empty Daily Notes from All Pages"?

seltzered:
Regarding hiding daily notes from all pages: Yes, a setting might be good enough. Worth noting though that I feel less bothered by this issue less now (on beta37) compared to when I was using beta25 now that the sort order is arranged by number of link references.

PR #638, opened 2021-02-13
fix: Allow deletion of Daily Notes

itsrainingmani:
This PR handles #528 by removing the check for `timeline-page?` in `node_page.cljs`. This renders the "Delete Page" button for Daily Notes.

There might still be some work done on this PR, particularly around the hiding of empty Daily Notes from the All Page List and Search.


tangjeff0:
> There might still be some work done on this PR, particularly around the hiding of empty Daily Notes from the All Page List and Search.

Can probably do that in a separate PR.

itsrainingmani:
Sounds good 👍🏼

itsrainingmani:
I created a separate `reg-event-fx` to handle the deleting a daily note on the Daily Notes page. This handler takes care of an edge case that I found while testing and removes the item being deleted from the `:daily-notes/items` vector in the db.

src/cljs/athens/events.cljs
```diff
@@ -338,6 +338,18 @@
          :dispatch [:page/create title uid]}))))
 
 
+(reg-event-fx
+  :daily-note/delete
+  (fn [{:keys [db]} [_ uid]]
+    (let [retract-blocks (db/retract-uid-recursively uid)
+          tx-data        (if (db/e-by-av :block/uid uid)  ;; If a page for the Daily Note exists
+                           retract-blocks                 ;; retract both the page and the child blocks
+                           (rest retract-blocks))         ;; else, only retract the child blocks
```
tangjeff0, line 347: Not sure I understand this `if` statement. How could a Daily Note page exist without children? 

Second, why not just dispatch `:page/delete` from within `:daily-note/delete`? All the code from `:page/delete` can be reused. For instance, looking at `:page/delete`, I see there is also a condition that checks for Linked Refs, that is not being checked here.

tangjeff0 (changes requested):

src/cljs/athens/events.cljs
itsrainingmani: I was trying to workaround this issue that popped up when dispatching `:page/delete` on the Daily Notes page. 

`"EXCEPTION" #error {:message "Nothing found for entity id [:block/uid \"02-16-2021\"]", :data {:error :entity-id/missing, :entity-id [:block/uid "02-16-2021"]}}`.

The only way I could get it to work was to remove the `[:db/retractEntity [:block/uid "02-16-2021"]]` item from the `retract-blocks` vector and pass that on to `tx-data`

Good catch on the linked refs part. I've fixed that now.

src/cljs/athens/events.cljs
```diff
@@ -338,6 +338,20 @@
          :dispatch [:page/create title uid]}))))
 
 
+(reg-event-fx
+  :daily-note/delete
+  (fn [{:keys [db]} [_ uid title]]
+    (let [all-blocks         (db/retract-uid-recursively uid)
+          retract-blocks     (if (db/e-by-av :block/uid uid)  ;; If a page for the Daily Note exists
+                               all-blocks                 ;; retract both the page and the child blocks
+                               (rest all-blocks))         ;; else, only retract the child blocks
```
tangjeff0, line 347: I still don't understand how a date page have child blocks if the date page doesn't exist.

tangjeff0: Why not just dispatch `:page/delete` from `:daily-note/delete`? (Assuming tx-data looks the same)

i.e.

```
(reg-event-fx
  :daily-note/delete
  (fn [{:keys [db]} [_ uid title]]
    (let [filtered-dn        (filterv #(not= % uid) (:daily-notes/items db)) ;; Filter current date from daily note vec
          new-db (assoc db :daily-notes/items filtered-dn)]
      {:fx [[:dispatch [:page/delete uid title]]]
       :db new-db})))
```

itsrainingmani: Ok. So I figured out why that error was popping up. When I just dispatched the `:page/delete` event directly from the button click. the daily note would be retracted, but it was still in the daily notes vector. The code here does prevent that from happening. Updating the PR now.

> (reg-event-fx
>   :daily-note/delete
>   (fn [{:keys [db]} [_ uid title]]
>     (let [filtered-dn        (filterv #(not= % uid) (:daily-notes/items db)) ;; Filter current date from daily note vec
>           new-db (assoc db :daily-notes/items filtered-dn)]
>       {:fx [[:dispatch [:page/delete uid title]]]
>        :db new-db})))
> ```

tangjeff0 (approved):
I think `is-timeline-page?` makes more sense than `daily-notes?` from a code perspective, since `daily-notes/items` is persisted even if someone clicks into a different page.

Also, if you try deleting daily notes a few times, it sometimes still fails. I believe this is because we can't guarantee that the `:daily-notes/items` is filtered before `posh` tries to update reactively.

But this works in most cases! Can fix later. 

Some more feedback that turned into a tangent on datascript vs posh, pull vs pull-many, eid vs unique id: https://www.loom.com/share/a1ccd13d60b54a0f82566ba324cf3d37

tangjeff0 (changes requested):
Still getting the alert, but making progress. Please see previous video/comments.

PR #656, opened 2021-02-16
fix(left-sidebar): reindex properly after adding or removing shortcuts

lambduhh:
Fixes issue where sometimes bookmarking will cause duplicate `:page/sidebar` values which leads to non-deterministic sorting of the left sidebar.


src/cljs/athens/events.cljs
```diff
@@ -450,21 +450,36 @@
       {:fx [[:dispatch [:transact tx-data]]]})))
 
 
+(reg-event-fx
+  :page/reindex-left-sidebar
+  (fn [_ _]
+    {:doc "This is used in the `left-sidebar` to smooth out duplicate `:page/sidebar` values when bookmarked. "}
```
tangjeff0, line 456: Does this `:doc` show up anywhere?

tangjeff0 (approved):

tangjeff0:
BTW there's one edge case where this doesn't work: the `Welcome` page that is created when the user loads Athens for the first time. This page has a `:page/sidebar` value of 999. It's not a big deal but it might come up.

src/cljs/athens/events.cljs
lambduhh: Just adding a docstring as metadata since `;;` was kinda ugly and this form didnt support `(defn foo "Prefered Doc-string" [x] )` notation.

PR #660, opened 2021-02-17
fix: block title is not parsed

### LINKED ISSUE #659: The title is not parsed.
**Problem**

The title is not parsed. 

**Screenshots/Demo**

image
image

**Athens Version**
1.0.0.40-BETA.40

juniusfree:
fix #659
image

Also fixed these related issues:
- overlapping content when editing the block title
athens-659

- unparsed breadcrumb on the block page
image

fixed:
image

### DIFF
--- a/src/cljs/athens/views/node_page.cljs
+++ b/src/cljs/athens/views/node_page.cljs
@@ -507,7 +507,9 @@
                           :class    "page-header"
                           :on-click (fn [e]
                                       (.. e preventDefault)
-                                      (dispatch [:editing/uid uid]))})
+                                      (if timeline-page?
+                                        (navigate-uid uid e)
+                                        (dispatch [:editing/uid uid])))})
           ;; Prevent editable textarea if a node/title is a date
           ;; Don't allow title editing from daily notes, right sidebar, or node-page itself.
           [button {:class    [(when show "active")]
--- a/src/cljs/athens/views/node_page.cljs
+++ b/src/cljs/athens/views/node_page.cljs
@@ -505,17 +505,11 @@
          [:h1 (use-style title-style
                          {:data-uid uid
                           :class    "page-header"
-                          :on-click (fn [e] (navigate-uid uid e))})
+                          :on-click (fn [e]
+                                      (.. e preventDefault)
+                                      (dispatch [:editing/uid uid]))})
           ;; Prevent editable textarea if a node/title is a date
           ;; Don't allow title editing from daily notes, right sidebar, or node-page itself.
-          (when-not timeline-page?
-            [autosize/textarea
-             {:value         (:title/local @state)
-              :id            (str "editable-uid-" uid)
-              :class         (when (= editing-uid uid) "is-editing")
-              :on-blur       (fn [_] (handle-blur node state linked-refs))
-              :on-key-down   (fn [e] (handle-key-down e uid state children))
-              :on-change     (fn [e] (handle-change e state))}])
           [button {:class    [(when show "active")]
                    :on-click (fn [e]
                                (.. e stopPropagation)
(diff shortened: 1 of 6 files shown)


juniusfree:
Added parse to the following:

1. Node title
athens-660-node

(An error occurs for this one after editing the node title. I'll investigate later)

2. Sidebar title
image

3. Athena search result
image

4. Inline search
image

tangjeff0:
I love how you are thinking about the other places where strings are rendered @juniusfree.

Aside from the code implementation, we should also think from the product perspective of _should_ we always format the text.

In the two images below, Roam doesn't format the text in inline search or in Athena.


image

image

juniusfree:
@tangjeff0 That's interesting. I was thinking that it's better if the blocks and page should be displayed consistently across the system.

juniusfree:
Reverted the changes to Athena and inline search.

tangjeff0 (changes requested):
I'm getting this error when trying to type something. Not sure why. Can't find any changes to your use of buttons 
image

juniusfree:
I'll check

juniusfree:
@tangjeff0

I fixed the issue above
image

juniusfree:
@tangjeff0 

The issue above is fixed.
image

src/cljs/athens/views/node_page.cljs
```diff
@@ -527,8 +521,15 @@
...
-          ;;(parse-renderer/parse-and-render title uid)]
+          (when-not timeline-page?
+            [autosize/textarea
+             {:value         (:title/local @state)
+              :id            (str "editable-uid-" uid)
+              :class         (when (= editing-uid uid) "is-editing")
+              :on-blur       (fn [_] (handle-blur node state linked-refs))
+              :on-key-down   (fn [e] (handle-key-down e uid state children))
+              :on-change     (fn [e] (handle-change e state))}])
+          [parse-renderer/parse-and-render (:title/local @state) uid]]
```
tangjeff0, line 532: Why does the location of this snippet need to change?

src/cljs/athens/views/node_page.cljs
```diff
@@ -505,17 +505,11 @@
          [:h1 (use-style title-style
                          {:data-uid uid
                           :class    "page-header"
-                          :on-click (fn [e] (navigate-uid uid e))})
+                          :on-click (fn [e]
+                                      (.. e preventDefault)
+                                      (dispatch [:editing/uid uid]))})
```
tangjeff0, line 510: So I agree we want to start editing the textarea when we click on the h1. However, I'm pretty sure `:editing/uid` is not actually used in any way for node pages. Not big deal either way.

More importantly, this change no longer lets the user navigate to a date page when they are on Daily Notes. Perhaps the `:on-click` handler should have an `if` statement for Daily Notes?

tangjeff0 (changes requested):

src/cljs/athens/views/node_page.cljs
juniusfree: The h1 and textarea input should be adjacent to each other that's why I moved this down. I'm not entirely sure why that's the case but I'm guessing it's the z-index on css.

src/cljs/athens/views/node_page.cljs
juniusfree: Fixed the issue with Daily Notes.

Sorry, I just need to clarify. When you say `:editing/uid` is not used for the node page, why does the node page subscribe to it?

tangjeff0: I wasn't sure when I first created node page whether or not it needed `:editing/uid`. Arguably, it could use it.

The purpose of `:editing/uid` is to let the component know whether it is focused or not. This is more useful for blocks, where the caret could be moved to another block. Enter, backspace, up and down arrows are all places that could update :editing/uid.

In the case of block-page and node-page, a user might go to a child block if they press enter on the header (I believe this might've actually been your first PR!).

Arguably, pressing backspace on the 0th child block should take you user's focus back to the title page. Roam doesn't do this, but Notion does do this.

However, it's not super obvious node-page and block-page actually need this anymore. Maybe that would be code to remove in another PR.

tangjeff0 (approved):

src/cljs/athens/views/node_page.cljs
juniusfree: So far, what I found is that `:editing/uid` is used to check whether to apply `is-editing` css class. I might try to investigate more later.

Thanks for the feedback.

PR #666, opened 2021-02-18
fix(keybinding):block save before navigation #612

### LINKED ISSUE #612: Block not saved when navigating with keyboard
**Problem**

It seems that saving edits to a block is triggered only when unfocusing the line. If you navigate with keyboard to a link (e.g. with ctrl-o), the line you were editing is not persisted to the database.

**Screenshots/Demo**

https://user-images.githubusercontent.com/2254049/106908000-28139e00-66ff-11eb-984b-4bb519e200dd.mp4

**Athens Version**
1.0.0-beta.33

arkRedM:
close #612

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -5,7 +5,8 @@
     [clojure.edn :as edn]
     [clojure.string :as string]
     [datascript.core :as d]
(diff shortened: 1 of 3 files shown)


tangjeff0 (commented):
I probably would've solved this by making the textarea `blur` in keybindings... I like your solution more :)

Issue #281, opened 2020-07-19
Workspaces


txnka:
### Prototype
[Link to the Figma prototype](https://www.figma.com/proto/QR47RdiCVV4QdAPTAclBwy/Athens-Sidebar-Sandbox?node-id=55%3A20&viewport=-1493%2C-1905%2C0.5&scaling=contain)

### Feature Description
Workspaces in Athens could serve as a way to group related pages and blocks in order to facilitate long-form writing by being able to remix (edit, change, rearrange) content without having to worry about changing the original blocks. Therefore it would be important for blocks / pages moved into a workspace to be copied into the workspace (instead of embedding them), while keeping a reference to the original. This could also serve as a way of versioning and allow the export of e.g. a whole essay with citations.
Workspaces could also be extended by introducing „view modes“ - different visualizations of the workspace content e.g. a graph view of just the workspace contents (instead of a graph of the whole DB), an outline view, a versioning view, a relationship view (how blocks relate to each other), etc. The view modes could also change automatically based on zoom level.

### Tasks

- [ ] Refinement of the feature scope

tangjeff0:
Closing, need to improve core product before we can get to this new feature

PR #673, opened 2021-02-21
feat: Remember Window Size

### LINKED ISSUE #616: Remember window size when close Athens
A nice to have:

Athens remembers the window size upon closing. I.e. when a user closed Athens and did resize the window, it remembers its last size when reopening. 

Best
Daniel

itsrainingmani:
This PR handles #616. The user can enable this feature via a toggle in the Settings page enabling Athens to "remember" the window size when Athens was last closed.
Technical Details

We add a new `init-windowsize` function that handles all window-size related functionality and is called within `init` in `athens.core`.

* Check if the "Remember Window Size" option is set to "on" in `localStorage`. (This value can be toggle on and off in the Settings Page). If it is, extract the last known window width and height from `localStorage` and set the size of the current window to those values. We also center the window in the current screen to avoid misalignment.
* Create an event handler to handle the "resize" event for the current window. This event handler gets the current size of the window and dispatches a re-frame event, `:window/set-size` which stores the provided values in `localStorage`.
* We also add some utility functions for reading the required values from `localStorage` and provide sane defaults if the values don't exist in `localStorage`.

Close #616

### DIFF
--- a/src/cljs/athens/core.cljs
+++ b/src/cljs/athens/core.cljs
@@ -83,10 +83,9 @@
           remember-ws?     (util/remember-ws?)
           [lastx lasty]    (util/get-window-size)]
       (when remember-ws?
-        (do
-          ;; (prn (str "Window Size on close - " lastx ", " lasty))
-          (.setSize curWindow lastx lasty)
-          (.center curWindow)))
+        (
+         (.setSize curWindow lastx lasty)
+         (.center curWindow)))
       (.on ^js curWindow "resized" (fn [e]
                                      (let [sender (.-sender e)
                                            [x y] (.getSize ^js sender)]
--- a/src/cljs/athens/core.cljs
+++ b/src/cljs/athens/core.cljs
@@ -83,12 +83,12 @@
           remember-ws?     (util/remember-ws?)
           [lastx lasty]    (util/get-window-size)]
       (when remember-ws?
-        ((.setSize curWindow lastx lasty)
-         (.center curWindow)))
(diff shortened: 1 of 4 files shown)


itsrainingmani:
Example of how this works - https://www.loom.com/share/3a29b6d8b52449aeaa06a1b502fcac04

tangjeff0 (approved):
This is an amazing job @itsrainingmani. 

Why is this setting an option? My intuition is that the window size should always be remembered! It looks fine know, but eventually we'll have lots more settings I imagine. You or I maybe should remove this as an option in a future PR?

---

There was an edge case when I tested this out on Mac using [Spectacle](https://www.spectacleapp.com/). Seems like it doesn't actually call the js event! But nbd.

---

Love that you did this yourself with localStorage. Is there any reason in particular you didn't use an external library for this? https://discord.com/channels/708122962422792194/708124073544122399/807138845912858634

Going forward, I wonder how we could have a more robust settings configuration. Don't love localStorage :joy:

- https://github.com/sindresorhus/electron-store
- https://github.com/nathanbuchar/electron-settings

What's interesting about these 2 libraries above is that they are pretty local only. However, settings in a web app like Notion are server only.

What would be a good user config persistence, that works for both local and remote apps? What if we leveraged the re-frame db so we have a cleaner clojure interface?

What if we leveraged datascript directly? Sharing config across a db doesn't sound interesting, but imagine if we could persist a "Workspace" - a snapshot of all the pages open, the right sidebar, etc. See #281

PR #726, opened 2021-03-03
enhance(parser): Add LaTeX support

neotyk:
closes #575

PR #767, opened 2021-03-07
feat(graph): local-graphs and view customization

arkRedM:
https://www.loom.com/share/b3f16aa7c4084c69b4f47a0e03f7bf65

I am not seeing a perf issue for my current graph, although if it's leading to a bad ux, I can include it in this one or come back to it later.

Added docs based on community call discussion. I think incrementally adding docs from each pr(where code has been added or modified) is better than out-of-context adding of docs.

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -6,7 +6,8 @@
                       random-uuid
                       goog.DEBUG
                       embed-id
-                      p]}
+                      p
+                      is-open?]}
            :unused-referred-var
            {:exclude {clojure.test [is deftest testing]}}
            :unsorted-required-namespaces {:level :warning}}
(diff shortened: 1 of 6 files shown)


arkRedM:
Strangely my local lint is ignoring with-let(as expected) -- not sure why this is failing here.

tangjeff0:
@neotyk updated `cljstyle` here. Maybe that's it? https://github.com/athensresearch/athens/pull/726/files?file-filters%5B%5D=No+extension

arkRedM:
I think it's only related to lint(clj-kondo)

tangjeff0:
Taking a look

borkdude:
Instead of ignoring  `r/with-let` bindings, I suggest adding the config `{:lint-as {reagent.core/with-let clojure.core/let}}`.

PR #768, opened 2021-03-07
fix(enter/tab): do not apply changes to previous line when fast keystrokes 

### LINKED ISSUE #602: fix(input): pressing Enter and then pressing tab /shift+tab results in the changes being applied to the previous line
**Problem**
After pressing enter to go to a new line and then pressing tab / shift+tab the tabbing goes to the previous line from which you pressed enter from.

**Screenshots/Demo**

https://user-images.githubusercontent.com/77367499/106806762-58433a00-6636-11eb-82da-93460fc9bc0d.mp4

This issue has a reproductivity rate that seems tied to how fast you are trying to make changes to the new line you are on.

**Athens Version**
1.0.0-BETA.33

neotyk:
Fixes #602 

Enter actions where unnecessarily delayed.

This undoes patch to fix race condition when Enter is pressed and
blocks are created too fast.

At the same time it introduces solution to continuous Enter,
by ignoring Enter until blur event fires.
No more race condition.

Please test some crazy Enter usage.

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -5,11 +5,10 @@
                       (devcards.core/defcard-rg)
                       random-uuid
                       goog.DEBUG
-                      embed-id
-                      p
-                      is-open?]}
+                      (com.rpl.specter/recursive-path)]}
            :unused-referred-var
            {:exclude {clojure.test [is deftest testing]}}
            :unsorted-required-namespaces {:level :warning}}
  :lint-as {day8.re-frame.tracing/fn-traced clojure.core/fn
-           day8.re-frame.tracing/defn-traced clojure.core/defn}}
+           day8.re-frame.tracing/defn-traced clojure.core/defn
+           reagent.core/with-let clojure.core/let}}
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -6,7 +6,8 @@
                       random-uuid
                       goog.DEBUG
                       embed-id
-                      p]}
+                      p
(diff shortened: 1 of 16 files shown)


tangjeff0:
Holding enter creates line breaks: https://www.loom.com/share/7fa9e258da4d4dd6ba81195dae71238e

neotyk:
Can't reproduce it locally on Mac.

Maybe we could rate limit enter?

So dispatch 1st enter when pressed and denounce for next 50ms. This would be similar to what we had, but would dispatch 1st Enter event immediately.

This is going to be a bit problematic because different text areas receive :on-key-down.

For some reasons :on-key-up is not getting triggered.

neotyk:
@tangjeff0 can't reproduce locally on Mac, can you please test with these changes on Linux?

tangjeff0:
Rate limiting works, but indent/unindent have a small bug. Textarea value is copied to a new block on quickly pressing enter and tab.

https://www.loom.com/share/0010be7543f14607a9572be32470c228

neotyk:
Nice catch, let me see if I can reproduce it.

neotyk:
@tangjeff0 Reproduced locally

neotyk:
Pushed fix to "copy" bug.
[Enter] and [Tab] when pressed very fast are registered on textarea of starting block.
After handling [Enter], `:editing/uid` is already pointing to new block, but the [Tab] event was registered in starting block.
I've put a safeguard to not process key events from textarea that is not registered for `:editing/uid`.

Issue #669, opened 2021-02-20
enhance(all-pages): allow sort by titles / links / times


yifanyin:
The page presents a table with all the notes listed in it. It would be nice to be able to sort them. For now, it seems to sort the notes by the number of links.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 300.0 USD (300.0 USD @ $1.0/USD)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/athensresearch/athens/669/100024999).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/athensresearch/athens/669/100024999).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
* $2,657,214.30 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 265 years, 8 months from now.
Please review their action plans below:


**1) [lesleyvdp](https://gitcoin.co/lesleyvdp) has started work.**

I am planning to figure out how the default sorting is done now. E.g. how it comes that the table is now sorted based on total amount of links.

I will then figure out what else the application should sort on, since all columns except for the body one seem to make sense to be able to sort on.

I will then implement the sorting itself and allow for these sorts to be activated/deactivated by clicking on the column titles.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/athensresearch/athens/669/100024999).

tangjeff0:
Will have some overlap at #776 cc @sboessen. 

See if you can run devcards via `yarn run shadow-cljs watch devcards`. Filters can sort by count or alphabetical in both directions (ascending and descending). This is one implementation. (I'm not sure if it's the best one).

https://github.com/athensresearch/athens/blob/master/src/cljs/athens/views/filters.cljs
https://github.com/athensresearch/athens/blob/master/src/cljs/athens/devcards/filters.cljs

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

[...]

PR #776, opened 2021-03-08
enhance(linked-refs): sort references by newest first

### LINKED ISSUE #728: enhance(linked-refs): sort references by newest first
Please search our GitHub and Discord to see if your feature has already been requested!
References, both linked and unlinked are shown chronologically. However, I would find more useful if they were sorted "newer first" instead of the current behaviour "older first"
in discord: https://discord.com/channels/708122962422792194/800439780910563358/815988756913324042

Athens Version
1.0.0-BETA.47

sboesen:
Fixes #728, sorts references by newest note ID instead of oldest ID.

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -598,7 +598,7 @@
        (mapv :db/id)
        merge-parents-and-block
        group-by-parent
-       (sort-by :block/id)
+       (sort-by :db/id)
        vec
        rseq))
 
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -598,7 +598,9 @@
        (mapv :db/id)
        merge-parents-and-block
        group-by-parent
-       vec))
+       (sort-by :block/id)
+       vec
+       rseq))
 
 
 (defn get-linked-block-references


tangjeff0:
This is a partial solution. What about for non-date pages, where the `:block/uid` is random?

What would it look like to add different sorts like in #669?

sboesen:
Good call outs - it does appear that complete implementation of this issue is blocked on implementation of #669, or will require further edits after that's complete.

tangjeff0:
I wouldn't say it's blocked. Both of can try to figure out sorting independently!

sboesen:
> What about for non-date pages, where the :block/uid is random?
Currently it's sorted by `block/:id` which appears to handle showing new non-date pages first just fine, see screenshot:

Screen Shot 2021-03-08 at 9 38 20 AM

>I wouldn't say it's blocked. Both of you can try to figure out sorting independently!
If we come up with the same implementation it wouldn't be blocked, agreed! Trying to minimize throwaway work.

tangjeff0:
There should also probably be a switch to hide date pages. What do you think @arkRedM @dauberd

sboesen:
So I think the "newest note sort" commit does cover use case as described in this issue after rereading the thread and comments.

> What would it look like to add different sorts like in #669?
I think we'd need new UI elements to allow toggling sorting of various attributes on linked reference pages, on #669 it's pretty clear that you can click the "title", "links", "body" etc headings and it might sort the notes, currently the Linked References notes don't have anything like that.

tangjeff0:
I'm not sure how your implementation works. `:block/id` is not used anywhere in the codebase!

sboesen:
Oh geez, looks like I'm doing page ID when I meant block ID, and yeah, that's not a thing. Back to the drawing board, sorry for the false start here!

sboesen:
So I tested and my change actually appears to do what we want as is (not including the proposed feature of toggling to show/hide date notes):

Screen Shot 2021-03-08 at 10 42 55 AM

Currently it is ordered by the block/page ID in descending order, which results in the latest block displaying first. After editing "older page newer ref" (which was created before "new page older ref") to add a link to `[[Index]]` it appears at the top.

sboesen:
Added new commit with the same behavior described/in screenshot above, but using `:db/id` which actually makes sense. I actually don't know how `:block/id` worked, but it did? Still learning :D

tangjeff0 (approved):

Issue #749, opened 2021-03-05
fix(electron): "Cannot read property '0' of null"


zendude213:
**Problem**
Can't install latest Mac update. On opening App, I'm prompted to update.. then it tries for a while and after a few seconds, pix below pops up. If I restart and select "cancel" as opposed to "update", things work as normal

**Screenshots/Demo**
![CleanShot 2021-03-05 at 04 46 24](https://user-images.githubusercontent.com/59263556/110098169-e1e94300-7d6d-11eb-9eb5-8bf6dd57aa3a.png)

**Athens Version**
Version 1.0.0-beta.47 (1.0.0-beta.47)

shebang42:
Same issue.  When upgrading from b47 to (presumably) b50 and from b49 to b50.  On both Macos 11.2 and 10.5.7.

Workaround:

1: **Download** Athens from https://athens-apps.s3.us-east-2.amazonaws.com/Athens-1.0.0-beta.50.dmg
2: **Close Athens**
3: **Install** (== copy to /Applications).  Select '**Replace**' with the popup "An item named “Athens.app” already exists in this location. Do you want to replace it with the one you’re moving?".

zendude213:
thanks.. now works!

tangjeff0:
This happened after #707, which updates `electron` and `electron-builder` so we could make the M1 build. cc @sboesen

tangjeff0:
May also be just be same error as #722

sboesen:
@tangjeff0 does look like the same issue as #722 at first glance - helps narrow it down that it appears to happen just upgrading from beta 46 - 47.

tangjeff0:
The difference between this issue and #722 is that this one seems to prevent people from actually updating.

sboesen:
Here is the error I'm seeing in `~/Library/Logs/Athens/main.log` on Mac when trying to update from Athens beta 49 -> 50:

```
[2021-03-05 16:28:35.401] [info] Error in auto-updater. Error: ZIP file not provided: [
  {
    "url": "https://athens-apps.s3.amazonaws.com/Athens-1.0.0-beta.50-arm64.dmg",
    "info": {
      "url": "Athens-1.0.0-beta.50-arm64.dmg",
      "sha512": "vi1mVDiVLGI9iUDnKxfl6h+7EkZ3imDq4QQ1zD9ildH0HOts9hYnEEa+vOR3LwZW+yzaUurFUGRb/JvnntolQA==",
      "size": 117351332
    }
[...]

Issue #722, opened 2021-03-02
Uncaught TypeError: Cannot read property '0' of null


mjio:
**Problem**
Athens is crashing on my system, macOS version 11.2.2. I'm using Athens 1.0.0-beta.46. The error happens when I try to update to version 1.0.0-beta.47.

```
message=Uncaught TypeError: Cannot read property '0' of null
source=file:///Applications/athens.app/Contents/Resources/app.asar/resources/public/js/compiled/renderer.js
lineno=15103
colno=505
error=TypeError: Cannot read property '0' of null
```

**Screenshots**
<img width="269" alt="Screen Shot 2021-03-02 at 11 04 57 am" src="https://user-images.githubusercontent.com/20990/109632591-7aea4500-7b47-11eb-8d41-6d3691251230.png">

**Athens Version**
1.0.0-beta.46

tangjeff0:
That’s super strange. Not at my computer right now. Will look into this tomorrow. In the meantime, you can download the latest version directly by using the same Mac download link or rollback to a previous version. Steps are somewhere on discord #troubleshooting I believe.

Sent via Superhuman iOS ( https://sprh.mn/?vip=tangj1122@gmail.com )

mjio:
I've updated manually to version 1.0.0-beta.47 and that works fine. If no one else sees this issue, I guess we can close it.

tangjeff0:
I had this issue too from Linux. Dang!

TillScout:
also occured on Windows, but the installation went through (showing a couple more error messages, though)

yifanyin:
I also encounter this message after clicking the update prompt. After seeing the same message and press "OK", the app turned blank. I just leave it be and seems like Athens still updated itself under the hood and quit afterward. When I restart Athens it is the 1.0.0-beta.47 version. I would suggest a better visual for the update process and automatic restarting after updates.

eerohele:
Got the same error on Big Sur.

tangjeff0:
This bug was introduced after trying to fix another one here: https://github.com/athensresearch/athens/commit/68eb2d8e9364caf6392aa2edef91e19f4c12b903

My guess is that we need a guard for `-webContents` as well.

tangjeff0:
Sentry gives us a more verbose error message:

![image](https://user-images.githubusercontent.com/8952138/110188545-f14da800-7dd0-11eb-8f00-b82b8923e307.png)

tangjeff0:
[...]

PR #777, opened 2021-03-08
fix(electron): issue updating Mac App

sboesen:
This PR fixes just the package.json config to re-enable ZIP target on Mac builds, which should re-enable auto-update on Mac.

Fixes https://github.com/athensresearch/athens/issues/749

Note: Issue #749 references an error popup, but was originally about auto update on Mac being broken, which this fixes.

#722 appears to reference that same error popup in #749, and is still not fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,13 +21,22 @@
     "generateUpdatesFilesForAllChannels": true,
     "afterSign": "electron-builder-notarize",
     "mac": {
-      "target": {
-        "target": "dmg",
-        "arch": [
-          "arm64",
-          "x64"
-        ]
-      },
+      "target": [
+        {
+           "target": "dmg",
+           "arch": [
+              "arm64",
+              "x64"
+           ]
+        },
+        {
+          "target": "zip",
+           "arch": [
+              "arm64",
+              "x64"
+           ]
+        }
+     ],
       "hardenedRuntime": true,
       "entitlements": "build/entitlements.mac.plist",
       "entitlementsInherit": "build/entitlements.mac.plist"


sboesen:
This issue should be reopened, I suspect auto-update now works for M1 macs but after testing from Athens Beta 52 on an Intel mac it downloads the ARM64 update of Beta 53 by default, breaking the current Athens installation.

Current workaround for Intel Macs is to click "Cancel" during update prompt and wait for fix.

Screen Shot 2021-03-09 at 1 53 58 PM

I suspect this may be a bug/feature gap in electron-updater after reviewing the relevant code introduced in the most recent electron-builder release 22.10.5 in an attempt to fix auto-update for M1 macs ([code here](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/MacUpdater.ts#L30-L37), relevant [PR describing the issue](https://github.com/electron-userland/electron-builder/pull/5524) were trying to fix) and I'm still looking to see if there's something we can do to avoid this issue or if we need to roll back electron-builder to 22.10.4 (which doesn't include these filters).

Will be testing on my own S3 bucket to see if I can get this working, should have an update later today or tomorrow morning. We'll definitely want to make sure that this works on M1 macs too after the change (and doesn't download the Intel version), something I'm unfortunately still limited in testing.

tangjeff0:
@sboesen created a new issue for this #793. Someone encountered the issue with x64 -> arm. My M1 was able to auto-update, so that was tested. By the way, you can create a local AWS server with Minio. See docs here: https://github.com/athensresearch/athens/pull/750/files

PR #779, opened 2021-03-09
fix(web): theme and graph conf issue #773

### LINKED ISSUE #773: fix(graph-viz): freezes on web demo
If you load the test DB on https://athensresearch.github.io/athens, trying to open a graph view (local and global) freezes. This may be because this loads `data/ego.datoms` which may contain Roam tuples that are invalid for the current schema.

arkRedM:
close #773


arkRedM:
seed data(graph-conf) added to DB when loading test db along with minor fixes in loading init theme(affecting graph which was dependant on app-db theme data)

Issue #570, opened 2021-01-26
Improve Performance for large DBs


tangjeff0:
After uploading the Roam [ego](https://roamresearch.com/#/app/ego) db to my personal Athens db, performance has greatly suffered. It takes over half a second for the page to update after writing a block to datascript. This seems to originate from `(anonymous)` functions and `datascript.pull_api` private functions. Images below:

![image](https://user-images.githubusercontent.com/8952138/105782083-a49cc480-5f28-11eb-9705-e13cf1077525.png)

![image](https://user-images.githubusercontent.com/8952138/105782097-ad8d9600-5f28-11eb-8593-a0308b9eb2e6.png)

![image](https://user-images.githubusercontent.com/8952138/105782110-b7af9480-5f28-11eb-8888-fe7b12a99684.png)

quolpr:
I think this question somehow relates to this issue. I am curious, how Athens will scale well if all data are stored in datascript that is in-memory DB? Like, a potential issue, I see that Athens should start lag on large datasets if the user has not enough RAM, and the first load time will always be increasing 🤔. What do you think? 

The same question I have to the Roam team too 😄️

tangjeff0:
Big pages are also quite slow.

https://discord.com/channels/708122962422792194/709246147549593601/809965655302340621
https://www.loom.com/share/e1044dcaafc34653b612a4ccaf235eaf

sid597:
Any updates on this issue?

tangjeff0:
@juniusfree is making a dent. He just did #756 and is looking to replace posh with reactive datascript. #1167

sid597:
> @juniusfree is making a dent. He just did #756 and is looking to replace posh with reactive datascript. #1167

Thats great, is posh not replaced ? A while back someone in discord mentioned they were making progress on this.

tangjeff0:
#784 #665 @sid597

PR #784, opened 2021-03-09
Replace posh with ratom

pithyless:
This is a continuation of the work done by @lambduhh on https://github.com/athensresearch/athens/pull/665. Kudos to @lambduhh for the idea!

The supporting code has changed significantly, though, so I thought it would be easier to open a separate PR.

## How to reproduce

The TL;DR to run this branch:

1. Uncomment `athens.tracing` in `shadow-cljs.edn`: `(-> :renderer :devtools :preloads)`
2. Choose your poison in `athens.posh/version` (`:posh` is original flavor, `:dbrx` is brand new fizzy flavor)
3. Rebuild (`lein dev`) and run Athens (`yarn run electron .`), open the DevTools Console, and click around.

Two reports will show up. `Epoch Stats` report percentiles for profiled functions between epochs (as understood by re-frame-10x). `Total Stats` is a running total for the entire session. Epochs can help you identify the performance characteristics of a specific action, while Totals can help approximate the general performance of the app during regular usage.

report1

## What are the results?

For my tests, I'm using a basic Welcome db + a large tagged markdown file.

### `:posh` - original flavor

Open/Close sections of the Welcome page:

posh1

Open/Close sections of the Markdown report:

posh2

⚠️ 408 ms just to process the `transact!`. So, with additional parsing and rendering we have failed the Doherty Threshold!

### `:dbrx` - brand new fizzy flavor

Open/Closing sections of the Welcome page:

dbrx1

Open/Close sections of the Markdown report:

dbrx2

💯 Max 40 ms response times.

## Next Steps

This should also benefit https://github.com/athensresearch/athens/issues/570

Athens is still noticeably slow when working with large files, but this is the first of several problems that needed to be fixed; the next biggest bottleneck may be in parsing and/or rendering.

In the meantime, we need to test and make sure this approach doesn't break on some edge cases. More 👀 and user reports would be most welcome!

## PR Code Review

The last two commits are intended for profiling and may be optionally merged, depending on preference. I split them out to make it easier to cherry-pick this branch.

1. https://github.com/athensresearch/athens/commit/75b1d57fbee60b0cb05e62e2b8c12e709712c092
2. https://github.com/athensresearch/athens/commit/d0f18d3e8f75096177c3eef61ef9f8c08e7bc370

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -68,6 +68,7 @@
    {:dependencies [[binaryage/devtools "1.0.0"]
                    [day8.re-frame/re-frame-10x "0.6.0"]
                    [day8.re-frame/tracing "0.5.3"]
+                   [com.taoensso/tufte "2.2.0"]
                    [cider/cider-nrepl "0.25.1"]]
 
     :source-paths ["dev"]}
--- a/shadow-cljs.edn
+++ b/shadow-cljs.edn
@@ -29,7 +29,8 @@
                                         :closure-defines {re-frame.trace.trace-enabled? true}
                                         :output-feature-set :es-next}
                      :devtools {:preloads [devtools.preload
-                                           day8.re-frame-10x.preload]}}
+                                           day8.re-frame-10x.preload
+                                           #_athens.tracing]}}
 
           ;; backend for electron (node.js)
           :main {:target :node-script
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -1,11 +1,11 @@
 (ns athens.db
   (:require
     [athens.patterns :as patterns]
+    [athens.posh :as posh :refer [pull q]]
     [athens.util :refer [escape-str]]
     [clojure.edn :as edn]
     [clojure.string :as string]
     [datascript.core :as d]
-    [posh.reagent :refer [posh! pull q]]
     [re-frame.core :refer [dispatch]]))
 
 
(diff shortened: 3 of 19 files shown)


tangjeff0:
Results from testing open/close of a bullet on my personal DB.

posh
image

dbrx

image


---

Results from testing open/close of a bullet the ego db. Can test by adding `athens.tracing` to the `:app`  `:preloads` vector, and also watching `shadow-cljs watch main renderer app`.

Posh 

image

dbrx

image

---

Time seems to be about the same or in some cases higher?

pithyless:
@tangjeff0 any chance you did not restart the app after changing the `athens.posh/version`? That's the only thing I can think of right now to explain your observations. Here is how it looks on my machine:

https://www.youtube.com/watch?v=8lAth0JH3TQ

Also, as I'm posting this now, I realize I made a mistake in how I interpret the results for the new dbrx `pull*` queries. Because it's all wrapped in an ratom, the profiler does not actually measure how long an operation took (laziness strikes again!). Sorry about the confusion and misreporting.

But this should not change the observations about the performance of `transact!`, which is a blocking operation. I think this PR is still valid to reduce the time for `transact!`, but it looks like further work is necessary to refactor the components to make them reactive and only do the minimum amount of work when fetching/transforming data for rendering (as @lambduhh and @neotyk have already been pointing out). That probably should be raised as an independent issue and PR, to avoid growing this one further.

From my point of view, there are two open questions:
1. Make sure this PR is actually improving `transact!` times. @tangjeff0 if you could double-check your observations, that'd be great. But it would also be beneficial if someone else could have a look over this PR and point out if I've made any goofball assumptions (like the one about profiling laziness).
2. If the PR does look OK, would we want to merge it or wait until we have a more comprehensive solution with the reactive read/renders resolved as well?

tangjeff0:
My results testing on my actual db:

posh
image

dbrx
image

PR #796, opened 2021-03-10
fix(electron): issue updating Mac (Intel, not Silicon)

### LINKED ISSUE #749: fix(electron): "Cannot read property '0' of null"

### LINKED ISSUE #793: Normal Macs update into m1 arm64 build
**Problem**
https://discord.com/channels/708122962422792194/800442937602342922/819129727193317397

Related to #749, #777

**Screenshots/Demo**
image

**Athens Version**

sboesen:
Fixes #749, fixes #793 due to default ordering resulting in electron-update choosing ARM64 binary on Intel Macs

See comment on #749 describing validation testing performed

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,15 +25,15 @@
         {
           "target": "dmg",
           "arch": [
-            "arm64",
-            "x64"
+            "x64",
+            "arm64"
           ]
         },
         {
           "target": "zip",
           "arch": [
-            "arm64",
-            "x64"
+            "x64",
+            "arm64"
           ]
         }
       ],


sboesen:
Note: It's still worth testing what update path M1 macs take, as that was not tested with the reorder change. If someone wants to test that before merging you can use my s3 bucket, just change these lines in `package.json` and make sure the version is on line 4 is set to 54 or below to force an update.
```
    "publish": {
      "provider": "s3",
      "bucket": "stefan-athens-apps",
      "region": "us-west-2"
    }
```

The binaries aren't code signed so you'll get an error when it tries to validate their signatures, but you can still find out which binary it chose to download by checking the logs in `~/Library/Logs/Athens/main.log` or finding the update file in `~/Library/Application Support/Caches/athens-updater/`

tangjeff0:
Looking at this. So far have repro'd on my M1 and Macbook that it by default tries to update to `arm`. Good for M1, not for MacBook :sweat_smile: 

Next: testing out update with your S3

sboesen:
> So far have repro'd on my M1 and Macbook that it by default tries to update to arm. Good for M1, not for MacBook 😅

Hopefully you mean with the existing build, not with this reordering change? Still not 100% sure why it only works with x64 listed first, but does update properly (or rather, download the update and then fail on code signing) on Intel via my s3 bucket!

tangjeff0:
> Hopefully you mean with the existing build, not with this reordering change?

Yes that's what I mean. The question now is if M1 will now try to auto-update to Intel. Happy to jump on Discord to solve this too!

sboesen:
Just noting from call with @tangjeff0 that M1 appears to auto-update to the default build, as well, so users that want the M1 build will need to manually download/replace Athens until either a workaround is implemented or electron-builder/electron-updater is updated to support intelligently selecting which update to download based on CPU architecture

tangjeff0:
Following up. Just tested 55 -> 56 and confirming that both Intel and Silicon update to Intel now. That means M1 users should not auto-update.

Issue #717, opened 2021-03-01
enhance(undo): undo by transaction rather than dropping entire db


DrHyde:
**Problem**

Undo (Apple-z) behaviour is a bit inconsistent. If you double-click-select a word and hit Apple-b to embolden and then immediately hit Apple-z to undo, I would expect to end up with no changes made. But A-z does nothing in that case. If you move the cursor off the line you just edited though and _then_ hit A-z the undo happens as expected. If instead you embolden a word by typing a double-asterisk either side of it and then hit A-z then each character you typed is undone by one A-z, as expected.

**Athens Version**

1.0.0-beta.46 on Mac OS Catalina

tangjeff0:
Undo can be improved significantly. Thank you @DrHyde

tangjeff0:
See #755 ![image](https://user-images.githubusercontent.com/8952138/110256378-7958bc80-7f4d-11eb-9ce4-6a0bb506829b.png)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 500.0 USD (500.0 USD @ $1.0/USD)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/athensresearch/athens/717/100024990).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/athensresearch/athens/717/100024990).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
* $2,652,246.45 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

tangjeff0:
closed by #803

DrHyde:
Confirmed fixed, thanks!

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__The funding of 500.0 USD  (499.85 USD @ $1.0/USD) attached to this issue has been **cancelled** by the bounty submitter__

 
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
[...]

PR #803, opened 2021-03-11
fix(undo/redo): textarea undo/redo; italics, highlight, underline, strikethrough #717

arkRedM:


tangjeff0:
Woweee

arkRedM:
Screen Shot 2021-03-12 at 2 37 37 AM

Basic formatting and cmd z behavior are addressed in this patch. 
This is not related to undo/redo a txn at all. Maybe move the bounty to another issue?

tangjeff0:
@arkRedM oh you're that's a separate issue, however, undo/redo should be implemented as a tx log. The current approach holds onto the entire last 10 db values, and drops and resets on undo/redo. This is extremely slow for big dbs - takes ~2 seconds for mine at the point.

I also think it's the basis for RTC, in terms of using a tx log.

arkRedM:
> I also think it's the basis for RTC, in terms of using a tx log.

That's the reason I started with this issue as well 😆, quickly discovered that's unrelated. But I was already knee-deep, the only way out was to leave the patch is, in an ideal state -- hence added a few qol features while also fixing the root issue

src/cljs/athens/listeners.cljs
```diff
@@ -99,7 +99,7 @@
                                        KeyCodes.BACKSLASH (if shift
                                                             (dispatch [:right-sidebar/toggle])
                                                             (dispatch [:left-sidebar/toggle]))
-                                       KeyCodes.H (util/toggle-10x)
+                                       KeyCodes.T (util/toggle-10x)
```
tangjeff0, line 102: Great idea to use ctrl-T instead of ctrl-H. H always led to weird behavior.

tangjeff0 (approved):
CC #649 @dauberd @avidrucker

PR #808, opened 2021-03-13
perf(undo/redo): undo by transaction rather than entire db

arkRedM:
Significantly improves undo/redo perf with a simple history tracking mechanism

--- read comments from the patch before proceeding ---

To make it even more complete adding a buffer history queue such as https://github.com/clj-commons/ring-buffer can limit the size of the edit history keeping it at a constant maximum. Not sure if this adds much value, but if perf is dropping after a ton of edits we can come back to it then.


tangjeff0:
When I undo then redo, I get the following error:

image

src/cljs/athens/db.cljs
```diff
@@ -525,44 +525,24 @@
...
-(defn find-next
-  [xs pred]
-  (fnext (drop-while #(not (pred %)) xs)))
-
+#_(def ^:const history-limit 10)
 
+;; this gives us customization options
+;; now if there is a pattern for a tx then the datoms can be
+;; easily modified(mind the order of datoms) to add a custom undo/redo strategy
+;; Not seeing a use case now, but there is an option to do it
```
tangjeff0, line 533: Custom undo/redo logic: undo tree?

- https://github.com/mbbill/undotree
- https://www.emacswiki.org/emacs/UndoTree

src/cljs/athens/db.cljs
```diff
@@ -525,44 +525,24 @@
...
-                                      (conj db-after)
-                                      (trim-head history-limit))))))))
+             (when-not (or (->> tx-report :tx-data (some (fn [datom]
+                                                           (= (nth datom 1)
+                                                              :from-undo-redo))))
+                           (->> tx-report :tx-data empty?))
+               (swap! history (fn [buff]
+                                (->> buff (remove (fn [[_ applied? _]]
+                                                    (not applied?))))))
+               (swap! history conj [(-> tx-report :tx-data first vec (nth 3))
```
tangjeff0, line 543: When I undo then redo, I get the following error with `nth`:

image

Maybe here?

tangjeff0 (changes requested):

src/cljs/athens/db.cljs
tangjeff0: I will say it undo is a lot faster!

src/cljs/athens/db.cljs
arkRedM: I've imagined a scenario where an op can be done in one step but undone in two steps or there is an explicit confirm action warning mid-step. 

Undotree looks very cool. A very power-user feature.

src/cljs/athens/db.cljs
arkRedM: Regarding the bug -- lists and lazy seq gotchas!

Regarding rtc -- the idea of db + tx1 + tx2 gives the final state is something that we can leverage. I won't say this is the final design idea but this is a good naive starting point.  (there is no restriction tx1 and tx2 should come from the same source 😉)
Challenge in rtc will be understanding n states(for n users) and coming to a conflict-free state on the server-side and pushing that state in a sane manner to all users(show errors or copy old blocks when there is a conflict between server data and user data -- since the log is append-only we are bound to run into this)

src/cljs/athens/db.cljs
tangjeff0: Yeah, I've never personally used it either in Vim or Emacs, probably why it's a plugin. Though undotree is basically a visualization for versioning (at the block level).

This is Version Story, a startup in the current YC batch. I guess what they found is that versioning is really important for legal docs: https://www.versionstory.com. 

Git Tree viz is a similar concept.

tangjeff0 (approved):

Issue #814, opened 2021-03-13
Can't see recent files


Scatchell:
**Problem**
When I use the "Find or Create a Page" (cmd + K) window, at the bottom is says: "Recent files". I assume there is meant to be a list of recent files here, but it is not populated for me. When I press the down or up arrow key it displays the error:

```
message=Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null
source=file:///Applications/Athens.app/Contents/Resources/app.asar/resources/public/js/compiled/renderer.js
lineno=282
colno=483
error=TypeError: Cannot read property 'getBoundingClientRect' of null
```

I am a new user, so feel free to correct me if I'm just missing something. I was looking for functionality to navigate via keyboard through recent files, and the option + left or option + right arrow keys were not working for me, so I figured this was the next best thing.

**Screenshots/Demo**

![image](https://user-images.githubusercontent.com/1375869/111043246-6f492900-8439-11eb-82c8-914cd544871e.png)

**Athens Version**

Athens: Version 1.0.0-beta.56 (1.0.0-beta.56)
OS: MacOS Catalina 10.15.7 (19H524)

Using desktop version installed via DMG

tangjeff0:
This is a good bug. Nothing should happen if the user presses up/down.

Scatchell:
Cool! Just to add in case it is related: a similar exception also occurs when I try to tag an item by adding a hash to a word and then immediately hit up/down to try to navigate to the next block.

![image](https://user-images.githubusercontent.com/1375869/111145385-13d58300-8580-11eb-9564-40902f57b70c.png)

tangjeff0:
Good catch. Both of these use the same underlying code, so should be the same fix!

hewrin:
Mind if I take a stab at this?

hewrin:
The problematic functions seem to be `scroll-if-needed` and `is-beyond-rect?` in `athens.util.cljs`

Doing something like this seems to fix it

```
(defn is-beyond-rect?
  "Checks if any part of the element is above or below the container's bounding rect"
  [element container]
  (if (not (nil? element))
      (let [el-box (.. element getBoundingClientRect)
        cont-box (.. container getBoundingClientRect)]
    (or
      (> (.. el-box -bottom) (.. cont-box -bottom))
      (< (.. el-box -top) (.. cont-box -top))))))
 ```
 
But I think there might be a more elegant way. Open to suggestions!

phrohdoh:
[...]

PR #834, opened 2021-03-16
fix: do not allow user to navigate through or open results of an empty page search

phrohdoh:
note: I'll rebase this to drop the `tmp` commit after approval but before this is merged (assuming it is, of course)

---

fixes https://github.com/athensresearch/athens/discussions/827

this is currently deployed to my fork's gh-pages: https://phrohdoh.github.io/athens/


phrohdoh:
There is likely a better solution (that actually addresses the root cause[[1]][[2]]) than simply not invoking the problematic functions.

[1]: https://github.com/athensresearch/athens/issues/814#issuecomment-800402007
[2]: https://github.com/athensresearch/athens/issues/814#issuecomment-800998015

PR #876, opened 2021-03-25
feat(block-ref): shift + "Drag'n'Drop" to copy block uid

### LINKED ISSUE #572: feat(block-ref): Ctrl + "Drag'n'Drop" to copy block uid
Similar to Roam, the ability to Drag'n'Drop a Block while pushing the Ctrl-key could create a block reference for the selected block.

psidiumcode:
close #572

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 .idea/
 
 .clj-kondo/.cache
+.lsp/
 
 # cache directory for dependencies installed with deps.edn
 .cpcache
--- a/src/cljs/athens/events.cljs
+++ b/src/cljs/athens/events.cljs
@@ -1064,7 +1064,7 @@
 
 (defn drop-link-child
   "Create a new block with the reference to the source block, as a child"
-  [source source-parent target]
+  [source target]
   (let [new-uid               (gen-block-uid)
         new-string            (str "((" (source :block/uid) "))")
         new-source-block      {:block/uid new-uid :block/string new-string :block/order 0 :block/open true}
@@ -1077,8 +1077,8 @@
 
 (reg-event-fx
   :drop-link/child
-  (fn [_ [_ source source-parent target]]
-    {:dispatch [:transact (drop-link-child source source-parent target)]}))
+  (fn [_ [_ source target]]
+    {:dispatch [:transact (drop-link-child source target)]}))
 
 
 (defn drop-link-same-parent
@@ -1127,7 +1127,7 @@
 
 (defn drop-link-diff-parent
   "Add a link to the source block and reorder the target"
-  [kind source source-parent target target-parent]
+  [kind source target target-parent]
   (let [new-uid             (gen-block-uid)
         new-string          (str "((" (source :block/uid) "))")
         t-order               (:block/order target)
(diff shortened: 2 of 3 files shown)


psidiumcode:
The following PR enables creating block links uppon drag'n drop while pressing the `shift` key. 

https://user-images.githubusercontent.com/81385607/112547648-ae756380-8dc3-11eb-8edb-e4c4fc8e4934.mov

Note that the key used is the `shift` key, not the `ctrl` which is requested in the issue.

**Steps**

Create events for when blocks are dropped as reference links:

- `:drop-link/child`
- `:drop-link/same`
- `:drop-link/diff`

These are similar to the regular drop events, with the difference that the do not retract the block that is being dragged and instead creates a new block with `(( block-uid ))` as the content.

Create event triggers:

The events are triggered when a bullet is dropped, it uses the html property `DataTransfer.effectAllowed` with value `link` to represent a reference drag. 
When a block is dragged with the shift key pressed (detected using the `MouseEvent.shiftKey` html property), the `DataTransfer.effectAllowed` is set to `link`, otherwize it is set to `move`.

**Further improvements**

There is a lot of code repetition, specially in the drop events, while it is probably good to have the regular block drop and the block link drop as separate events, the code could improved by creating a generic function and passing the required arguments for the drop functions.

**Problems encountered**

The `shift` key was used because using any other key `ctrl`, `cmd` did not allow the bullet to be dropped. After releasing the mouse button while holding down the key, the bullet would return to it's original place. This is probably a style or similar which I could not find.

tangjeff0:
Hello @psidiumcode. Very impressive first PR. I used `ctrl` and it worked on Linux. Perhaps this is a Mac issue?

Second, do you intend to sign the CLA or no?

psidiumcode:
1. It could be a Mac issue with Electron, I found this old issue with the `DataTransfer.effectAllowed` property that mentions a similar behaviour to what I experienced https://github.com/electron/electron/issues/7207. A good way to test this would be to create a simple Electron project with draggable elements using the `DataTransfer.effectAllowed` properties like in this example: https://codepen.io/simoncodrington/pen/rOqPQV, to test the behaviours on all platforms.
2. The CLA is now signed

tangjeff0:
> It could be a Mac issue with Electron, I found this old issue with the DataTransfer.effectAllowed property that mentions a similar behaviour to what I experienced electron/electron#7207. A good way to test this would be to create a simple Electron project with draggable elements using the DataTransfer.effectAllowed properties like in this example: codepen.io/simoncodrington/pen/rOqPQV, to test the behaviours on all platforms.

Sounds good - I'll test this using your suggestion as well as on my Mac this week.

TillScout:
Nice, I have always hoped for this!
As a suggestion for improvement: Maybe add an additional modifier for creating a block embed.

tangjeff0:
After, testing this on my M1 Mac, it indeed does not work when replacing `shiftKey` with `ctrlKey`, `metaKey`, or `altKey`.


`shift` is acceptable for now! (Not a big deal, but the original feature request #572 says to use `ctrl`. For Mac in Roam, it is `alt`.) @psidiumcode not sure how you want to get paid out for this bounty. DM me on Discord or email jeff@athensresearch.org. We can also do crypto now.

PR #898, opened 2021-04-01
doc(backend): start architecture decision record for backend

agentydragon:
Iterating on this based on conversation with Abhinav and Matei.

Abhinav has been working on a Datahike backend with a shared server and always-assumed-online clients communicating over websocket. I think it would make sense to think a bit more about alternatives, because there might be options that could give us more for less buck.

- [ ] Add section about distributing/updating the server

### DIFF
--- a/doc/adr/README.md
+++ b/doc/adr/README.md
@@ -1,0 +1,1 @@
+Architectural Decision Records - see https://adr.github.io/.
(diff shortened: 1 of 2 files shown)


agentydragon:
I'm probably going to add a bunch of commits to the branch, I assume it'll get squashed on merge.

jsmorabito:
just consolidated all docs to this repo, can you make this pr there?

tangjeff0:
This is a different sort of document then the others. This might be better under a `specs/` folder, but still in handbook.

tangjeff0:
There may be a better home for this document.

PR #939, opened 2021-04-11
fix: Checking a TODO item in block embed results to error

### LINKED ISSUE #819: fix: Checking a TODO item in block embed results to error
**Problem**
Checking a TODO item in block embed results to error

To replicate:
1. Create a block with a todo
2. Create a block-embed of the block you made in step no. 1
3. Check the todo in the block-embed

**Screenshots/Demo**
image

**Athens Version**
1.0.0-BETA.60

juniusfree:
fix #819

**Problem:** The `uid` that is passed to `todo-on-click` is the updated one from `recursively-modify-block-for-embed`. Which means it won't be found when `db/get-block` is called.
**Solution:** Use the `original-uid`, if available, or `uid` if not.

athens-819

I have a question though. Will the `uid` from `recursively-modify-block-for-embed` be used in `parse-and-render`?

---
~~This other issue is not included here. I'll investigate further. https://github.com/athensresearch/athens/issues/819#issuecomment-811677701~~

Also fixes this comment https://github.com/athensresearch/athens/issues/819#issuecomment-811677701

**Problem:** The `uid` of the nested blocked is passed to `parse-and-render` instead of the `uid` of the parent block
**Solution:** Update the `uid` using the `uid` of the parent block
athens-819-2

### DIFF
--- a/src/cljs/athens/views/blocks.cljs
+++ b/src/cljs/athens/views/blocks.cljs
@@ -640,7 +640,7 @@
                            :on-click       (fn [e] (textarea-click e uid state))
                            :on-mouse-enter (fn [e] (textarea-mouse-enter e uid state))
                            :on-mouse-down  (fn [e] (textarea-mouse-down e uid state))}]
-       [parse-and-render local uid]
+       [parse-and-render local (or original-uid uid)]
        [:div (use-style (merge drop-area-indicator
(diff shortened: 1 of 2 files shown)


tangjeff0:
Good stuff @juniusfree :)

PR #949, opened 2021-04-11
refactor: use logical positioning styles

shanberg:
- Refactors (almost) all the positioning CSS to use logical styles to better support different writing directions
- Rename "left sidebar" to "nav sidebar"
- Rename "right sidebar" to "ref sidebar"


tangjeff0:
So we don't want to use "right" or "left" because in RTL languages these directions mean something else? This is coming from discussion at https://github.com/athensresearch/athens/discussions/935

tangjeff0:
Following up on my question @shanberg

shanberg:
@tangjeff0 Sorry about the wait, thought I'd already responded.

Yes that's the idea. CSS logical properties aren't 100% supported, and I expect we'll have to make a second pass when Athens is intended to run in most/all browsers, but those styles should be fallbacks for these values, not the other way around...unless we want to make this a little more robust and a lot more complicated.

tangjeff0:
This is the right thing to do, but it's a lot of changes, so I want to hold off on this until we do more research on what all needs to be done for LTR. Sorry @shanberg

PR #972, opened 2021-04-15
fix: non-formatted breadcrumb on page block

### LINKED ISSUE #970: Breadcumbs should be rendered
**Problem**
Breadcumbs should be rendered when markups are used like **bold**, _italic_ or highlighted !

**Screenshots/Demo**
image

**Athens Version**
1.0.0-BETA.68 - Intel OSX

juniusfree:
closes #970

image

### DIFF
--- a/src/cljs/athens/views/block_page.cljs
+++ b/src/cljs/athens/views/block_page.cljs
@@ -119,7 +119,7 @@
                ^{:key breadcrumb-uid}
                [breadcrumb {:key (str "breadcrumb-" breadcrumb-uid)
                             :on-click #(breadcrumb-handle-click % uid breadcrumb-uid)}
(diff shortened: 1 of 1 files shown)


tangjeff0:
@juniusfree I have a challenge for you:

Can we expand the breadcrumb instead of navigating to a link or block ref?

Demo: https://www.loom.com/share/1b8b8b59d75947538d62caca00945dfd

Note that clicking on a [[link]] navigates to that page, when we probably want to expand breadcrumb.

Note that clicking on a ((block-ref)) expands breadcrumb, but actually it first navigates to the original block first. See how I have to press `back` twice.

juniusfree:
@tangjeff0 I'll check that later. I'll let you know when I have a question.

PR #976, opened 2021-04-16
fix: expand the breadcrumb when [[link]], don't navigate

juniusfree:
**Current behavior:** See https://github.com/athensresearch/athens/pull/972#issuecomment-820496888
**Expected behavior:** When clicking the parsed block reference or page reference on the breadcrumb, Athens should navigate to the block page only
**Solution**: Add an invisible element above the parsed content to prevent it from getting clicked.

athens-972-comment

### DIFF
--- a/src/cljs/athens/views/block_page.cljs
+++ b/src/cljs/athens/views/block_page.cljs
@@ -4,7 +4,7 @@
     [athens.db :as db]
     [athens.parse-renderer :as parse-renderer]
     [athens.router :refer [navigate-uid]]
-    [athens.style :refer [color ZINDICES]]
+    [athens.style :refer [color]]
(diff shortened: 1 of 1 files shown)


tangjeff0:
What if there was just one `span`?

What if the one `span` just did `(.. e preventDefault)` `:on-click`?

juniusfree:
@tangjeff0 I can't make it work using `preventDefault` but I updated it with one `span` using `pointer-events`. It still works the same. I added the `relative` property because the `opacity` in hashtag and page-link components are messing up with `pointer-events`.

tangjeff0 (approved):

tangjeff0:
Great work @juniusfree  :)

PR #982, opened 2021-04-17
chore(electron): update from 11 to 12

solarkraft:
`contextIsolation` now defaults to true and needs to be flipped back explicitly, see https://www.electronjs.org/docs/breaking-changes#default-changed-contextisolation-defaults-to-true

Electron 12 enables native Wayland (future/current Linux application display system) support (using the flags `--enable-features=UseOzonePlatform --ozone-platform=wayland`), which means sharp text on HiDPI screens and direct (not forcibly smoothed/wobbly) scrolling (using the flag `--disable-smooth-scrolling`).


tangjeff0:
@solarkraft does this mean that you want Wayland support?

Sounds like we need to a flag in our build script to support Wayland, is that right?

solarkraft:
> does this mean that you want Wayland support?

Yep! 

> Sounds like we need to a flag in our build script to support Wayland, is that right?

Nope, Electron handles it as is. Actually _running_ it on Wayland doesn't work with the packaged Appimage however, since flags provided to the Appimage aren't passed on to Electron. 

A remaining issue preventing defaulting to Wayland is that on compositors that for some reason don't support server side decorations for Wayland apps (Gnome ...........) there won't be any decoration on the window. Fixing this would require client side decoration support, but it's definitely not something anyone would accidentally stumble over due to how explicitly opt in Wayland is in this form.

tangjeff0:
@solarkraft thanks for this. I've never really dug into display servers like X and Wayland, though I've mainly used Linux for the past few years. Luckily the Linux peeps generally seem to be good at solving their own problems :)

PR #991, opened 2021-04-19
fix: Delete a block and "undo" cannot bring it back

### LINKED ISSUE #987: Delete a block and "undo" cannot bring it back
#951 

> Has anyone encountered this bug/feature? Once I select a whole block (nodes included, with a blue shade over the block) and delete it. "Undo" will not bring back the block. The same goes with cutting the whole block (see my question here). It is a bit inconvenient if the deleted block has several references attached to it.

juniusfree:
fix #987 

**Current Behavior:** Given a deleted block/s (deleted by selection ie. the selection is indicated with blue highlight), when the user tries to undo the action, nothing happens.
**Expected Behavior:** Given a deleted block/s (deleted by selection ie. the selection is indicated with blue highlight), when the user tries to undo the action, the deleted block/s should be brought back.

**Problem:** The value of `:editing/uid` is the `uid` of the deleted block (or one of the deleted blocks if multiple blocks are deleted) which makes Athens "freeze" (I'm not sure what's the right word here). Undo for deleted blocks actually works when you click anywhere on Athens before trying to undo.
**Solution:** Update the `:editing/uid` to `nil` after deleting the selected blocks so the system won't "freeze".

athens-987

**Note:** When the deleted block is used in block embed, the block embed is not brought back after the undo because the `uid` of the block embed becomes "invalid" instead of the `uid` of the deleted block. I'll try to fix it separately.

athens-987-block-embed

### DIFF
--- a/src/cljs/athens/events.cljs
+++ b/src/cljs/athens/events.cljs
@@ -420,8 +420,9 @@
           retract-vecs      (mapcat #(retract-uid-recursively %) sanitize-selected)
           reindex-last-selected-parent (delete-selected sanitize-selected)
           tx-data           (concat retract-vecs reindex-last-selected-parent)]
-      {:dispatch [:transact tx-data]
-       :db       (assoc db :selected/items [])})))
+      {:fx [[:dispatch [:transact tx-data]]
+            [:dispatch [:editing/uid nil]]]
+       :db (assoc db :selected/items [])})))
 
 
 ;; Alerts


tangjeff0:
Really great catch @juniusfree! Revisiting `athens.listeners` code, I see that `editing-uid` has to be false for the datascript undo to work.

juniusfree:
@tangjeff0 Thanks! But honestly, I got lucky. I realized I didn't even check `athens. listener`.

My conclusion is wrong that it's not working because `:editing-uid`'s value is a deleted block. It's not working because it should be `nil`.

I wonder if there are other ways to fix this. I might try later.

PR #993, opened 2021-04-20
enhance(db-picker): Polish db dialog

shanberg:
- Replace Material UI tabs with new picker
- Propertly portal DB dialog(!!)
- Added shadow color to theme
- Minor tweaks to dialog style

Screen Shot 2021-04-19 at 11 33 08 PM

### DIFF
--- a/src/cljs/athens/views/filesystem.cljs
+++ b/src/cljs/athens/views/filesystem.cljs
@@ -16,6 +16,8 @@
     [athens.views.textinput :as textinput]
     [athens.ws-client :as ws-client]
     [cljs.reader :refer [read-string]]
+    [cljsjs.react]
+    [cljsjs.react.dom]
     [clojure.edn :as edn]
     [datascript.core :as d]
     [komponentit.modal :as modal]
@@ -53,7 +55,7 @@
    :align-self "stretch"
    :overflow "hidden"
    :transition "box-shadow 0.2s ease, filter 0.2s ease;"
-   :background (color :background-plus-1)
+   :background (color :background-color)
    :padding "1px"
    ::stylefy/manual [[:&:hover {}]
                      [:button {:text-align "center"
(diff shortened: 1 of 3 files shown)


Limezy:
Looks great 👍

src/cljs/athens/views/filesystem.cljs
```diff
@@ -38,6 +44,40 @@
...
+                               :padding "0.5rem 0.5rem"
+                               :color "inherit"
+                               :display "flex"
+                               :justify-content "center"
+                               :align-items "center"
+                               :position "relative"
+                               :z-index "0"
+                               :background "inherit"}
+                      [:svg {:margin-inline-end "0.25em" :font-size "1.25em"}]
+                      [:&:hover {:filter "contrast(105%)"}]
```
tangjeff0, line 74: There's an effect when hovering over an already active tab, but no effect when hovering over an inactive tab (including no `cursor: pointer`). Should there be one?

src/cljs/athens/views/filesystem.cljs
```diff
@@ -217,32 +257,43 @@
...
-                      [join-remote-comp remote-graph-conf]
-
-                      (= 1 (:tab-value @state))
-                      [create-new-local state]
-
-                      (= 0 (:tab-value @state))
-                      [open-local-comp loading db-filepath])]
-         :on-close close-modal}]])))
+      (js/ReactDOM.createPortal
+        (r/as-element [:div (use-style modal-style)
```
tangjeff0, line 268: Wait did we finally get React Portal to work? So we can now use this pattern for tooltips, dropdowns, etc.?

https://discord.com/channels/708122962422792194/800579670629679165/833918121224044573

tangjeff0 (approved):

src/cljs/athens/views/filesystem.cljs
tangjeff0: Created a PR for this https://github.com/shanberg/athens/pull/4/files

src/cljs/athens/views/filesystem.cljs
shanberg: There is some styling, but it can be clearer. I want to make a more focused pass on all the control components to make sure they all have all the right styles, and creating additional use cases like this picker will help do that.

PR #994, opened 2021-04-20
feat(app-toolbar): basic tooltips

shanberg:
Just using :title attribute here.

I'm looking into more detailed tooltips with keyboard shortcut hints and better style, but this is a start and very easy.


tangjeff0:
Didn't even realize there was a `title` attribute for buttons

PR #995, opened 2021-04-20
fix: Shift+click title of reference won't open it in the right sidebar

### LINKED ISSUE #958: ⇧ + clic on linked page title won't open it in the side bar
**Problem**
I expect that every time I ⇧ + clic on a link, that will open the link in the side bar.
This works well within the search bar, and also within the main page.
It also works for any link inside a linked reference, but not in the title !

**Screenshots/Demo**
Capture d’écran 2021-04-13 à 13 42 38

**Athens Version**
1.0.0-BETA.68 - Intel OSX

juniusfree:
fix #958 

**Current behavior:** Given a linked/unlinked reference, when the user shift+click on the title, the page won't open to the right sidebar
**Expected behavior:** ...the page should open to the right sidebar

**Problem:** The `e` is not given which means the `navigate-uid` function won't check for `shift` key event.
**Solution:** Add the `e` argument
athens-958


Limezy:
Well done @juniusfree 👍

PR #1012, opened 2021-04-22
fix: allow for multiple block-refs in a block.

### LINKED ISSUE #1010: Multiple block-ref in a block is not parsed.
**Problem**
- Multiple block-ref in a block is not parsed.

**Screenshots/Demo**
image

**Athens Version**
1.0.0-BETA.71

neotyk:
Fix #1010 by making block-ref parsing not greedy.


tangjeff0 (approved):
So why are there two test files that have the same tests?

Issue #1030, opened 2021-04-23
Bugs found for Find or Create input


BenchLord:
As you can see in the video there are a few interesting things
1) Find or Create input gives nasty error when no notes exist
2) If a daily note exists the Find or Create form will take you there if `Enter` is pressed fast enough. This happens even if you type in the correct note name in the form.

Not bugs necessarily but other few things:
1) Empty bullets stick around and are shown on the Body column of the All Pages page
2) The option menu for notes (add/remove shortcut, show local graph, delete page) should also be available on the All Pages page

https://user-images.githubusercontent.com/6004385/115935377-4b253280-a450-11eb-9afc-3600e7899326.mov

avidrucker:
@BenchLord Thank you for the clear explanation and demo video.

sid597:
- [x] Find or Create input gives nasty error when no notes exist : Could not reproduce in beta.89 can be tested in the athens demo [here](https://athensresearch.github.io/athens/)
- [x] If a daily note exists the Find or Create form will take you there if Enter is pressed fast enough. This happens even if you type in the correct note name in the form. : This was resolved by @juniusfree 
- [x] Empty bullets stick around and are shown on the Body column of the All Pages page : fixed by @shanberg
- [X] The option menu for notes (add/remove shortcut, show local graph, delete page) should also be available on the All Pages page : More like a feature request, can create a new discussion around this, add/remove shortcut is already mentioned in the discussions.

PR #1040, opened 2021-04-24
enhance(all-pages): hide empty block bullets in all pages view

shanberg:
- Slightly refactored styles in `all-pages` table for easier dev usability
- Fixed some typos in `all-pages` styles, restoring some minor visual polish
- Decreased opacity of block marker bullets in block preview in all-pages items
- Hide empty blocks in all-pages block previews (solving part of #1030)


tangjeff0 (approved):
1. Do you think we should render markup
image

2. Do you think we should paginate? The reason we don't is that datalog queries are such that querying for all `n` pages takes the same amount of time as querying for 5 pages. However, rendering all the DOM components is still quite expensive, which is why this can be slow.

shanberg:
> 1. Do you think we should render markup
Yes, but I want to use this as an opportunity to think about how to separate block-level markup (headings, embeds, images, todos) from inline markup (bold, italic, links). I think we should render inline markup, but not block markup.

> 2. Do you think we should paginate?
Paginating is rational here because the list is unbounded, but I'd prefer to virtualize the list instead, if possible.

tangjeff0:
What does virtualize the list mean?

shanberg:
@tangjeff0 render only what's visible:
- https://github.com/bvaughn/react-window
- https://react-window.vercel.app/#/examples/list/fixed-size

This is probably a good approach for cases where we can provide our own on-page search, like the All Pages list, and not a good approach for rendering actual pages, because ctrl-f search won't work if the content isn't in the DOM.

PR #1044, opened 2021-04-26
fix: use index.transit file if already exists

DominicChiampi:
**Current behavior:** When dev athens is launched for the first time, an `index.transit` file is created, overwriting the existing `index.transit` created by the production athens app.
**Expected behavior:** If dev athens is launched and does not have a filepath in local storage, it should check for the `index.transit` file at the default location

**Problem:** Dev athens and production athens use different instances of chromium, so they have different localstorages. App startup assumes that if there is no filepath in localstorage, it is absolutely the first time athens has been run, so it creates a new database at the default location, overwriting the existing database.
**Solution:** At startup, check for an `index.transit` file at the default location before creating a new database.


tangjeff0 (approved):
Kudos for your first PR. Avoids data loss too, which should be engineering principle #1 for Athens.

PR #1045, opened 2021-04-27
rfct(node-page): replace nodepage dropdown with popover

shanberg:
- Replace dropdown code with MUI Popover component for improved behavior and control
- Duplicates some dropdown styles, scoped to the node-page. These styles should be moved back up to the main dropdown styles once this work is replicated across the app
- Very small change to button active style, to reduce button attention-gettingness when dropdown open


tangjeff0:
Much cleaner. Looking forward to refactoring the rest! Might make sense to even make `m-popover` part of the `dropdown` component.

Issue #1016, opened 2021-04-22
Increase visual signal of "hovered" state of arrows and bullets next to blocks


shanberg:
Sometimes it’s hard to tell what your cursor is over, and sometimes the cursor completely covers the arrow or bullet, hiding the current indicator.

avidrucker:
-[ ] mark this with the `ux` label

tangjeff0:
Please begin building habit of using `fix issue` or other keyword. Thank you @shanberg https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #1046, opened 2021-04-27
feat(blocks): improved visibility of hover/focus of block bullets

shanberg:
- This PR includes changes to the base light theme
    - Slightly darker base background color(!)
    - Improved color scaling from base background to lighter backgrounds
- Addresses Improved visual indication of hover of bullets and toggles #1016
- Slightly more clickable area given to block bullets
  - Note: highlight is slightly larger than actual mouseable area

Screen Shot 2021-04-26 at 9 40 26 PM


tangjeff0:
Much more clickable. Thanks Stuart!

Issue #977, opened 2021-04-16
Daily Pages scroll is slow


mikejones3:
**Problem**

When on the daily pages and scrolling to another day it is pretty slow to load on both my mac M1 and Windows laptop

**Athens Version**

1.0.0-beta.67

agentydragon:
I think https://github.com/athensresearch/athens/pull/1061 should make this a bit better. No fancy optimizations, just making the past pages start loading a bit earlier than they currently do.

juniusfree:
Fixed in https://github.com/athensresearch/athens/pull/1136

PR #1061, opened 2021-04-27
enhance: Trigger loading of daily pages more eagerly

agentydragon:
Fixes https://github.com/athensresearch/athens/issues/977.

We currently only load new daily pages when user scrolls to the very
bottom of the container also containing the "Earlier" header, and the
loading is also debounced by 500 ms, which adds a 500 ms delay before
the earlier pages actually start loading.

This commit shortens the debounce delay to just 100 ms, and triggers
the loading of earlier pages above the "Earlier" heading. Together this
should make loading of past pages feel a bit smoother, and it does seem
a bit smoother for me.

### DIFF
--- a/src/cljs/athens/views/pages/daily_notes.cljs
+++ b/src/cljs/athens/views/pages/daily_notes.cljs
@@ -47,6 +47,9 @@
 
 (defn scroll-daily-notes
   [_]
+  ; Load more daily notes if bottom of viewport is <1 pixel above bottom of
+  ; #daily-notes. The #daily-notes element contains all daily notes, but not
+  ; the "Earlier" banner.
   (let [daily-notes @(subscribe [:daily-notes/items])
         el          (getElement "daily-notes")
         offset-top  (.. el -offsetTop)
@@ -62,7 +65,7 @@
       (< bottom-delta 1) (dispatch [:daily-note/next (get-day (uid-to-date (last daily-notes)) 1)]))))
 
 
-(def db-scroll-daily-notes (debounce scroll-daily-notes 500))
+(def db-scroll-daily-notes (debounce scroll-daily-notes 100))
 
 
 (defn safe-pull-many
@@ -92,12 +95,15 @@
       (if (empty? @note-refs)
         (dispatch [:daily-note/next (get-day)])
         (let [notes (safe-pull-many @note-refs)]
-          [:div#daily-notes (use-style daily-notes-scroll-area-style)
-           (doall
-             (for [{:keys [block/uid]} notes]
-               ^{:key uid}
-               [:<>
-                [:div (use-style daily-notes-page-style)
-                 [node-page/page [:block/uid uid]]]]))
+          [:div (use-style daily-notes-scroll-area-style)
+           ; When user scrolls to the bottom of the #daily-notes div, daily
+           ; notes from more distant past are loaded.
+           [:div#daily-notes
+            (doall
+              (for [{:keys [block/uid]} notes]
+                ^{:key uid}
+                [:<>
+                 [:div (use-style daily-notes-page-style)
+                  [node-page/page [:block/uid uid]]]]))]
            [:div (use-style daily-notes-notional-page-style)
             [:h1 "Earlier"]]])))))


tangjeff0:
Thoughts on going from 500 ms to 100 ms @shanberg ? This feels a lil too fast :smile:

agentydragon:
I'm not sure why the debounce is there anyway. If it's to avoid doing the scroll area test too frequently because it's slow (which I don't think it is), then I think lowering it to 100 ms should still be fine. If the debounce is there to artificially introduce a delay (i.e., "past page appears 500 ms after you scroll to bottom of your daily note"), I don't actually see why would we want that.

Is there any UX reason why we would want there to be a delay? From my point of view I'd be totally happy getting a lazily loading infinite scroll where I'd never see a "Earlier" header.

agentydragon:
(sidenote: I'd also like to get rid of `:min-height "calc(100vh - 10rem)"`, I don't see it as providing me any value. but that's more a "style customization" issue - https://github.com/athensresearch/athens/discussions/835)

shanberg:
As I recall, the design for this area of Athens happened in about 10 minutes, on a whim, and it was never fully thought through or strongly tested, but a few points are worth reviewing:

1. I added a min-height to help users focus on a single Daily Note at a time, and to provide a clean fresh space for new notes and thoughts. This is fit for a specific use pattern, and may not be appropriate long-term, but I like it.
2. Delaying the loading of a new Daily Note (via debouncing, or requiring a little push beyond the scroll-to-bottom, etc.) was intended to prevent accidentally scrolling too much and creating a ton of empty Daily Notes which you had no intention of using. If that's still valuable and still function with 100ms, I'm fine with this change.

I don't have strong feelings about how this proceeds, only that we don't lose the baby in the bathwater by accident or for lack of context.

src/cljs/athens/views/pages/daily_notes.cljs
```diff
@@ -92,12 +95,15 @@
...
+          [:div (use-style daily-notes-scroll-area-style)
+           ; When user scrolls to the bottom of the #daily-notes div, daily
+           ; notes from more distant past are loaded.
+           [:div#daily-notes
+            (doall
+              (for [{:keys [block/uid]} notes]
+                ^{:key uid}
+                [:<>
+                 [:div (use-style daily-notes-page-style)
+                  [node-page/page [:block/uid uid]]]]))]
```
tangjeff0, line 107: Why create another div?

tangjeff0: bump @agentydragon

agentydragon: sorry about the delay.

- both the daily pages, and the "Earlier" header below, still need to be wrapped in `daily-notes-scroll-area-style`
- but in this PR I'm trying to get the reload to start earlier - so I moved the `daily-notes` ID to a container that contains *only* the daily pages, but not also the "Earlier" header. 

does the make sense?

agentydragon: > 2. Delaying the loading of a new Daily Note (via debouncing, or requiring a little push beyond the scroll-to-bottom, etc.) was intended to prevent accidentally scrolling too much and creating a ton of empty Daily Notes which you had no intention of using. If that's still valuable and still function with 100ms, I'm fine with this change.

Yeah, if you just hold down `PgDn`, this PR will create pages at ~5x the speed as before, but I don't think that's a common enough problem to avoid it at the cost of slowing down page loading that much.

agentydragon:
> 2. Delaying the loading of a new Daily Note (via debouncing, or requiring a little push beyond the scroll-to-bottom, etc.) was intended to prevent accidentally scrolling too much and creating a ton of empty Daily Notes which you had no intention of using. If that's still valuable and still function with 100ms, I'm fine with this change.

Yeah, if you just hold down `PgDn`, this PR will create pages at ~5x the speed as before, but I don't think that's a common enough problem to avoid it at the cost of slowing down page loading that much.

src/cljs/athens/views/pages/daily_notes.cljs
tangjeff0: I think simply changing the time should be good enough? @agentydragon

agentydragon:
Ok, I'll shelve the other change for now.

Issue #840, opened 2021-03-17
Self-Hosted Real-Time Collaboration


jsmorabito:
- Related: #930 #1075 #1074 https://github.com/athensresearch/athens/discussions/1082

# Issues ([Project Board](https://github.com/athensresearch/athens/projects/16))

- [ ] fix bad state that occurs from multiple writers #1140
- [ ] importing existing index.transit into backend db #1101
- [ ] parameterize `token` #1100
- [ ] presence UI: the "Online" functionality is wrong - who is on the block and page
- [ ] auto-reconnect. if computer goes to Sleep, have to manually reconnect #1102
- [ ] merging athens-backend into main athens and archiving athens-backend
- [ ] set HTTP ports in a straightforward way
- [ ] set port for nREPL
- [ ] fix hard code for filepath
- [ ] fix undo/redo

# Live Server

Setup a remote EC2 instance. Anyone can connect with the following credentials:

`address`: ec2-3-16-89-123.us-east-2.compute.amazonaws.com
`token`: x

arkRedM:
>  merging athens-backend into main athens and archiving athens-backend

why?

Other issues are in https://github.com/athensresearch/athens/discussions/1082

tangjeff0:
> why?

We want to keep front and backend changes synced up right now as a monorepo.

arkRedM:
I agree it can ease things that way.
But there is not much interaction between the two systems other than data format.
Things I want to point out.

Why split?
1. too many unnecessary commits for someone looking at one of backend or frontend
2. Simple separation of concerns 
3. Independent evolution
    a. changing version of deps without fear of breaking something in other system
    b. Focusing/attracting the right dev to the right place easily.
    c. Optimizations are different etc.

Why monorepo?
1. Dev workflow(specifically with PRs. not much effect on working on each)
2. Shared functionality 

Other than this, lein can easily handle combining without much tweaking.

tangjeff0:
Zooming out to a higher-level architecture view, there are two broad kinds of models:

- Single-writer (e.g. Datomic): front-end datascript db is read-only, backend single writer will reject new transactions if the last-tx-id is wrong (because the constructed tx-data might be wrong)
- Multiple-writers (e.g. [Web After Tomorrow](https://tonsky.me/blog/the-web-after-tomorrow/)): client is always responsive to writes, and somehow conflicts are all eventually resolved even if offline mode (Tonsky calls offline a special case of "infinite lag")

[...]

PR #1074, opened 2021-04-28
Self hosted ADR

neotyk:
Proposed Self Hosted ADR.

### DIFF
--- a/doc/adr/0003-self-hosted.md
+++ b/doc/adr/0003-self-hosted.md
@@ -1,0 +1,86 @@
+# 3. Self Hosted
+
+Date: 2021-04-27
+
+## Status
+
+Proposed
+
+## Context
+
+We want to enable enterprise customers to self host Athens knowledge graph on premise.  
+Uses should be able to collaborate on same graph, edit block & pages as well as see presence of other collaborators.
+
+## Decision
+
+We need a backend server that at least talks Datalog.
+
+For 1st backend we've chosen Datahike, because it's closeness to Datascript that we already use.  
+Another possible backend could be Crux or Datomic.
+
+### Communication Layer
+
+Because we're running in a very controlled environment (as in we don't have to support Samsung Internet & FB embedded browser)  
+we can assume availability of modern Web APIs, so for client server communication we can rely purely on WebSockets.
+
+### Concurrency
+
+To avoid problems of concurrent updates, we'd like to have single writer approach.  
+This means that updates to central graph can be done synchronously.
+
+### Connecting clients
+
+When client connects we have to get it in sync with backend.  
+
+We have 2 options here:
+
+1. Stream all the updates that happened since last tx-id client has. (easier to do because we'll need tx ingest on client for normal connected work)
+2. Provide a snapshot of DB state. (more effective on busy DB)
+
+We also have to authorize user.  
+For now we can assume trust based system, so if you say you're "Plato" we believe you w/o verification.  
+Authentication is a topic that deserves ADR of it's own.
+
+### Updating sever
+
+When Socrates makes a modification to connected DB (create, update, delete)  
+this change has to send to connected backend to be transacted.
+
+Changes that Socrates made should not be transacted locally, they should only be transacted on backend.
+
+### Updating clients
+
+Whenever transaction is executed on the backend it should be broadcasted to connected clients.  
+Because of potential `:db/id` clashes we'll need to normalize transaction data for transport (datsync is a good example of it).
+
+Client on receiving transaction data, transacts it to local database, and the change is reflected in UI.
+
+### Presence
+
+We also want to communicate to Socrates what other collaborators are present in his context.
+
+To do that we want to broadcast presence related information via same WebSocket we have for DB.
+
+Presence event types:
+* connect/disconnect
+* viewing page
+* editing block
+
+### Distribution
+
+Docker and Docker Compose are convenient for consumption.
+
+### Persistence
+
+Graph DB should be persisted to some form of permanent storage, best if it can store data in Docker volume so clients can easily manage their backups.
+
+## Consequences
+
+Currently Athens has only Datascript DB in the frontend, now that we're introducing connected (self-hosted) DB, we need to change that.
+
+If Athens is connected to self-hosted instance only place where DB modifications should happen is [on the server](#Updating Clients).  
+This means that our Frontend DB stays, and is still used for UI, but in read only mode.
+
+So when users make an edit it has to be transacted on the backend, and in order for edit to show up in users UI is by receiving broadcast  
+from the server and transacting it to local DB.
+
--- a/doc/adr/0002-staged-parser.md
+++ b/doc/adr/0002-staged-parser.md
@@ -5,7 +5,7 @@ Date: 2021-04-04
 
 ## Status
 
-Proposed
+Accepted.
 
 
 ## Context


doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+## Context
+
+We want to enable enterprise customers to self host Athens knowledge graph on premise.  
+Uses should be able to collaborate on same graph, edit block & pages as well as see presence of other collaborators.
+
+## Decision
+
+We need a backend server that at least talks Datalog.
+
+For 1st backend we've chosen Datahike, because it's closeness to Datascript that we already use.
```
agentydragon, line 18: its

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+When client connects we have to get it in sync with backend.  
+
+We have 2 options here:
+
+1. Stream all the updates that happened since last tx-id client has. (easier to do because we'll need tx ingest on client for normal connected work)
+2. Provide a snapshot of DB state. (more effective on busy DB)
+
+We also have to authorize user.  
+For now we can assume trust based system, so if you say you're "Plato" we believe you w/o verification.  
+Authentication is a topic that deserves ADR of it's own.
```
agentydragon, line 42: its

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+
+When client connects we have to get it in sync with backend.  
+
+We have 2 options here:
+
+1. Stream all the updates that happened since last tx-id client has. (easier to do because we'll need tx ingest on client for normal connected work)
+2. Provide a snapshot of DB state. (more effective on busy DB)
+
+We also have to authorize user.  
+For now we can assume trust based system, so if you say you're "Plato" we believe you w/o verification.
```
agentydragon, line 41: the problem of deciding "is the user actually Plato as they claim" is authentication. "is the user allowed to access this resource" is authorization.

(see e.g. https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-vs-authorization)

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+
+Because we're running in a very controlled environment (as in we don't have to support Samsung Internet & FB embedded browser)  
+we can assume availability of modern Web APIs, so for client server communication we can rely purely on WebSockets.
+
+### Concurrency
+
+To avoid problems of concurrent updates, we'd like to have single writer approach.  
+This means that updates to central graph can be done synchronously.
+
+### Connecting clients
```
agentydragon, line 31: this does not seem to mention any way to verify the client is allowed to access the server at all. we should have at least some simple security for that. shared secret would be enough for now. we just shouldn't allow anyone who can access the IP & port to read/write.

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+2. Provide a snapshot of DB state. (more effective on busy DB)
+
+We also have to authorize user.  
+For now we can assume trust based system, so if you say you're "Plato" we believe you w/o verification.  
+Authentication is a topic that deserves ADR of it's own.
+
+### Updating sever
+
+When Socrates makes a modification to connected DB (create, update, delete)  
+this change has to send to connected backend to be transacted.
```
agentydragon, line 47: has to be sent

agentydragon (changes requested):

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+
+We want to enable enterprise customers to self host Athens knowledge graph on premise.  
+Uses should be able to collaborate on same graph, edit block & pages as well as see presence of other collaborators.
+
+## Decision
+
+We need a backend server that at least talks Datalog.
+
+For 1st backend we've chosen Datahike, because it's closeness to Datascript that we already use.  
+Another possible backend could be Crux or Datomic.
```
tangjeff0, line 19: Datalevin has also entered the chat. @den1k

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+
+When Socrates makes a modification to connected DB (create, update, delete)  
+this change has to send to connected backend to be transacted.
+
+Changes that Socrates made should not be transacted locally, they should only be transacted on backend.
+
+### Updating clients
+
+Whenever transaction is executed on the backend it should be broadcasted to connected clients.  
+Because of potential `:db/id` clashes we'll need to normalize transaction data for transport (datsync is a good example of it).
```
tangjeff0, line 54: cc @metasoarous

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+To do that we want to broadcast presence related information via same WebSocket we have for DB.
+
+Presence event types:
+* connect/disconnect
+* viewing page
+* editing block
+
+### Distribution
+
+Docker and Docker Compose are convenient for consumption.
```
tangjeff0, line 71: I thought we didn't need Docker because we have .jar

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Persistence
+
+Graph DB should be persisted to some form of permanent storage, best if it can store data in Docker volume so clients can easily manage their backups.
```
tangjeff0, line 75: Either filesystem or postgres, which can be attached to Crux, Datahike, Datomic.

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+### Presence
+
+We also want to communicate to Socrates what other collaborators are present in his context.
+
+To do that we want to broadcast presence related information via same WebSocket we have for DB.
+
+Presence event types:
+* connect/disconnect
+* viewing page
+* editing block
```
tangjeff0, line 67: Athens should be read-only when disconnected. Should not be allowed to edit any textareas. Can still navigate though I suppose.

tangjeff0: Dot should also be green/red for connected/disconnected, and yellow while syncing.

tangjeff0: Anything else? cc @shanberg

tangjeff0 (commented):
Additional considerations that were not discussed:

* Can now run a web app in addition to desktop app. This is helpful because an on-prem user wold have to only build the frontend once now, rather than on each client. (On-Prem users have their own codebase that has some modifications)
* Backend code to be merged into frontend as a monorepo so front and back don't have to version each other.
* What config should be available on backend. There's `token`, anything else?

den1k (commented):
Looks like there is no story to avoid or automatically resolve conflicts between multiple clients / users, meaning last write wins?

doc/adr/0003-self-hosted.md
```diff
@@ -0,0 +1,86 @@
...
+To avoid problems of concurrent updates, we'd like to have single writer approach.  
+This means that updates to central graph can be done synchronously.
+
+### Connecting clients
+
+When client connects we have to get it in sync with backend.  
+
+We have 2 options here:
+
+1. Stream all the updates that happened since last tx-id client has. (easier to do because we'll need tx ingest on client for normal connected work)
```
den1k, line 37: with multiple users, there will be duplicate tx-id's

doc/adr/0003-self-hosted.md
shanberg: @tangjeff0 Looks like this is mostly about the presence of other folks, not the sync status of the DB, but with regard to the presence stuff: hard to say without using it for a bit, but I know there are some changes I'd like to make in the near term to improve readability and accessibility
- Add icons to differentiate without color
- Adjust green for better visibility to red/green colorblind folks

tangjeff0:
Closing for now, continue discussion on #1082 or #840. Can re-open as we discover more here.

PR #1075, opened 2021-04-28
durability and consistency with second db

arkRedM:
Setting up context, the first part of rtc that you are seeing in athens-backend and app is a way to stream transactions. We used
1. datahike as single write
2. datascript as client db
3. ws to stream txns

But the current one won't cut it(my past self who thought this was enough is will be surprised :sweat_smile: ). More specifically the data flow is not enough to ensure data durability and consistency.

1. Durability - https://www.loom.com/share/2a73776b2c42489b9e4df80aaa94acfe
2. Consistency - https://www.loom.com/share/8613cc912cb44d7482125b1f0caee3a3


tangjeff0:
Please use https://github.com/athensresearch/athens/discussions/1082 for discussion, because it includes both frontend and backend.

tangjeff0:
Closing for now, discussion at #840 and #1082

arkRedM:
Need to rethink some data flow issues. Closing for now

tangjeff0:
Do you want to share what you are thinking in public? May be able to hop in and help out e.g. https://github.com/julionav/athens/pull/1#issue-647664768 by @julionav

arkRedM:
Discovered some bad data flow issues. 

- there were some weird issues with datahike which resolved once I updated to latest(will look into it a little deeper) 
- Second db(remote sync db) isn't great for memory 
- We don't specially transact backend corrections(by invariants) to view db(potentially result in weird state when we do).  
- Not exactly sure but there are issues with datsync client implementation with syncing with new ids(remote id and local id are not getting linked right).

PR #1093, opened 2021-04-30
fix: when block-embed is deleted, replace with content instead of "invalid"

juniusfree:
Expected Outcome: Given a block embed, when the user deleted the referred block, then the block embed should display the `uid`.
Current Outcome: ...the block embed uid becomes `invalid`

Problem: The block embed is set to show `invalid`
Solution: Show the `content` instead of invalid
Not solved yet: Users need to refresh or re-edit the block-embed to re-render it.
athens-1093


tangjeff0 (approved):
good fix

PR #1094, opened 2021-04-30
fix: Removal of a daily journal page creates a 404

### LINKED ISSUE #1022: Removal of a daily journal page creates a 404 error
**Problem**
Removal of a daily journal page creates a 404 error, instead of returning to the page list as usual

**Screenshots/Demo**
image

**Athens Version**
1.0.0-BETA.74 - Intel OSX 10.11

juniusfree:
fix #1022 (reported in v1.0.0-beta.74, reproduced in v1.0.0-beta.78 as well)

expected outcome: Given a page, when the user deletes it, then Athens should re-direct to the list of pages
current outcome: ...Athens re-directs to 404

1. For daily notes page
Problems: No re-direct to the list of pages
Solution: Ensure that Athens will navigate to the list of pages of deleting a page

athens-1022-daily-del

2. For non-daily notes page
Problem: `state` is passed to `menu-dropdown` instead of `daily-note?` as a 2nd argument which makes the check for `daily-note?` always true.
Solution: Remove the `state` argument

athens-1022-page-del

### DIFF
--- a/src/cljs/athens/views/pages/node_page.cljs
+++ b/src/cljs/athens/views/pages/node_page.cljs
@@ -395,11 +395,11 @@
                       [:> BubbleChart]
                       [:span "Show Local Graph"]]]]
                    [:hr (use-style menu-separator-style)]
-                   [button {:on-click #(if daily-note?
-                                         (dispatch [:daily-note/delete uid title])
-                                         (do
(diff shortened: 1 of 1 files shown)


tangjeff0 (approved):
good catch!

Issue #756, opened 2021-03-06
perf(search): faster search for (()) and ctrl-k


tangjeff0:

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 300.0 USD (300.0 USD @ $1.0/USD)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/athensresearch/athens/756/100024989).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/athensresearch/athens/756/100024989).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
* $2,651,746.45 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

KingMob:
Could someone add a bit of a description? This is pretty vague, at least to me.

den1k:
some thoughts:

Looks like [this code](https://github.com/athensresearch/athens/blob/454682547046109d0cea3512d17d4fb3095397ec/src/cljs/athens/db.cljs#L443-L459) is the culprit
  - both `d/q` and `d/pull` are known to be slow
  - 1st improvement: avoid `d/q` and filter/search `:v` in `(d/datoms @conn :avet :block/string)` instead
    - use [`d/pull-many` on the result](https://www.notion.so/vimsical/Product-Search-in-Clojure-DataScript-3d467efc55244f3693127f8d9da5f262#e12a4f5ec73d477d94d17dc03346e270)
  - even better: use `d/datoms` and `d/entity`. Q here is how come d/entity is not used everywhere in the UI? It's lazy and way faster than pull.

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 265 years, 6 months from now.
Please review their action plans below:


**1) [juniusfree](https://gitcoin.co/juniusfree) has started work.**

1. Update the search queries
2. Use transducers
3. Other refactors

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/athensresearch/athens/756/100024989).

gitcoinbot:
[...]

PR #1095, opened 2021-05-01
feat: fuzzy search

juniusfree:
Posting this for feedback especially for performance improvement and/or better approach to fuzzy search.

Trying to solve https://github.com/athensresearch/athens/issues/756 + better search result

Currently tested in Athena only.

In terms of speed, search result returns in <200ms on my PC and db that's why the `debounce` is reduced.

More importantly, this will help the user find the notes that they're looking for even if they provided an inexact phrase.
image
image

The code is adapted from https://gist.github.com/vaughnd/5099299

TODO
- [ ] add to `(())`
- [ ] include exact search for blocks

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -2,6 +2,7 @@
   (:require
     [athens.patterns :as patterns]
     [athens.util :refer [escape-str]]
+    [clojure.core.reducers :as r]
     [clojure.edn :as edn]
     [clojure.string :as string]
     [datascript.core :as d]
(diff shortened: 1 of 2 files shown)


tangjeff0:
I think this is more of a UX thing than a performance thing :)

This thread by @denik may be helpful! https://github.com/athensresearch/athens/pull/20#issuecomment-654701150 Pretty sure LogSeq implements fuzzy too

juniusfree:
That's the same article that I referred to! I should have checked if there are previous discussions already.

I updated the function using reducers. I'm not expecting this to be merged. :) I just want to learn something new. Thanks for the feedback @tangjeff0!

juniusfree:
Tested in a large db. Still slow. Closing for now.

PR #1105, opened 2021-05-04
enhance(all-pages): allow sort by titles / links / times

### LINKED ISSUE #669: enhance(all-pages): allow sort by titles / links / times

Em-AK:
close #669 

This is a minimal implementation of the feature.

### Needs improvement

1.  Show a triangle icon just before or after the column header that is actively sorted by. The icon should represent the current sorting direction: :small_red_triangle: :small_red_triangle_down: 
No triangle is visible in other columns.

2. Store `sorted-by` and `growing?` state into re-frame layer instead of local ratoms. So that the sorting preferences are maintained when the user is leaving "All pages" then coming back to it.

### DIFF
--- a/src/cljs/athens/devcards/blocks.cljs
+++ b/src/cljs/athens/devcards/blocks.cljs
@@ -1,6 +1,6 @@
 (ns athens.devcards.blocks
   (:require
-    [athens.views.core :refer [block-component]]
+    [athens.views.blocks.core :refer [block-component]]
     [devcards.core :refer-macros [defcard-rg]]))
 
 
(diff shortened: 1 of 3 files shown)


tangjeff0 (approved):
Awesome first PR! Let me know if you wanna use Paypal or Transferwise on Discord DM :)

tangjeff0:
I'm sure @shanberg might be some thoughts on All Pages toggle :)

Issue #1004, opened 2021-04-21
fix: redo incorrectly does undo on Mac


tangjeff0:
**Problem**: When user tries to redo, they actually undo, losing their latest changes!

**Screenshots/Demo**

**Athens Version**

kamalmarhubi:
This is happening on Linux as well, at least in v1.0.0-beta.76.

tangjeff0:
Not sure how to repro this, and when this happens. Are you using a different keyboard @kamalmarhubi ?

ssjoleary:
Ah, I stumbled across this!

Undo/redo gets a bit confused in a couple of scenarios:
* Either undoing or redoing in an empty block deletes that block and the block immediately above it
* Navigating to a different block and either undoing or redoing straight away has a kind of unpredictable behaviour but blocks, other than the block in focus, will start disappearing

Once the bug starts the undo/redo functionality will continue to remove things until everything's gone from the page or until the user starts typing again.

I _think_ the changes are still available in the history, it's the logic for stacking/reapplying the changes that's getting confused (rather than the undo/redo removing it from the history completely). Either way a panicked user that encounters the bug and tries to alternate between undo's and redo's will end up removing things that can't be brought back.

The undo/redo functionality looks to have been updated in March, #808 , and seems nicely contained so I'll take a stab at a fix. I think I've seen @arkRedM about on the Discord too so I'll ping them for some advice 😉

kamalmarhubi:
I wonder if this is related: https://github.com/athensresearch/athens/blob/640420f8fa81ccf88b0a3dc73b55f89949e91a87/src/cljs/athens/views/blocks/textarea_keydown.cljs#L479-L482

Issue #625, opened 2021-02-09
refactor: keybindings


tangjeff0:
Currently keybindings look at key-code and modifier keys (ctrl, shift, meta, alt).

This leads to many edge cases all over the place! 

We should instead be using a keybinding library.

- https://dmauro.github.io/Keypress/
- https://github.com/jaywcjlove/hotkeys
- https://github.com/piranha/keymage
- https://craig.is/killing/mice

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 300.0 USD (300.0 USD @ $1.0/USD)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/athensresearch/athens/625/100024991).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/athensresearch/athens/625/100024991).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
* $2,652,547.74 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 265 years, 8 months from now.
Please review their action plans below:


**1) [anukul](https://gitcoin.co/anukul) has started work.**

Replace `goog.events KeyCodes` usage with a keybinding library, refactoring any complicated usage (e.g. nesting `cond` for key combinations etc.)

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/athensresearch/athens/625/100024991).

den1k:
FYI, wrote a `case`-like macro for keybindings that could come in handy here: https://github.com/den1k/stuffs/blob/main/src/stuffs/keybind.cljc#L278-L294

julionav:
I'll work on this :muscle:

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 265 years, 6 months from now.
Please review their action plans below:


**1) [julionav](https://gitcoin.co/julionav) has started work.**

[...]

PR #1127, opened 2021-05-07
fix: redo incorrectly does undo

ssjoleary:
There's still a small bit of undesirable behaviour* but undo/redo will mostly work as expected and won't cause any data to be lost permanently.

From what I can see, the history of the text that's typed into a block is stored in electrons undo/reo history while we track the history of the blocks on the page in an `atom`. The swap from electrons history to the block history `atom` when an undo action no longer makes changes is where the bug is.

To properly fix this I think we'll need to move away from electrons undo/redo history to have better control over the undo/redo actions to make the experience smoother and more predictable. This feels like scope creep for this issue though! #1004 

I think I've seen some discussion around implementing a new text area, maybe they come with undo/redo functionality builtin or via plugins?

*constantly hitting undo on an empty block queues up each 'undo' action so you need to redo each 'undo' before you start seeing changes
*once the cursor leaves a block then its history is lost and undo/redo doesn't do anything when the cursor returns to the same block

### DIFF
--- a/src/cljs/athens/views/blocks/textarea_keydown.cljs
+++ b/src/cljs/athens/views/blocks/textarea_keydown.cljs
@@ -476,10 +476,6 @@
                                                                                       (sort-by :block/order)
                                                                                       (mapv :block/uid))]
                                                           (dispatch [:selected/add-items children]))
-      ;; When undo no longer makes changes for local textarea, do datascript undo.
-      (= key-code KeyCodes.Z) (let [{:string/keys [local previous]} @state]
-                                (when (= local previous)
-                                  (dispatch [:undo])))
 
       (= key-code KeyCodes.B) (surround-and-set "**")
 


tangjeff0:
Awesome first PR @ssjoleary!


So does this close #1004? One thing I'm not sure about is why this seems to happen only on Mac. It might have to do with `ctrl` vs `cmd`, and or with how ctrl/cmd-z creates a special JS DOM event.

I think if there was a guard making sure `shift` wasn't true within `handle-shortcuts` function, then this would also fix the bug. Having to manually check against `shift` is why I'm looking forward to a larger keybinding refactor #625.

> *once the cursor leaves a block then its history is lost and undo/redo doesn't do anything when the cursor returns to the same block

> I think I've seen some discussion around implementing a new text area, maybe they come with undo/redo functionality builtin or via plugins?

This is a good question. @shanberg do you know if those ContentEditable come with better undo/redo APIs?

ssjoleary:
Thanks @tangjeff0 👍 

Yes, this would close #1004 I think. I've reproduced the bug on Windows and can confirm that this PR fixes it there too. I don't have access to a Linux machine yet so I can't say for certain if the fix works there as well.

Oh yeah, a guard could do it too. I have a feeling that a guard might mean that we'd have to add another shortcut to handle datascript redo's within the context of a block. I might have another look at that and see what comes up.

shanberg:
> This is a good question. @shanberg do you know if those ContentEditable come with better undo/redo APIs?

Slatejs, ProseMirror, and Draftjs each provide history APIs, which I assume are each at varying levels of maturity and customizability. Most others probably do too. Seems like a common feature.

PR #1132, opened 2021-05-09
ci: cache cljstyle and add alias to lein

Em-AK:
Downloading cljstyle binary and running it takes just **11s** in GH actions.
Running the style job with `lein cljstyle` takes **30s** when the deps are already cached:
https://github.com/Em-AK/athens/actions/runs/825896007

I let you judge if this is a blocker.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,6 +68,17 @@ jobs:
            fetch-depth: 1
            submodules: 'true'
 
+       - name: Cache deps
+         uses: actions/cache@v1
+         id: cache-deps-style
+         with:
+           path: ~/.m2/repository
+           key: ${{ runner.os }}-maven-style-${{ hashFiles('project.clj') }}
+
+       - name: Fetch deps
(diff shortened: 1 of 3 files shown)


tangjeff0 (approved):
Nice job :)

tangjeff0:
@Em-AK Wait, so this is slower. Is the point just to use an alias, not performance?

Em-AK:
@tangjeff0 the main advantage I see is in dev
  * `lein cljstyle` works on any platform, in dev it is a simple alternative to manually installing the correct version of `cljstyle`
  * can coexist with the cljstyle binary in dev for those who prefer faster startup time
  * cljstyle version is set in a more obvious/visible place
  
  If perf is a problem in CI then I can revert the `script/style` to run the downloaded binary.
  But then there are two places (lein alias and scripts) where the version of cljstyle must be modified on lib upgrade.

tangjeff0:
@Em-AK perf is not a big issue for CI, at least not yet. Simplicity of installation and execution seem valuable here.

PR #1133, opened 2021-05-10
fix #576

### LINKED ISSUE #576: Pressing enter to split block with multiple links throws exception

tkembo:
Fix #576

### DIFF
--- a/src/cljs/athens/effects.cljs
+++ b/src/cljs/athens/effects.cljs
@@ -153,18 +153,13 @@
                            new-page-refs  (new-page-refs-to-tx-data (:page/refs assert-data) eid)
                            old-block-refs (old-block-refs-to-tx-data (:block/refs retract-data) eid assert-string)
                            old-page-refs  (old-page-refs-to-tx-data (:page/refs retract-data) eid assert-string with-db old-titles)
-                           ; disabling premature retractions
-                           ; tx-data        (concat []
-                           ;                        new-titles
-                           ;                        new-block-refs
-                           ;                        new-page-refs
-                           ;                        old-titles
-                           ;                        old-block-refs
-                           ;                        old-page-refs)]
                            tx-data        (concat []
                                                   new-titles
                                                   new-block-refs
-                                                  new-page-refs)]
+                                                  new-page-refs
+                                                  old-titles
+                                                  old-block-refs
+                                                  old-page-refs)]
                        tx-data)
 
                      ;; [assertion]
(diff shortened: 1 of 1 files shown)


agentydragon:
I'd recommend adding a test to check that it works now. (It's just a habit I have. That way you can both show that your PR fixes it, and improve code quality.)

tangjeff0:
Good suggestion @agentydragon. We don't have any tests here yet, which we definitely need, but it might be hard to get setup. 

For now, it would be helpful to hear your thought process @tkembo -- what you think was going wrong, and what you think your fix does.

tkembo:
Hi @agentydragon 
While I agree that adding tests are almost always better practise, in this particular instance, I don't think a test is necessary as the particular fix was a minor UI bugfix on a component that already has a good unit test (the unit test is in /test/athens/walk_test.clj).
I tried adding a case for the double link that was causing the issue (something like "[[hey]][[hey]]") but the test still passed before the bug fix and it still passed after. I guess it's because it's a bug on the UI that's only invoked when you press enter. 

@tangjeff0 we do have unit tests actually ( in /test/athens/walk_test.clj). But you'd have to manually run 'lein test' to test that.
To explain the problem simply: when you put two links on the same line, it saves them in the db. When you press enter in between them, it assumes that the second link was already saved on it's own line so it looks for it in the wrong place and can't find it. This fix just makes sure that when you press enter, it looks in the correct location in the db.

neotyk:
Probably outside of the scope, but Electron has E2E testing https://www.electronjs.org/spectron which we probably should integrate, this would allow for testing situations like this.

sid597:
@tkembo I don't think your solution works, please check out the video. The issue of throwing an exception is resolved but it creates a bigger issue which is pages not getting created.

https://www.loom.com/share/35664573bc7b4985b433ba9487fcaa09

tangjeff0 (changes requested):
https://github.com/athensresearch/athens/pull/1133#issuecomment-845856861

tkembo:
@sid597 you're right. I've now pushed another commit that reverses the first commit but fixes the first problem while avoiding the second one. 
As far as I can see, the problem is that when `parse-for-links` is called it first asserts  the link (i.e. add block refs to it thus making it a page) but then retracts all second links if they are on the same page. I've disabled that and that should fix both problems.

tangjeff0:
@tkembo you are right that is the problem, but there are cases where we do want to assert new links and retract old ones upon modifying a string. The issue happens when a user splits a block in the middle between two links.

tkembo:
@tangjeff0 I hear you. I've now taken another approach which, as far as I can see, solves the problem with no side effects:
I noticed that when we parse for links, the last link to be parsed is not immediately available when one presses enter. But reversing the order in which they are parsed seems to solve the problem.

tkembo:
hi @tangjeff0 @sid597 did you get a chance to look at the last commit?

tkembo:
hi @tangjeff0 @sid597 just following up on this again. Also, if you found another way to fix this or if this issue is no longer important it's ok too - please just let me know.

sid597:
@tkembo I will review your pull request tomorrow

tangjeff0 (changes requested):
@tkembo this isn't a full solution, it just changes the order of the data, which isn't reliable. I'm not fully sure the best solution to this, but it would require some more up front design and need to be tested very systematically before merging in because this is such a core part of Athens: the serialization between plain-text and graph db.

Feel free to start a discussion or architectural doc on this and sharing. Closing for now.

PR #1146, opened 2021-05-12
feat(app-toolbar): updated filesystem/sync icons

juniusfree:
@shanberg I updated the [sync icons](https://www.craft.do/s/MAGVSD9LgwXSnF/b/3BF560E4-65B3-40F9-9F30-289176B28928/Better_Accessibility). Let me know your thoughts. Thanks.

Synced
image
image

Syncing
image
image

Error
image
image

Database modal
image
image

App toolbar right
image
image


shanberg:
This is a huge improvement in readability.

I don't think there's a clear answer to what material-ui icon best represents a knowledge graph database, so I won’t dwell too much on that change, but we should use the same icon wherever the same concept is being represented.

Relatedly, this icon/status indicator can replace the filesystem dialog button icon, and the database icon in the filesystem dialog itself.

juniusfree:
I'll try to see if there is a more relevant icon that'll represent the graph db and use it in the filesystem.

juniusfree:
@shanberg I can't find other icons for db so I just use Storage icon for the filesystem and just combine it with the sync status icon.

tangjeff0:
Is this ready for review? @shanberg can you review too?

src/cljs/athens/views/app_toolbar.cljs
```diff
@@ -179,9 +165,28 @@
                      :title "Open Settings"
                      :active   (= @route-name :settings)}
              [:> Settings]]
+
             [button {:on-click #(dispatch [:modal/toggle])
                      :title "Choose Database"}
-             [:> LibraryBooks]]
+             [:div {:style {:display "flex"}}
```
shanberg, line 171: This sequence of added divs and styles could possibly be refactored into a single style applied to the button itself, and using `::stylefy/manual [[:div {:display "flex"}...]] to style its children.

src/cljs/athens/views/app_toolbar.cljs
```diff
@@ -179,9 +165,28 @@
...
+              [:> Storage {:style {:align-self "center"}}]
+              [:div {:style {:margin-left "-10px"
+                             :align-self "flex-end"}}
+               (cond
+                 (= @socket-status :closed)
+                 [:> Error (merge (use-style sync-icon-style)
```
shanberg, line 177: A possibly cleaner way to change the colors according to status would be to add a class on the button, `.closed`, `.running` etc., and change the color in the style object.

shanberg (approved):
Beautiful. Really ties the toolbar together.

tangjeff0 (approved):

PR #1166, opened 2021-05-15
fix(keybindings): Using keyboard shortcut for italics incorrectly places cursor after, instead of inside asterisks #1161

### LINKED ISSUE #1161: Using keyboard shortcut for italics incorrectly places cursor after, instead of inside asterisks
When hitting CMD-I on a Mac, Athens pastes in two asterisks.  That's good!
Unfortunately -- unlike with CMD-B for Bold -- the cursor is placed _after_ the asterisks, requiring the user to backspace.

DESIRED BEHAVIOR:
After hitting CMD-I (or CTRL-I on Windows), Athens should paste in two asterisks (as it currently does) and then place the cursor in between the two asterisks (which it does not do)

**Athens Version**
Beta 80

sawhney17:
Fixing the issue of creating italics and having the cursor end up outside the created italics using the keyboard shortcut. 

My first _proper_ clojure commit, my previous ones were just simple hyperlink fixes here and there.

I think we can close  #1161


tangjeff0 (changes requested):
Awesome first PR! I would probably try to extract `surround-and-set` into a top level function, and then parameterize the number so that we don't rewrite `surround-and-set1` if the only difference is `2` vs `1` :) 


```
(defn surround-and-set
  "Default to n=2 because it's more common."
  ([e state surround-text]
   (surround-and-set e state surround-text 2))
  ([e state surround-text n]
   (let [{:keys [head tail selection start end target]} (destruct-key-down e)
         selection?       (not= start end)]
     (.preventDefault e)
     (.stopPropagation e)
     (let [selection (surround selection surround-text)
           new-str   (str head selection tail)]
       ;; https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand
       ;; textarea setval will lose ability to undo/redo

       ;; other note: execCommand is probably the simpler way
       ;; at least until a new standard comes around

       ;; be wary before updating electron - as chromium might drop support for execCommand
       ;; electron 11 - uses chromium < 90(latest) which supports execCommand
       (swap! state assoc :string/local new-str)
       (.. js/document (execCommand "insertText" false selection))
       (if selection?
         (do (setStart target (+ n start))
             (setEnd target (+ n end)))
         (set-cursor-position target (+ start n)))))))


;; TODO: put text caret in correct position
(defn handle-shortcuts
  [e uid state]
  (let [{:keys [key-code head tail selection target value shift]} (destruct-key-down e)]
    (cond
      (and (= key-code KeyCodes.A) (= selection value)) (let [closest-node-page  (.. target (closest ".node-page"))
                                                              closest-block-page (.. target (closest ".block-page"))
                                                              closest            (or closest-node-page closest-block-page)
                                                              block              (db/get-block [:block/uid (.. closest -dataset -uid)])
                                                              children           (->> (:block/children block)
                                                                                      (sort-by :block/order)
                                                                                      (mapv :block/uid))]
                                                          (dispatch [:selected/add-items children]))
      ;; When undo no longer makes changes for local textarea, do datascript undo.
      (= key-code KeyCodes.Z) (let [{:string/keys [local previous]} @state]
                                (when (= local previous)
                                  (if shift
                                    (dispatch [:redo])
                                    (dispatch [:undo]))))

      (= key-code KeyCodes.B) (surround-and-set e state "**")

      (= key-code KeyCodes.I) (surround-and-set e state "*" 1)

      (= key-code KeyCodes.Y) (surround-and-set e state "~~")

      (= key-code KeyCodes.U) (surround-and-set e state "--")

      (= key-code KeyCodes.H) (surround-and-set e state "^^")

      ;; if caret within [[brackets]] or #[[brackets]], navigate to that page
      ;; if caret on a #hashtag, navigate to that page
      ;; if caret within ((uid)), navigate to that uid
      ;; otherwise zoom into current block

      (= key-code KeyCodes.O) (let [[uid _]   (db/uid-and-embed-id uid)
                                    link      (str (replace-first head #"(?s)(.*)\[\[" "")
                                                   (replace-first tail #"(?s)\]\](.*)" ""))
                                    hashtag   (str (replace-first head #"(?s).*#" "")
                                                   (replace-first tail #"(?s)\s(.*)" ""))
                                    block-ref (str (replace-first head #"(?s)(.*)\(\(" "")
                                                   (replace-first tail #"(?s)\)\)(.*)" ""))]

                                ;; save block before navigating away
                                (db/transact-state-for-uid uid state)

                                (cond
                                  (and (re-find #"(?s)\[\[" head)
                                       (re-find #"(?s)\]\]" tail)
                                       (nil? (re-find #"(?s)\[" link))
                                       (nil? (re-find #"(?s)\]" link)))
                                  (let [eid (db/e-by-av :node/title link)
                                        uid (db/v-by-ea eid :block/uid)]
                                    (if eid
                                      (router/navigate-uid uid e)
                                      (let [new-uid (athens.util/gen-block-uid)]
                                        (.blur target)
                                        (dispatch [:page/create link new-uid])
                                        (js/setTimeout #(router/navigate-uid new-uid e) 50))))

                                  ;; same logic as link
                                  (and (re-find #"(?s)#" head)
                                       (re-find #"(?s)\s" tail))
                                  (let [eid (db/e-by-av :node/title hashtag)
                                        uid (db/v-by-ea eid :block/uid)]
                                    (if eid
                                      (router/navigate-uid uid e)
                                      (let [new-uid (athens.util/gen-block-uid)]
                                        (.blur target)
                                        (dispatch [:page/create link new-uid])
                                        (js/setTimeout #(router/navigate-uid new-uid e) 50))))

                                  (and (re-find #"(?s)\(\(" head)
                                       (re-find #"(?s)\)\)" tail)
                                       (nil? (re-find #"(?s)\(" block-ref))
                                       (nil? (re-find #"(?s)\)" block-ref))
                                       (db/e-by-av :block/uid block-ref))
                                  (router/navigate-uid block-ref e)

                                  :else (router/navigate-uid uid e))))))
```

PR #1167, opened 2021-05-16
refactor: remove posh, use datascript

juniusfree:
TODO
- [x] remove all posh-related codes and update with equivalent datascript functions.
- [ ] make the components reactive
  - [x] page component
- [ ] update the datascript queries with faster queries
- [ ] more testing

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -227,7 +227,7 @@
               (siblings ?uid ?sib)
               [?sib :block/uid ?sib-uid]
               [?sib :block/order ?sib-o]]
-            @dsdb rules uid) ;;TODO: determine the purpose of rules here
+            @dsdb rules uid)
        (sort-by second)
        last
        first
@@ -272,11 +272,12 @@
 (def roam-node-document-pull-vector
   '[:node/title :block/uid :block/string :block/open :block/order {:block/children ...}])
 
-;; TODO:
+
 (defn get-block-document
   [id]
-  (->> (d/pull @dsdb block-document-pull-vector id)
-       sort-block-children))
+  (ratom/make-reaction
+    #(->> (d/pull @dsdb block-document-pull-vector id)
+          sort-block-children)))
 
 
 (defn get-node-document
@@ -289,7 +290,7 @@
      #(->> (d/entity db id)
            sort-block-children-datom))))
 
-;; TODO: to test
+
 (defn get-roam-node-document
   [id db]
   (->> (d/pull db roam-node-document-pull-vector id)
@@ -349,7 +350,7 @@
                          [(dec ?target-o) ?prev-sib-order]
                          [?sib :block/order ?prev-sib-order]
                          [?sib :block/uid ?uid]]
-                       @dsdb rules uid) ; TODO: determine the purpose of rules
+                       @dsdb rules uid)
         older-sib (get-block [:block/uid sib-uid])]
     older-sib))
 
@@ -595,7 +596,7 @@
   (let [parents (reduce-kv (fn [m _ v] (assoc m v @(get-parents-recursively v)))
                            {}
                            ref-ids)
-        blocks (map (fn [id] (get-block-document id)) ref-ids)]
+        blocks (map (fn [id] @(get-block-document id)) ref-ids)]
     (mapv
       (fn [block]
         (merge block {:block/parents (get parents (:db/id block))}))
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -6,7 +6,8 @@
     [clojure.string :as string]
     [datascript.core :as d]
     [re-frame.core :refer [dispatch]]
-    [reagent.core :as r]))
+    [reagent.core :as r]
+    [reagent.ratom :as ratom]))
 
 
 ;; -- Example Roam DBs ---------------------------------------------------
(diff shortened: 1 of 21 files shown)


juniusfree:
@tangjeff0
1. It seems like wrapping the datascript db in a ratom is enough to make it reactive. But I'm unsure if I missed something there.
2. I realized that putting all updates to this one PR might make it hard to review and test. I'm planning to have a separate PR for related queries and optimizations (e.g. by namespace). What do you think?

juniusfree:
@tangjeff0 Do you think this can be reviewed already? I'll update the other queries separately because they're either complex or related to other queries.

juniusfree:
@tangjeff0 Search is almost instantaneous now. From 200-500ms to less than 20ms on my PC.

Fix #756 https://github.com/athensresearch/athens/pull/1167/commits/10d65bb44d7f90a9ab8e631c13eca41be8b89f51

juniusfree:
I created a separate PR for the search. https://github.com/athensresearch/athens/pull/1191

It might be better to refactor by feature.

src/cljs/athens/db.cljs
```diff
@@ -266,13 +264,13 @@
 
 (defn get-block-document
   [id]
-  (->> @(pull dsdb block-document-pull-vector id)
+  (->> (d/pull @dsdb block-document-pull-vector id)
```
neotyk, line 267: We want this to be reactive.
Please use reagent.ratom/reaction.

neotyk (commented):
Major part of our inefficiency with data script comes from deref-ing all the time in helper functions.

This removes reagent's ability to rerender only thighs that have changed, because we aren't returning reaction-like values.
This causes unnecessary stress on the DB.

If we'd wrap helper fns into reactions we can then use form-3 components and rip benefits of updating only things that have changed automatically.

juniusfree:
To follow on @neotyk comment above and to ensure that I understand the issue, I checked #784. Let me know if the following is correct or if I missed something.

It seems like the problem with #784, is that the functions in db.cljs were not refactored to return a `reaction`.

eg.
refactored `pull`
https://github.com/pithyless/athens/blob/c2f33675c8663a2e76278e46f96face39513b0ff/src/cljs/athens/dbrx.cljs#L15-L18

using `pull` but deref'ed immediately instead of returning a `reaction`
https://github.com/pithyless/athens/blob/c2f33675c8663a2e76278e46f96face39513b0ff/src/cljs/athens/db.cljs#L263-L266

`node-page` component is receiving the deref'ed value instead of `reaction`
https://github.com/pithyless/athens/blob/c2f33675c8663a2e76278e46f96face39513b0ff/src/cljs/athens/views/node_page.cljs#L584

To fix this, do something like this.

e.g. 
```
;; in db.clj
(defn get-node-document
          [id]
          (ratom/make-reactions
             (->> (d/pull @dsdb node-document-pull-vector id)
               sort-block-children)))
```
```
;; in component

(defn page
  [ident]
  (let [node (db/get-node-document ident)]
    [:div @node]))
```

juniusfree:
- Tested the make-reaction in node page component and sub-components
- Refactored the node page sub-components to prevent unnecessary re-render

I think the refactoring of the other components should be done in a separate PR

juniusfree:
- made `get-node-page-document` 4x faster in my PC

tangjeff0 (changes requested):
Can you write-up some of your changes for this PR @juniusfree ?

How you are using `r/atom`, `make-reaction` instead of posh, how linked/unlinked refs are made now, etc.

juniusfree:
## Replace `posh` with `reagent`
- Previous Work
    - https://github.com/athensresearch/athens/pull/784
    - https://github.com/athensresearch/athens/pull/665
- Problem
    - Performance issues with `posh`
- Solution
    - Replace `posh` with `datascript`
    - `reagent.core/atom` wraps the datascript db to make it reactive from queries. This is like the `app-db` in `re-frame`
    - Follow the dominoes 4-5-6 data flow pattern of `re-frame` to make it organized. https://day8.github.io/re-frame/flow-mechanics/
        - For query -> use the `clojure.core/ratom` functions like `make-reaction`, `reaction`, `track`. All queries, if possible, should be in the `db.clj` file. To make the queries faster, use `d/entity` or `d/datoms`. `d/pull`, `d/q` might still be used for complex queries.
        - View -> To make the component reactive, it'll deref the value from the `make-reaction`. This is equivalent to the `subscribe` fn in `re-frame`
- TODO
    - Update the other datascript db queries

## Node Page
- This might be out of scope for this PR. I just want to check the effect of removing `posh`. I can remove this easily.
- Problem
  - The node page component instance is reloaded for state updates in linked or unlinked ref.
- Solution
  - Manage the state of linked and unlinked ref separately from the main component
    - showing/hiding the linked/unlinked ref shouldn't affect the main node page component
    - newly created linked or unlinked ref shouldn't update the main component because they will directly deref the queries on their own instead of getting them from the main node page component

@tangjeff0

juniusfree:
1. Reverted the changes to `node-page` except for those that are directly affected by the removal of `posh`
2. `posh` is completely removed
3. Most db queries are now using `make-reactions`

juniusfree:
Something feels wrong with this. I'll work on other issues first. I might get back to this in the future.

tangjeff0:
What feels wrong @juniusfree

juniusfree:
@tangjeff0
1. Lack of exploration for other potential implementation like the following:
    1. wrap the calling function in a `make-reaction`, `track`, etc. functions instead of updating the functions in `db`.
    2. create a library like what have been done on the two previous PRs
    3. continue to use `posh`, optimize the existing code.
2. And also, I realized that I need to improve my understanding of software design to solve broader problems like this.

juniusfree:
Working again with this. I'll reopen once I got more updates.

PR #1168, opened 2021-05-16
fix(parser): remove support for underscores so URLs can use

### LINKED ISSUE #1058: Parser: fails to parse some page links
**Problem**

The following blocks:

```
{{[[TODO]]}} [[Buy]] at least 2 [[Cha hai]]s and 2 [[Gaiwan]]s from [[Yunnan Sourcing]]
```

```
{{[[TODO]]}} learn how to do unsupervised pre-training with [[Restricted Boltzmann machine]]s or what not
```

do not render with correct links in my Athens database:

* 
* 

I suspect this might be because there's an alphabetic character after the links.

**Athens Version**

1.0.0-beta.77 454682547046109d0cea3512d17d4fb3095397ec

cc @neotyk

### LINKED ISSUE #1059: Parser: cannot combine component with fenced code block
**Problem**

This block:

```
{{[[TODO]]}} 
\`\`\`
hello
\`\`\`
```

Renders as:

image

I'd expect it to render a pre-formatted code block after the TODO component, instead of parsing the triple-backticks as inline fenced code block containing two backticks on each side.

**Athens Version**

1.0.0-beta.77 454682547046109d0cea3512d17d4fb3095397ec

cc @neotyk

neotyk:
Fixes #1058 & fixes #1059 

Removes support for `_` as emphasis boundary.


agentydragon:
I would recommend adding tests against the bugs I reported. That way when the parser is updated in the future, we can be sure it'll stay OK.

agentydragon:
Sent PR: https://github.com/athensresearch/athens/pull/1174

Issue #1117, opened 2021-05-06
Toolbar Presence


jsmorabito:
# API

- `username` - required
- `:block/uid` - optional
- `:page/uid` - optional, not in datascript (yet), just derived from `:block/uid`
- Constraints: 10 users max

v2

- click on avatar to **follow** that person
- show more context on hover

# Screenshots

![image](https://user-images.githubusercontent.com/8952138/121932068-474bb900-ccf9-11eb-8ade-480ab3ecd3b6.png)
![image](https://user-images.githubusercontent.com/8952138/121932123-592d5c00-ccf9-11eb-9e64-9df4f05ef276.png)

tangjeff0:
Defer to #1119

tangjeff0:
Closing because we already have PR draft #1361

tangjeff0:
Closing because we already have PR draft #1361

PR #1170, opened 2021-05-16
feat(rtc): v1 self-hosted real-time collaboration

neotyk:
See [Project Overview](https://github.com/athensresearch/athens/projects/16)

Merging #1117 into this, see https://github.com/neotyk/athens/pull/2


src/cljc/athens/common_events.cljc
```diff
@@ -0,0 +1,17 @@
+(ns athens.common-events
+  "Event as Verbs executed on Knowledge Graph"
+  (:require [clojure.string :as string]))
+
+(defn paste-verbatim->tx [uid text start value]
```
agentydragon, line 5: Consider adding a test.

neotyk:
Proposed event flow

> - We 1st receive `re-frame` event (that already is a verb)
> - if remote
>     - create `event-id`
>     - read `last-seen-tx`
>     - start waiting for `event-id` ack
>         - ui: event queued
>     - send via #[[WebSocket]] event
>         - with `event-id, last-seen-tx, event-args`
>     - on server: receive event
>         - check `last-seen-tx`
>             - reject event when potentially conflicting
>                 - let client know what is the `last-global-tx`
>             - working from current head
>                 - resolve event in db context into txs (**common-events**)
>                 - `transact!`
>                 - send via #[[WebSocket]] event ack
>                 - broadcast tx log to connected clients
>     - on client
>         - when ack received
>             - remove `event-id` from awaiting ack
>             - ui: event completed
>         - when rejection received
>             - ui: notify user of unsuccessful event execution
>             - update `last-global-tx`
>             - ui: lagging db
>                 - subscription `::subs/remote-ahead?`
>                     - `last-seen-tx < last-global-tx`
> - else
>     - resolve event in local db context into txs (**common-events**)
>     - `transact!`
> - on client: when received `tx-log` broadcast
>     - apply received `tx-log` to local `DataScript` db
>     - ui: automatic update via reactive
>

sid597:
I have a few questions regarding this :
1. How does the server know some device has disconnected? 
2. Can a client have stale data because of a network partition between the server and that client? I think the answer to this question depends on how question 1 is handled.

tangjeff0:
> How does the server know some device has disconnected?

Websocket servers typically provide a very easy API for detecting `connect` and `disconnect` events. @sid597

sid597:
Thank you @tangjeff0

I will have to read more regarding the Websocket approach, questions I will have in mind while reading it are : 
- Is network partition the same as `disconnect`? (assuming network will heal )
- Difference between network delay and disconnect, how is it determined what case is what.
- In case of network delays :
    - How is the situation of transactions arriving out of order is addresed?
    - Does the library resend a transaction, if so in what case and how would same transaction arriving on the client-side handled?

tangjeff0:
> Is network partition the same as disconnect? (assuming network will heal )

Yes. Heal = reconnect

> Difference between network delay and disconnect, how is it determined what case is what.

If there is delay but no disconnect, should be handled by TCP or networking layer.

[Offline is a special case of lag compensation where lag is infinite. Offline mode should look just like a normal app without real-time updates. ](https://tonsky.me/blog/the-web-after-tomorrow/#:~:text=offline%20is%20a%20special%20case%20of%20lag%20compensation%20where%20lag%20is%20infinite.%20offline%20mode%20should%20look%20just%20like%20a%20normal%20app%20without%20real-time%20updates.%20)

> How is the situation of transactions arriving out of order is addresed?

Transactions should arrive in order, should be handled by TCP layer or websocket library.

sid597:
Now that all the baseless assumptions and worries I had are resolved, how can I help in this implementation @neotyk @tangjeff0

src/cljs/athens/events.cljs
```diff
@@ -1698,28 +1699,34 @@
...
-                                      text
-                                      (subs value start)))
-         tx-data [{:db/id        [:block/uid uid]
-                   :block/string new-string}]]
-     {:dispatch [:transact tx-data]})))
+         local?                (not (:db/remote-graph-conf db))]
+     (if local?
+       {:dispatch [:transact (common-events/paste-verbatim->tx uid text start value)]}
+       ;; TODO event handler for `:remote/paste-verbatim`
+       {:dispatch [:remote/paste-verbatim uid text start value]}))))
```
tangjeff0, line 1729: So every datascript+re-frame transaction needs to check if `local?`?

Is this a good use case for [interceptors](https://day8.github.io/re-frame/Interceptors/)? Every `ds+rf` event can have a `local` interceptor that decides to write transaction on client datascript or through backend datahike.

src/clj/athens/self_hosted/web/presence.clj
```diff
@@ -0,0 +1,79 @@
...
+
+(defn presence-handler
+  [req]
+  (http/as-channel req
+                   {:on-open    #'open-handler
+                    :on-receive #'receive-handler
+                    :on-close   #'close-handler}))
+
+(compojure/defroutes presence-routes
+  (compojure/GET "/ws-presence" [] presence-handler))
```
tangjeff0, line 79: Please add `(compojure/GET "/ws-presence" [] <h1>hello world</h1>)` for easy check that server is up.

tangjeff0 (commented):
Awesome job so far @neotyk. See comments about interceptor

dev/clj/user.clj
```diff
@@ -0,0 +1,25 @@
+(ns user
+  (:require [athens.self-hosted.core      :as app]
+            [clojure.tools.namespace.repl :as repl]
+            [com.stuartsierra.component   :as component]))
```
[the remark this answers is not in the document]
neotyk: And component works on both clj & cljs.

src/cljs/athens/events.cljs
neotyk: I need to check, haven't had the need for using interceptors in past.

neotyk: Ok, checked.
Using interceptors to "route" events would be possible (through self modifying property of interceptors)
but it would be confusing, because we'd register event-handler `:paster-verbatim` with this magic interceptor,
and magic interceptor would re-route event to other handler, this would result in very unintuitive code.

We could inject cofx to test if we're in local vs self-hosted mode.

tangjeff0: Cofx is probably the right move then! Both Cofx and Interceptors are good at reusing some code for events, but Interceptors are a little too magical.

src/cljs/athens/events.cljs
```diff
@@ -1818,3 +1824,93 @@
...
+       :fx [[:dispatch-n events]]})))
+
+
+(reg-event-db
+  :remote/reject-event
+  (fn [db [_ {:keys [event-id reason data] :as rejection-event}]]
+    (js/console.log "reject event" (pr-str rejection-event))
+    (let [awaited-event  (->> (:remote/awaited-events db)
+                              (filter #(= event-id (:event/id %)))
+                              first)
```
agentydragon, line 1879: duplicated with lines 1854-1856, consider deduplicating

(e.g.: `(first-awaited-event db)`)

agentydragon: also lines 1893..1895

src/cljs/athens/self_hosted/client.cljs
```diff
@@ -0,0 +1,397 @@
...
+    (js/console.log "User online:" username)))
+
+
+(defn- server-event-handler
+  [{:event/keys [id last-tx type args] :as packet}]
+  (js/console.log "<-" id ", last-tx:" last-tx ", type:" type)
+  (js/console.debug "WSClient: server event:" (pr-str packet))
+  (if (schema/valid-server-event? packet)
+
+    (condp = type
```
agentydragon, line 241: `case`? https://clojuredocs.org/clojure.core/case

test/athens/events/awaited_events_test.cljs
```diff
@@ -0,0 +1,146 @@
...
+    (let [awaited-events (rf/subscribe [:remote/awaited-events])]
+      (is (not (nil? @awaited-events)))
+      (is (empty? @awaited-events)))))
+
+
+(deftest awaiting-one-event
+  (rf-test/run-test-async
+    (fixture/test-fixtures)
+    (rf/dispatch-sync [:boot/web])
+    (let [awaited-events (rf/subscribe [:remote/awaited-events])]
```
agentydragon, line 29: might be useful to share this setup boilerplate to make each test a bit shorter.

agentydragon: also might be useful to factor into one function the test of "is this an empty collection" (e.g., like on lines 32,33), or the test of "is a collection containing this single element"

src/clj/athens/self_hosted/components/web.clj
```diff
@@ -0,0 +1,118 @@
...
+
+
+(defn make-handler
+  [datahike]
+  (compojure/routes health-check-route
+                    (make-ws-route datahike)))
+
+
+(defrecord WebServer
+  [config httpkit datahike]
```
tangjeff0, line 91: Why is this ordered `[config httpkit datahike]`? I would expect this to be ordered `[config datahike httpkit]`? because config and datahike are dependencies and httpkit is added in constructor.

src/clj/athens/self_hosted/components/web.clj
```diff
@@ -0,0 +1,118 @@
...
+                                            result)))))))))
+
+
+(defn- make-websocket-handler
+  [datahike]
+  (fn websocket-handler
+    [request]
+    (http/as-channel request
+                     {:on-open    #'open-handler
+                      :on-close   #'close-handler
```
tangjeff0, line 69: Why pass symbol instead of function?

src/cljs/athens/events.cljs
```diff
@@ -1,48 +1,79 @@
...
      :local-storage/set! ["db/filepath" filepath]}))
 
 
-(reg-event-db
+(rf/reg-event-fx
+  :remote/connect!
+  (fn [{:keys [db]} [_ connection-config]]
+    (js/console.log ":remote/connect!" (pr-str connection-config))
+    {:db                     (-> db
+                                 (dissoc :db/filepath)
```
tangjeff0, line 47: TODO: unlisten `:db/filepath` node.js file listener

src/cljs/athens/effects.cljs
```diff
@@ -369,21 +374,64 @@
...
+(rf/reg-fx
   :right-sidebar/scroll-top
   (fn []
     (let [right-sidebar (js/document.querySelector ".right-sidebar-content")]
       (when right-sidebar
         (set! (.. right-sidebar -scrollTop) 0)))))
+
+
+;; TODO: temporary, and limits to one client opened.
+(def self-hosted-client (atom nil))
```
tangjeff0, line 398: I know this is temporary, but is this the same as `ws-connection`?

https://github.com/athensresearch/athens/blob/63aa3515ad71012f32ea0ff1776093e7c6cc1a62/src/cljs/athens/self_hosted/client.cljs#L18

src/cljs/athens/self_hosted/client.cljs
```diff
@@ -0,0 +1,449 @@
...
+    (.removeEventListener "open" open-handler)))
+
+
+(defn- close-handler
+  [event]
+  (js/console.log "WSClient Disconnected:" event)
+  (let [connection (.-target event)
+        url        (.-url connection)]
+    (remove-listeners! connection)
+    (delayed-reconnect! url)))
```
tangjeff0, line 340: Why reconnect on close?

src/cljs/athens/effects.cljs
```diff
@@ -1,23 +1,28 @@
...
     [day8.re-frame.async-flow-fx]
-    [goog.dom.selection :refer [setCursorPosition]]
-    [posh.reagent :as p :refer [transact!]]
-    [re-frame.core :refer [dispatch reg-fx subscribe]]
-    [stylefy.core :as stylefy]))
+    [goog.dom.selection          :refer [setCursorPosition]]
+    [malli.core                  :as m]
+    [malli.error                 :as me]
+    [posh.reagent                :as p :refer [transact!]]
+    [re-frame.core               :as rf]
```
tangjeff0, line 24: I think we should refactor `[re-frame.core :refer [dispatch reg-fx subscribe]]` to `[re-frame.core :as rf]` in a separate PR here and in `events.cljs`.

src/clj/athens/self_hosted/components/web.clj
tangjeff0: `#'open-handler` - a ref, can develop and live reload this value. run-time. refer to current value

`open-handler` - refer to compile-time function. always refer to old one

src/clj/athens/self_hosted/components/web.clj
tangjeff0: `defrecord` semantics destructure the map based on key name.

tangjeff0: Order doesn't matter, but it does make sense to order meta-data first (I'm right :))

src/cljs/athens/effects.cljs
tangjeff0: `component` and `re-frame` both frameworks. This is first instance of using `component` on front-end. How to use together if at all?

src/cljs/athens/self_hosted/client.cljs
tangjeff0: `on-close` means we didn't try to stop the connection, it was unintentional like a network issue or closing laptop, so we should try to reconnect.

There is a specific close event that should only be called intentionally.

tangjeff0:
Some notes and TODOs from https://github.com/neotyk/athens/pull/2:

- [ ] generate colors and names when a user joins
- [ ] `:presence/all-online` is sent a user joins, but should probably also send `:presence/all-editing`, otherwise need to wait until someone clicks for the first presence uid to be shared
- [ ] delete old presence code
- [ ] convert `:presence/users` from vector into map


Later
- [ ] convert namespaced keys into nested map, e.g. `:presence/users` into `:presence {:users ...}`

src/cljs/athens/effects.cljs
```diff
@@ -270,6 +271,10 @@
 (rf/reg-fx
   :transact!
   (fn [tx-data]
+    ;; 🎶 Sia "Cheap Thrills"
```
sid597, line 274: 😆

src/cljs/athens/events/remote.cljs
```diff
@@ -0,0 +1,458 @@
...
+
+(rf/reg-event-fx
+  :remote/connect!
+  (fn [{:keys [db]} [_ connection-config]]
+    (js/console.log ":remote/connect!" (pr-str connection-config))
+    {:db                     (-> db
+                                 (dissoc :db/filepath)
+                                 (assoc :db/remote connection-config))
+     :remote/client-connect! connection-config
+     :local-storage/set!     ["db/remote" connection-config]
```
sid597, line 22: @neotyk Connection config is a map so why are we not serializing it while saving in local storage?

neotyk: Because didn't implement config, yet

PR #1171, opened 2021-05-16
fix(roam-import): fix roam-date regex to match ordinal numbers in roam dates

### LINKED ISSUE #1135: Error importing Roam Research's EDN database file
Error message：

```
message=Uncaught #error {:message "Conflicting upsert: [:node/title \"December 20, 2020\"] resolves to 1449, but entity already has :db/id 1273", :data {:error :transact/upsert, :assertion [1449 :node/title "December 20, 2020"], :conflict {:db/id 1273}}}
source=file:///Applications/Athens.app/Contents/Resources/app.asar/resources/public/js/compiled/renderer.js
lineno=4802
colno=204
error=#error {:message "Conflicting upsert: [:node/title \"December 20, 2020\"] resolves to 1449, but entity already has :db/id 1273", :data {:error :transact/upsert, :assertion [1449 :node/title "December 20, 2020"], :conflict {:db/id 1273}}}
```

https://user-images.githubusercontent.com/72077265/117704125-127dab00-b1fd-11eb-9688-29f99ea8f20e.mp4

julionav:
This fixes the import issues from #1135

### Problem analysis

- Seems to be caused by a conflict on the page November 23, 2020.
- The problem is caused by a conflict on the transformation of a daily page in [[Roam]] in [[Athens]]. [[Roam]] uses  the ordinal number notation, and [[Athens]] the cardinal notation. 
    - However, In my [[Roam]] database, there are 2 pages that are transformed to [[November 23, 2020]]
        - [[November 23rd, 2020]]
        - And a misspelled [[November 23nd, 2020]]
    - Even though the second date is misspelled, Athens transforms it to [[November 23, 2020]] through a [[Regex]].
        - `#"(?<=\d{1,2})(st|nd|rd|th),(?=\s\d{4})"`
            - This matches `23nd` and `23rd` and treats them the same.
    - The solution proposed solution is to use a more strict conversion between ordinal notation to cardinal notation.

### Solution

`roam-date` regex pattern was rewritten to only match correct ordinal numbers in dates, so the conflict won't happen.

Before, `November 23nd, 2020` (not a correct date, it should be `November 23rd, 2020`) was matched. Now only `November 23rd, 2020` will match

closes partially #1135


tangjeff0:
This is a partial solution to #1135. The actual problem is that datascript is attempting to add a page that already exists. In your case, that is two Roam date pages that are converted to the same Athens date page because of a not-very-strict regex.

However, I think this common case for #1135 is when a user imports a page that already exists in Athens. For instance, if you open both Athens and Roam to today, 5/16/21, it will attempt to create a "May 16, 2021" page in Athens, but it already exists.

julionav:
@tangjeff0 I think there is some kind of code that does that particular merge since the app asks you if you want to merge pages that exists both in Athens and Roam (like [[TODO] and [[DONE]]

This particular problem seems to happen when the Roam DB is adapted to Athens expected format (in `update-roam-db-dates`)

We can check with @mkyos to see if this fixes the issue (which is also date page related)

PR #1176, opened 2021-05-17
fix(keybindings): place caret correctly after ctrl-i italics

### LINKED ISSUE #1161: Using keyboard shortcut for italics incorrectly places cursor after, instead of inside asterisks

sawhney17:
Fixed the changes requested on the original pull request. Close #1166 and close #1161.


tangjeff0 (approved):

tangjeff0:
Git likes it when the top-level commit message is <= 80 chars long.

PR #1191, opened 2021-05-18
perf(search): faster search for (()), [[]] and ctrl-k

### LINKED ISSUE #756: perf(search): faster search for (()) and ctrl-k

juniusfree:
Making search instantaneous. From ~300ms to ~50ms (or less) on my PC . Tested in relatively large db https://roamresearch.com/#/app/KJV/page/lFeHFy8qE

fix #756

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -417,11 +417,11 @@
            case-insensitive-query (re-case-insensitive query)]
        (sequence
          (comp
-           (map #(d/entity @dsdb (:e %)))
            (filter (every-pred
-                     #(not= exact-match (:node/title %))
-                     #(re-find case-insensitive-query (:node/title %))))
-           (take n))
+                     #(not= exact-match (:v %))
+                     #(re-find case-insensitive-query (:v %))))
+           (take n)
+           (map #(d/entity @dsdb (:e %))))
          (d/datoms @dsdb :aevt :node/title))))))
 
 
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -1,12 +1,12 @@
 (ns athens.db
   (:require
-    [athens.patterns :as patterns]
-    [athens.util :refer [escape-str]]
-    [clojure.edn :as edn]
-    [clojure.string :as string]
-    [datascript.core :as d]
-    [posh.reagent :refer [posh! pull q]]
-    [re-frame.core :refer [dispatch]]))
+   [athens.patterns :as patterns]
+   [athens.util :refer [escape-str]]
+   [clojure.edn :as edn]
+   [clojure.string :as string]
+   [datascript.core :as d]
+   [posh.reagent :refer [posh! pull q]]
+   [re-frame.core :refer [dispatch]]))
 
 
 ;; -- Example Roam DBs ---------------------------------------------------
@@ -416,14 +416,14 @@
      (let [exact-match            (when exclude-exact-match? query)
            case-insensitive-query (re-case-insensitive query)]
        (sequence
-         (comp
-           (filter (every-pred
-                     #(not= exact-match (:v %))
-                     #(re-find case-insensitive-query (:v %))))
-           (take n)
-           (map #(d/entity @dsdb (:e %))))
-         (d/datoms @dsdb :aevt :node/title))))))
-
+        (comp
+         (filter (every-pred
+                  #(re-find case-insensitive-query (:v %))
+                  #(not= exact-match (:v %))))
+         (take n)
+         (map #(d/entity @dsdb (:e %))))
+        (d/datoms @dsdb :aevt :node/title))))))
+        
 
 (defn get-root-parent-node
   [block]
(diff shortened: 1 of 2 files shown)


sawhney17:
Can't wait to get my hands on this!

Limezy:
So cool !

sawhney17:
I just tried the dev build and it's fast! No discernible lag at all on my 3MB database even on the electron build.

Limezy:
@juniusfree I sent you my 32500 blocks / 100 pages / 13MB database for your performance tests !

src/cljs/athens/db.cljs
```diff
@@ -417,17 +413,16 @@
...
+     (let [ignore-dup             (when ignore-dup query)
+           case-insensitive-query (re-case-insensitive query)]
+       (sequence
+         (comp
+           (filter #(re-find case-insensitive-query (:v %)))
+           (filter #(not= (:v %) ignore-dup))
+           (take n)
+           (map #(:e %))
+           (map #(d/entity @dsdb %)))
+         (d/datoms @dsdb :aevt :node/title))))))
```
arkRedM, line 425: Use ->> to rewrite it in more readable form
(->> (d/datoms) .. 
         (map (comp :e #(d/entity ..))
         (take n)
         (filter (apply every-pred(https://clojuredocs.org/clojure.core/every-pred) filter1 filter2))
         seq)

src/cljs/athens/db.cljs
```diff
@@ -417,17 +413,16 @@
...
-                               :where
-                               [?node :node/title ?title]
-                               [(re-find ?query-pattern ?title)]
-                               [(not= ?title ?query)]]                ;; ignore exact match to avoid duplicate
-                             @dsdb
-                             (re-case-insensitive query)
-                             (when ignore-dup query))
-                        (take n))]
-       results))))
+     (let [ignore-dup             (when ignore-dup query)
```
arkRedM, line 416: ignore-dup -> ignore-dup?

also using ignore-dup in filter is making it ambiguous. 
try a better name - query-ignore or better

src/cljs/athens/db.cljs
```diff
@@ -445,18 +440,19 @@
...
-            @dsdb
-            (re-case-insensitive query))
-       (take n)
-       (map get-root-parent-node)
-       (remove nil?)
-       (mapv #(dissoc % :block/_children))))))
+     (let [case-insensitive-query (re-case-insensitive query)]
+       (->>
+         (d/datoms @dsdb :aevt :block/string)
+         (sequence
```
arkRedM, line 446: use ->> instead of (sequence (comp fn1 fn2))
more easier to read

arkRedM (changes requested):
Minor readability improvements

juniusfree:
Thanks @arkRedM. I'll check them and also fix the bug that I found tomorrow.

agentydragon (changes requested):
+1 to @arkRedM's suggestion for `->>` instead of `comp`, otherwise looks good to me

tangjeff0:
So excited for this!!! @juniusfree

src/cljs/athens/db.cljs
juniusfree: @arkRedM

1. This is a transducer that's why I use `sequence` and `comp`. Same as https://github.com/athensresearch/athens/blob/95ca61c4c1ea5554b49a7c196ad1540cc8bea1d2/src/cljs/athens/db.cljs#L446 and https://github.com/athensresearch/athens/blob/95ca61c4c1ea5554b49a7c196ad1540cc8bea1d2/src/cljs/athens/db.cljs#L452
2. Thanks for pointing out the use of `every-pred`. I finally realized its use case.

src/cljs/athens/db.cljs
juniusfree: Noted. I'll update.

src/cljs/athens/db.cljs
arkRedM: I agree this is the use case for transducers, but in terms of readability and what is happening(for someone who is looking at it) will find it easier if all steps are laid out instead of hiding some complexity inside a transducer(makes it easier to debug).

It's totally fine by me to keep it as is, but see if one is more readable than other.

src/cljs/athens/views/athena.cljs
```diff
@@ -165,18 +164,16 @@
...
-                                             vec)})))
-            1000))
+  (fn [query]
+    (if (str/blank? query)
+      (reset! state {:index   0
+                     :query   nil
+                     :results []})
+      (reset! state {:index   0
+                     :query   query
+                     :results (->> (into (search-in-block-content query)
```
arkRedM, line 174: we can write this better
concat is more idiomatic here and we can ditch the ->> from old since it's only one operation on data 

(vec (concat [(exact-node-title)] (node-title..) ))

juniusfree: Updated

src/cljs/athens/db.cljs
juniusfree: @arkRedM Another thing to consider is that `transducers`, at least from what I read, is faster https://groups.google.com/g/clojure/c/iIju063QmG0/m/GKzD_SpvBgAJ

And also, `comp` is read like `->>` when used in transducers. https://clojure.org/reference/transducers#_defining_transformations_with_transducers

But all this is theory and you have more experience. :) Let me know your final thoughts on this. I'll update this to whatever is more appropriate.

arkRedM (approved):
I think all the changes from a code pov are complete. Please do test locally with large dbs if more can be optimized. Thanks

juniusfree:
@arkRedM Thanks again for the feedback.

tangjeff0 (approved):
Holy !@#& this is fast.

src/cljs/athens/db.cljs
tangjeff0: Should be pretty easy to `time` performance between using transducer vs `->>` right? The Google Groups link references CLJ, not CLJS so maybe difference. Either way, I haven't used transducers yet and this is way faster so I love it.

PR #1193, opened 2021-05-18
feat(datalog-console): respond to datalog-console messages in browser

groundedSAGE:
Adds support for the [Datalog Console chrome extension](https://chrome.google.com/webstore/detail/datalog-console/cfgbajnnabfanfdkhpdhndegpmepnlmb)

### DIFF
--- a/src/cljs/athens/core.cljs
+++ b/src/cljs/athens/core.cljs
@@ -6,6 +6,7 @@
     [athens.coeffects]
     [athens.components]
     [athens.config :as config]
+    [athens.db :refer [dsdb]]
     [athens.effects]
     [athens.electron]
     [athens.events]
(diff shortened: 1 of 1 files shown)


tangjeff0 (approved):
Extension doesn't seem to recognize datalog DB, but it does work for browser. DB is recognized when using the Homebase Todo App.

# Athens

image


# Homebase Todo App

image

src/cljs/athens/core.cljs
```diff
@@ -98,6 +117,7 @@
   (style/init)
   (stylefy/tag "body" style/app-styles)
   (listeners/init)
+  (init-datalog-console)
```
arkRedM, line 120: I think we should put this under dev/debug flag.

groundedSAGE:
@tangjeff0 that is strange it appears to show up when the page is refreshed but not on the initial load. Might have to periodically check for the attribute that is used for notification.

src/cljs/athens/core.cljs
groundedSAGE: Yea I'm 50/50 on whether it should be a debug only option. I kind of like the idea of giving users an admin console into their own database. Only downside I can see is some resource usage overhead and the need for more security as the console becomes more powerful.

tangjeff0: I think this can be available to everyone.

PR #1195, opened 2021-05-18
fix(unlinked refs): update unlinked refs when page changes

### LINKED ISSUE #1177: Unlinked references are incorrectly sticky from page to page
**Problem**
Unlinked references show references for earlier-viewed page when unlinked reference section is open when user navigates to an unrelated page

**Steps to reproduce**
1) Navigate to a page that has unlinked references
2) Expand the unlinked references section
3) Navigate to a page that has NO unlinked references
4) Note that the unlinked references section on that page is inappropriately showing the unlinked references for the FIRST page (e.g., unlinked references persist showing in subsequently-viewed pages)
5) Navigate to ANOTHER page that has NO unlinked references
6) Notice that -- *again* -- the page shows the unlinked references from the *first* page

If the user collapses and then again expands the unlinked references section on subsequent pages without linked references, the pages will then no longer (incorrectly) show the unlinked references from the first page.

**Athens Version**
b80

datokrat:
This fixes #1177 (sticky unlinked references).

It basically adds another atom to the local state of the component to keep track of changes of the block UID, and when they occur, it resets the local state.

However, it might be worth considering to move the block or page state from an atom into `app-db`, although I'm not sure how this might work exactly, since more than one instance of the block/page component can be present (which makes global state hard to use).

### DIFF
--- a/src/cljs/athens/views/pages/node_page.cljs
+++ b/src/cljs/athens/views/pages/node_page.cljs
@@ -530,8 +530,13 @@
   Similar to atom-string in blocks. Hacky, but state consistency is hard!"
   [_ _ _ _]
   (let [state         (r/atom init-state)
-        unlinked-refs (r/atom [])]
+        unlinked-refs (r/atom [])
+        block-uid     (r/atom nil)]
     (fn [node editing-uid linked-refs]
+      (when (not= @block-uid (:block/uid node))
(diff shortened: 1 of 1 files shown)


tangjeff0 (approved):
Wow very simple solution!

PR #1203, opened 2021-05-21
fix(#1107): drop support for *** for thematic-break

### LINKED ISSUE #1107: Small parser problem : bold + italics don't work if nothing before in a block
**Problem**

Bold + Italics won't work if there is nothing before in a block
Couldn't find any other combo that doesn't work except bold + italics + something inside.

* `***Test***` doesn't work
* `***^^Test^^***` won't work

When hitting `Enter` the block will split after the three `*` and make a new line.

However, all other combinations, including when there is text before or when the nesting of rendering commands is in a different order : 

* `^^***Test***^^` works
* `Hello ***Test***` works

I've also noticed two strange things : 
* `* Test` will be rendered as `* Test`
* `** Test` will be rendered as `Test` (double star has disappeared)
* `*** Test` will produce the same behaviour by splitting the block in two lines

**Athens Version**
1.0.0-BETA.78 - Intel OSX 10.11.6 - El Capitan

zkry:
This PR addresses issue #1107. 

The root cause was that the block parser was interpreting a `***` to be a thematic block which in turn would cause something like `***bold and emphasis***` to be `[:block [:thematic-block "***"] [:paragraph-text "bold and emphasis***"]]`.  I added a negative lookahead in the regex of thematic block to prevent it from matching if there is a `***` which could be interpreted as the end of an emphasis + strong-emphasis.


tangjeff0:
@neotyk please review

neotyk:
@zkry Thank you for PR.

Maybe we could just drop support for `***` as thematic break?

src/cljc/athens/parser/impl.cljc
```diff
@@ -24,7 +24,7 @@ block = (thematic-break /
          fenced-code-block /
          block-quote /
          paragraph-text)*
-thematic-break = #'[*_-]{3}'
+thematic-break = #'(\\*{3}(?!\\w[^\n]*(?<!\\s)\\*{3}(?!\\w))|[_-]{3})'
```
neotyk, line 27: That's a lot of regex.
I think we could drop support of `***` as thematic break.

neotyk (changes requested):

src/cljc/athens/parser/impl.cljc
zkry: I agree, it makes more sense to do this. I can update the PR for this.

neotyk:
LGTM

tangjeff0 (changes requested):
Failing test

tangjeff0 (approved):
Awesome first PR @zkry :muscle: !!!

PR #1209, opened 2021-05-21
feat(toolbar): add database menu

shanberg:
- Should replace "Merge from Roam" button in app toolbar: now lives as an option on the current database.
- Should replace "Database" button in app toolbar: filesystem workflow is started from inside the menu instead.
- Should replace sync status indicator in toolbar: status indicated on database icon instead.
CleanShot 2021-05-21 at 13 13 16@2x


sid597:
I am taking up this issue

sid597:
Db-picker checklist : 
Note : syncing and local storage operations take some.
 - [x]  Add a new DB 
 - [x]  Restart Athens to see if the newly added DB is there along with all the previous DB
 - [x]  Open a new DB 
 - [x]  Restart Athens to check if DB added in the last session is still there and is selected by default
 - [x]  Move the current DB: Expect to see the new location reflected in DB-picker
 - [x]  Delete a DB: Remove the current DB and select a new DB from the list
 - [x]  Restart Athens and check if delete is reflected in the list
 - [x]  Select a DB from the list: Expected to see pages related to the selected DB
 - [x]  Check sync: Make changes in the current DB and check if the sync is working
 - [x]  Change DB while sync: Should not be able to switch DB while the sync is in progress in current DB
 - [x]  Delete the last DB in list: Should not be able to delete the last remaining DB in the list 
 - [ ]  Rename a DB: Sync up with stuart
 - [ ]  Remote DB: Sync up with Alex and Stuart

tangjeff0:
Where's your code for this @sid597 ?

sid597:
@tangjeff0 I was refactoring and reformatting my code it was terrible before, then after that had to recheck the above-mentioned test cases and after that, I and @shanberg figured out how to push my changes. A few things I want to address : 

- Also, note that currently, Delete will not actually delete from the disk please read the last point in the list above.
- Remote DB implementation is pending because I think we should first clean up some current code then add this, also cannot test it out locally.
- After a DB is opened, added, deleted it takes some time to save those changes (my theory, this might actually be a bug) so if we close and reopen Athens just after doing these operations it might seem like the last operations did not go through.
- I tested this from my side based on the above points this by no means exhaustive testing so please add cases I missed and also if I wrongly marked some test done/tested correclty but it is not.

sawhney17:
Question about the renaming of the DB. Will the image links be changed if the name of the DB or folder it is in is renamed? Because changing that could possibly break all the images in someones DB

tangjeff0 (changes requested):
Overall great job. Some general feedback that is actually not bad at all:

- Did a great job documenting functionality and your thoughts. I don't think documenting the types is necessary.
- Don't need to start a new line so much on comments or for each new set of brackets/parens.

src/cljs/athens/electron.cljs
```diff
@@ -199,9 +201,177 @@
...
+        the set of all paths is O(n).
+      - Also re-frame is in-memory so saving data which can be
+        easily calculated is not a good idea I think"
+
+    [db-list check-path]
+    (let [path-set (into #{}
+                         (reduce
+                           (fn [paths db-list-item]
+                             (conj paths (get db-list-item :path)))
+                           [] db-list))]
```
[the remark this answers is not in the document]
sid597: Fun fact I first thought of using a map then thought more and somehow arrived at the conclusion I need to overcomplicate this. `get` is a relic of a past commit I missed to replace this.

src/cljs/athens/electron.cljs
```diff
@@ -199,9 +201,177 @@
...
+        (if (not duplicate)
+          (let [dbname  (get-db-name dbpath)
+                newdb   {:path dbpath
+                         :name dbname}
+                all-dbs (conj
+                          current-db-list
+                          newdb)]
+            {:db       (assoc db :db-picker/all-dbs all-dbs)
+             :dispatch [:local-storage/set-db-picker-list]})
+          (js/alert (str "Database already in the list"))))))
```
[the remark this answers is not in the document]
sid597: Yeah, I checked there is already an alert event, I will use that.

src/cljs/athens/events.cljs
```diff
@@ -25,8 +25,15 @@
...
-  (fn [_ _]
-    db/rfdb))
+  (fn [db [_]]
+    "This reframe db values are reset whenever a new db
+     is added i.e on open, move and add, due to this resetting
+     db-list also gets reset, we want to remember the previous
+     db so I think I should copy previous db list, reset it and
+     update the new rfdb with  db-list"
+    (let [previous-db-list (get db :db-picker/all-dbs)]
+      db/rfdb
```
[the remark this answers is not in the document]
sid597: This sets the reframe's DB to the initial state with default values.

src/cljs/athens/events.cljs
```diff
@@ -25,8 +25,15 @@
...
+      (assoc db :db-picker/all-dbs previous-db-list))))
```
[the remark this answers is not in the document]
sid597: Yes, as I stated above reframe's the state is reset after a new file is opened or added, this also resets the db-picker list to an empty vector, so the idea here is, before reseting the db state, get the current db-list and after reset rebind the previous value of db-picker list.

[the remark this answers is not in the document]
sid597: But this is not the correct way to solve this I already have an event that checks if local storage has the DB list and if does then update reframe's state with that list so I will use this event.

src/cljs/athens/views/app_toolbar.cljs
```diff
@@ -158,15 +146,15 @@
                [:<>
                 [:> Replay]
                 [:span "Re-connect with remote"]]])
-            [button {:on-click #(swap! merge-open? not)
+            #_ [button {:on-click #(swap! merge-open? not)
```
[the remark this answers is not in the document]
sid597: @shanberg

src/cljs/athens/views/app_toolbar.cljs
```diff
@@ -158,15 +146,15 @@
...
                      :title "Merge Roam Database"}
              [:> MergeType]]
             [button {:on-click #(router/navigate :settings)
                      :title "Open Settings"
                      :active   (= @route-name :settings)}
              [:> Settings]]
 
-            [button {:on-click #(dispatch [:modal/toggle])
+            #_ [button {:on-click #(dispatch [:modal/toggle])
```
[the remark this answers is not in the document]
sid597: @shanberg

sid597:
@sawhney17  good catch, yeah will handle that. Under the hood, I am thinking of mapping `rename` to `move`, rename can have a different UI for taking the new name input but after that, it is the same as `move` I think, and `move` already takes care of moving images.

src/cljs/athens/electron.cljs
```diff
@@ -214,161 +216,122 @@
...
+      - Also re-frame is in-memory so saving data which can be easily calculated is
+        not a good idea I think"
     [db-list check-path]
-    (let [path-set (into #{}
-                         (reduce
-                           (fn [paths db-list-item]
-                             (conj paths (get db-list-item :path)))
-                           [] db-list))]
-      (contains? path-set check-path)))
+    (contains? (into #{} (map #(:path %) db-list)) check-path))
```
tangjeff0, line 235: Do you know if using `set` is better than `into #{}`? Genuine question, I don't know the difference.

src/cljs/athens/electron.cljs
```diff
@@ -214,161 +216,122 @@
...
 
   (reg-event-fx
     :db-picker/remove-db-from-list
     (fn [{:keys [db]} [_ db-path]]
-      "Remove the selected db from db-list.
-       Update local storage value with the new db list
-       Args : db-path : type str"
+      "Remove the selected db from db-list. Update local storage value with the new db list"
       (let [current-db-list (:db-picker/all-dbs db)
             new-db-list     (into [] (filter
```
tangjeff0, line 286: Do you know if using `vec` is better than `into []`? Genuine question

src/cljs/athens/electron.cljs
```diff
@@ -214,161 +216,122 @@
...
+      is happening, instead show an alert describing the situation."
+      (let [file-exists? (and db-path (.existsSync fs db-path))]
+        (cond
+          file-exists? (cond
+                         synced? {:dispatch-n [[:db/update-filepath db-path]
+                                               [:boot/desktop]]}
+                         :else   {:dispatch  [:alert/js "Database is saving your changes, if you switch now your changes will not be saved"]})
 
+          :else        {:dispatch-n [[:alert/js "This database does not exist, removing it from list"]
+                                     [:db-picker/remove-db-from-list db-path]]}))))
```
tangjeff0: I did not mean to make a nested `cond` from nested `if` lol.

I meant something like

```
(cond 
  (and file-exists synced?) {}}
  (and file-exists (not synced?) {})
  :else {})
```

In otherwords, nested `if` or nested `cond` is hard to read (for me), so I tend to flatten into a flat `cond` and use `and`.

src/cljs/athens/electron.cljs
```diff
@@ -214,161 +216,122 @@
...
       (let [val                 (cljs.reader/read-string
-                                  (js/localStorage.getItem
-                                    "db-picker/all-dbs"))
+                                  (js/localStorage.getItem "db-picker/all-dbs"))
             current-db-filepath (:db/filepath db)]
         (cond
-          (nil? val)  {:dispatch [:db-picker/add-new-db
-                                  current-db-filepath]}
+          (nil? val)  {:dispatch [:db-picker/add-new-db current-db-filepath]}
           :else       {:db (assoc db :db-picker/all-dbs val)}))))
```
tangjeff0: Just `if` here, no need for `cond`

tangjeff0 (changes requested):

tangjeff0:
I'm not sure if there is a strong consensus on character count per line, indentation, new lines in Clojure.

Just did 2nd code review. Main thing I'm concerned with is how we can test this before merging or releasing. I'm 99% sure there is a bug here that could lead to data loss. It's almost impossible to detect this without using extensively, just because so much state is involved between filesystem, localStorage, and Athens.

src/cljs/athens/electron.cljs
sid597: Yeah I also don't know the difference, I remembered using `into #{}` to solve 4clojure problems so here also I used it. From what I could find they both have the [same performance](https://insideclojure.org/2015/01/06/set-perf/) as of Clojure 1.7

src/cljs/athens/electron.cljs
sid597: As of clojure 1.7 depending on the type that is being converted one of them is better than the other, the data is shown [here](https://insideclojure.org/images/20150107/numbers.png)

src/cljs/athens/electron.cljs
sid597: Ok I will fix it

src/cljs/athens/electron.cljs
sid597: Ok will fix it, after using if I felt it hard to read so went with `cond`

sid597:
Yeah I am also concerned about this so I am using it myself, as I said before
> "After a DB is opened, added, deleted it takes some time to save those changes (my theory, **this might actually be a bug**)"

Here I said "save" but did not mention save where local-storage, disk, the reason being I am also not sure where the issue is so I was thinking of fully understanding the current implementation regarding the filesystem, syncing etc. What this could have enabled me to do was understand all the points where things could go wrong, also could refactor the current implementation to make it more simple and clean. Controlled tests can also be written to simulate race conditions, I mentioned [this distributed system testing ](https://tech.redplanetlabs.com/2021/03/17/where-were-going-we-dont-need-threads-simulating-distributed-systems/) resource I think we can use this here.

A hack (which is a dirty way) I can think of is, when a user wants to switch db we check when did the last sync happen and if it has not been "x" sec since then, then we show an alert to wait and let filesystem/local storage sync. In my tests I have found data loss does not happen if one waits some "x" sec after sync and then switch, quit or restart. But this approach is wrong on many levels and I don't think we should take this even if it works. Eventually, we have to look and refactor our code so why not do it now.

src/cljs/athens/views/app_toolbar.cljs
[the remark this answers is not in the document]
shanberg: @tangjeff0 This had been moved into the DB menu since it affects the DB directly and is mostly a one-time action, but the rest of the code hasn't caught up to that change yet. I'll uncomment this for now.

tangjeff0:
@sid597 so are you able to consistently reproduce this issue of saving?

What else needs to be done here? @sid597 has a checkilst, last few points seem small or not necessary for this PR. @shanberg has lots of checkmarks, but it seems out of date.

# @sid597's checklist

image

# @shanberg's checklist

image

sid597:
@tangjeff0 I can't say consistently because I did not keep track of each time I tested, but I can say that I don't remember when this worked anytime. I will test this ~50 times while keeping track of the result to provide a more consistent reply. Yeah, my checklist is updated

shanberg:
> # @shanberg's checklist

My checklist is obsoleted by @sid597's, so I've removed it.

src/cljs/athens/electron.cljs
neotyk:
```clojure
(into #{} [1 2 1])
;=> #{1 2}

(set [1 2 1])
;=> #{1 2}

;; Though I expected it to behave like `clojure.core/sorted-set`
(apply sorted-set [1 2 1])
;=> #{1 2}
(sorted-set [1 2 1])
;=> #{[1 2 1]}
```

Would this be more readable:
```clojure
(seq (filter #(= check-path (:path %)) db-list))
```

src/cljs/athens/electron.cljs
```diff
@@ -199,9 +203,136 @@
...
+        not a good idea I think"
+    [db-list check-path]
+    (contains? (into #{} (map #(:path %) db-list)) check-path))
 
   ;; Events
 
+  (reg-event-fx
+    :db-picker/add-new-db
+    (fn [{:keys [db]} [_ dbpath]]
+      "Add a new to db picker list.
```
neotyk, line 242: `docstring`s should go before `args`

src/cljs/athens/electron.cljs
```diff
@@ -199,9 +203,136 @@
...
+       Can be invoked when a db is added, opened, moved, selected or removed.
+       Update local storage value with the new db list"
+      (let [current-db-list (:db-picker/all-dbs db)
+            duplicate?       (check-duplicate-db
+                               current-db-list
+                               dbpath)]
+        (if duplicate?
+          {:dispatch [:alert/js (str "Database already in the list")]}
+          (let [dbname  (get-db-name dbpath)
+                newdb   {:path dbpath
```
neotyk, line 252: Can we add here `:type` please?
Currently `:fs`, with self-hosted `:lan-party`

src/cljs/athens/electron.cljs
```diff
@@ -199,9 +203,136 @@
...
+                         :name dbname}
+                all-dbs (conj current-db-list newdb)]
+            {:db       (assoc db :db-picker/all-dbs all-dbs)
+             :dispatch [:local-storage/set-db-picker-list]})))))
```
neotyk, line 256: With recent `re-frame` updates it's recommended to use `:fx` instead of `:dispatch`.

https://github.com/Day8/re-frame/blob/master/docs/Effects.md#the-effects-map

src/cljs/athens/electron.cljs
```diff
@@ -199,9 +203,136 @@
...
+      :local-storage/set-db-picker-list
+      (fn [{:keys [db]} _]
+        "Save current db-picker list to local storage. Why using `pr-str`? Checkout
+         this link for discussion on how to serialize data to store"
+        ;; https://stackoverflow.com/questions/67821181/how-to-set-and-get-a-vector-in-local-storage
+        (let [current-db-list (:db-picker/all-dbs db)]
+          {:local-storage/set! ["db-picker/all-dbs" (pr-str current-db-list)]})))
+
+  (reg-event-fx
+    :local-storage/check-db-picker-list
```
neotyk, line 269: Rename candidate: this is actually reading/creating this list.

src/cljs/athens/views/db_menu/core.cljs
```diff
@@ -0,0 +1,150 @@
...
+                            (js/alert "Can't remove last db from the list")
+                            (do
+                              (dispatch [:db-picker/remove-db-from-list (:path db)])
+                              (dispatch [:db-picker/delete-db (:path db)])))}
+               "Delete"]])]))
+
+
+(defn db-menu
+  []
+  (r/with-let [ele (r/atom nil)]
```
neotyk, line 101: `ele` is used for 2 purposes, which often is surprising when reading code.

1st usage as DOM element itself in `:anchorEl`, to get DOM element of reagent component use `:ref` (correct way of doing this)[https://github.com/reagent-project/reagent/blob/master/doc/FAQ/UsingRefs.md#answer]

2nd usage is predicate if menu is open. it's better to express it explicitly as so, or maybe even manage it re-frame itself?

sid597:
@tangjeff0 I tested the saving issue in a constrained environment and I found out the case when saving does not happen instantly (to humans) and this has nothing to do with local storage, file sync, the saving issue is related to how Athens is closed. 

> This is my setup and how I start and stop Athens:
- I use ubuntu with i3wm (tilling manager)
- To start: Athens I do `lein dev` and once the build is complete I do `yarn run electron .`
- To exit: I have 2 options
   - `ctrl+shift+q` (ctrl is my `mod` key)
   - `ctrl+c` in terminal
 
> What does it have to do with Athens and this issue?

 After making some changes that are stored in local storage if  `ctrl+c` is used to exit Athens before ~45 sec from the last local storage change, then the local storage data would be lost. I tested this scenario many times and found the magic no 45. The local storage data is saved if `ctrl+shift+q` is used to exit, also if Athens in the midst of syncing and `ctrl+c` is used Athens will abort without showing the "sync in progress" alert.

> Why does this happen?

This is just my speculation but I think: 
- `ctrl+shift+q`: Chromium saves its local storage and session data on the disk before exiting.
- `ctrl+c`: This kills the process with SIGINT, which terminates the process unless it is handled/ignored by the target. The handling is left up to the application, from the behaviour we have seen above I think chromium does not handle this and hence does not save its local storage data on the disk before exiting. And by default, there is ~45sec time before 2 consecutive local storage to disk saving on chromium.

I could not confirm the above regarding chromium by googling, I can dig deeper into this issue if that is needed.

> What should/can we do?
- We can handle the `ctr-c` but I don't know how that would help in saving the local storage value to disk.
- This is I think mostly a developer side related issue because normal users will use the "close" button in the app.
- Dig deep to find the root cause of this issue so can be fixed in a proper way.


> End result : 

I am confident (based on data) that we can merge this and will not face data loss issues when the user uses the "close" button to exit or in the case of devs using tiling manager `mod+shift+q` is used.

Note: This is tested from ubuntu + i3wm+ exit via terminal, this can be simulated in Windows (I think task manager kill would the same as exit via terminal) and MacOs. Unfortunately, I cannot test this on Mac and Windows.

src/cljs/athens/views/db_menu/core.cljs
sid597: @shanberg

src/cljs/athens/electron.cljs
sid597: @neotyk Ok I did not know that, in the website docs they are [still using](https://day8.github.io/re-frame/EffectfulHandlers/#step-1-of-plan) `:dispatch`

src/cljs/athens/electron.cljs
sid597: @neotyk I think that would not work for `fn` because `fn` spec is : `(fn name? [params*] exprs*)`

src/cljs/athens/electron.cljs
sid597: > `(seq (filter #(= check-path (:path %)) db-list))` 

This looks better and readable, also learned `seq` is the idiomatic way of checking if a seq is empty or not.

tangjeff0:
It's interesting that you found 45 seconds. I think the default value is 15 seconds, and 15 is `step` for that input.

localStorage should update faster than every 45 seconds. It is a synchronous API.

sid597:
@tangjeff0  Yeah, I also thought  that localStorage should update faster than 45 secs but I could not find any resource (checked out mdn and `chrome.storage` API which is used for extensions) which talked about this , do you know of resource which sheds light on this?

tangjeff0:
@sid597 localstorage is fundamentally synchronous. We have to some digging to find why it does not update immediately.

sid597:
@tangjeff0 I think localStorage is synchronous for setting values in memory but it is not the case for the data getting written to physical disk. I mean the key-value getting reflected in the dom/dev-tools does not guarantee that data is physically written. I think this is synchronous for memory updating and async for the physical disk.

src/cljs/athens/electron.cljs
oxalorg: @tangjeff0 @sid597 Yup `vec` should be faster in most of the cases than `into` but not significantly, so unless this is in a hot path it shouldn't matter much.

`vec` is also preferred over `into` in the style guide: https://github.com/bbatsov/clojure-style-guide#converting-sequences-to-vectors

**BUT there's a catch here**: `filter` produces a `lazy-seq` which does some extra work with chunking and stuff. So if we really want performance (and we want this to be evaluated eagerly) we should use `filterv` which removes the overhead of first converting to lazy-seq chunks and directly returns a vector (fun-fact: it uses a transient vector underneath so it's even faster)

Another thing to keep in mind would be if we're combining multiple filter/maps/other operations then using a transducer would be the most effecient. It will do a single pass over the initial collection without creating additional intermediary collections.

src/cljs/athens/electron.cljs
sid597: @neotyk  where will we use `:type`? I am thinking of using `:is-remote` because the information if a db is remote or not is needed in some places. Should I add both?

sid597:
@filipesilva we should close this one also

filipesilva:
Closing due to https://github.com/athensresearch/athens/pull/1378#issuecomment-896991480

PR #1212, opened 2021-05-22
enhance(ui): Stop content shift when scrollbars appear/disappear

zolrath:
Overlay Scrollbars were released into the wild effectively on accident, removed, readded and seem to be generally confused.

They have recently been [readded to Chrome](https://chromium-review.googlesource.com/c/chromium/src/+/2739606) and by using `overflow-y: overlay` we remove the content shift that occurs when the scrollbar appears and disappears.

In the future this problem can be solved using [scrollbar-gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) but this feature is not yet implemented.

This video shows the shift before/after this change:

https://user-images.githubusercontent.com/454563/119237988-b36d3f80-baf4-11eb-86ec-83f25f2debbe.mp4


zolrath:
I'd like to do a test of this with the web version of Athens as Firefox doesn't support `overlay` and I'd like to ensure proper fallbacks are in place.

shanberg:
We can probably do this by getting the browser from the useragent at the top level and applying this to Firefox only, but we need a solution to Stylefy's parent selector problem first: https://github.com/Jarzka/stylefy/issues/59#issuecomment-846449821

zolrath:
I think I've found a fairly simple solution in stylefy by using:
```clojure
   :overflow-y "auto"
   ::stylefy/supports {"overflow-y: overlay"
                      {:overflow-y "overlay"}}
```
I've maintained the previous behavior on the web version using Firefox while allowing Electron for Desktop and the web version using Chrome to benefit from the lack of content shift.

tangjeff0:
Pleasee review @shanberg

src/cljs/athens/main/core.cljs
```diff
@@ -40,6 +40,7 @@
                                                    :nodeIntegration true
                                                    :worldSafeExecuteJavaScript true
                                                    :enableRemoteModule true
+                                                   :enableBlinkFeatures 'OverlayScrollbars'
```
shanberg, line 43: Could you add a comment here along the lines of "Remove when `scrollbar-gutter` is supported"?

shanberg (changes requested):
Looks great, 2 qs

1. The build/style check should succeed if you run `cljstyle fix` locally
2. Do you have a clear sense of when to use `overlay` and when not to? There are a handful of other scroll areas in the app and I'm wondering if there's any reason not to update those too.

zolrath:
> 1. The build/style check should succeed if you run cljstyle fix locally
Will do!

> 2. Do you have a clear sense of when to use `overlay` and when not to? There are a handful of other scroll areas in the app and I'm wondering if there's any reason not to update those too.

I went with the minimal amount of changes to fix the document shift but effectively overlay floats the scrollbar instead of making it take up space in the container.
The only reason I could see to not use it would be in situations where there is important content very close to the right edge of the content which would be obscured/look odd if the thin scrollbar was overlaid on top of it.

src/cljs/athens/main/core.cljs
zolrath: Done!

shanberg (approved):
Beautiful. Excellent little improvement.

zolrath:
Is there a cljstyle configuration file? Running `cljstyle fix` recommends changes to comments in dozens of files, mostly related to changing some `;` comments to `;;` and some `;;;` comments to `;;`

tangjeff0 (approved):
Awesome first PR @zolrath !

PR #1217, opened 2021-05-23
perf(blocks): reduce blocks DOM weight using grid, refactor drop-indicators

### LINKED ISSUE #1221: Margin/gap around empty blocks makes clicking on them difficult
**Problem**

To reproduce, hit `Enter` a few times to create some empty blocks. Move your mouse cursor past them, and you'll see instances where the cursor changes from a text cursor to the default pointer. When it's in that default state, clicking does not select that block, even though it looks like your mouse is aligned with the block or its bullet on the left-hand side.

This gap or margin seems to exist only on blocks. If there's text on two blocks next to each other, there's never any gap between them, which makes selecting one or the other much easier.

See the video below for an example.

**Screenshots/Demo**

https://user-images.githubusercontent.com/1153921/119266324-377c0180-bb9f-11eb-9573-48a6394fd7d5.mp4

**Athens Version**

1.0.0-beta82

shanberg:
**Goal**: Improve performance of long pages by reducing the DOM weight of blocks.
**Method**: Only render what's absolutely necessary for a given interaction.
- Block children toggle only rendered when there are children
- Block textarea(!) only rendered when editing or mouse is over the block
- Drop area indicators only rendered during the appropriate dragging state

**Result**: Shaved 2.7MB off of my Welcome page. Benefits are greater the more blocks are shown on a page.

This might also improve rendering performance by reducing the amount of work that `autosize/textarea` needs to do under some conditions, but I didn't nail down a great test for that, and the DOM weight reduction is good enough alone.

Also fixes #1221

CleanShot 2021-05-23 at 06 29 03@2x


src/cljs/athens/views/blocks/content.cljs
```diff
@@ -335,19 +334,18 @@
...
+                               :on-change      (fn [e] (textarea-change e uid state))
+                               :on-paste       (fn [e] (textarea-paste e uid state))
+                               :on-key-down    (fn [e] (textarea-keydown/textarea-key-down e uid state))
+                               :on-blur        (fn [_] (db/transact-state-for-uid (or original-uid uid) state))
+                               :on-click       (fn [e] (textarea-click e uid state))
+                               :on-mouse-enter (fn [e] (textarea-mouse-enter e uid state))
+                               :on-mouse-down  (fn [e] (textarea-mouse-down e uid state))}])
          ;; TODO pass `state` to parse-and-render
-         [parse-and-render (:string/local @state) (or original-uid uid)]
-         [drop-area-indicator/drop-area-indicator #(when (= :child (:drag-target @state)) {;;:color "green"
```
shanberg, line 351: I'm not sure why the drop-area-indicator was in the `content` as well as in `core`.

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -204,7 +219,7 @@
                                    (fn [{:block/keys [original-uid uid] :as block}]
                                      (assoc block :block/uid (or original-uid uid)))
                                    block)
-             {:search/keys [] :keys [dragging drag-target]} @state
```
shanberg, line 207: Kondo said this `drag-target` was unused so I removed it, but I'm not sure that was the right thing to do.

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -55,10 +55,15 @@
...
+                                :background (style/color :link-color :opacity-lower)}]
                      [:&.is-selected:after {:opacity 1}]
-                     [:.block-body {:display "flex"
+                     [:.block-body {:display "grid"
+                                    :grid-template-columns "1em 1em 1fr auto"
+                                    :grid-template-rows "0 1fr 0"
+                                    :grid-template-areas "
+                                      'above above above above'
+                                      'toggle bullet content refs'
+                                      'below below below below'"
```
shanberg, line 66: Defining a grid here is what made most of this possible, so we don't have to leave elements in as placeholders.

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -204,7 +219,7 @@
                                    (fn [{:block/keys [original-uid uid] :as block}]
                                      (assoc block :block/uid (or original-uid uid)))
                                    block)
-             {:search/keys [] :keys [dragging drag-target]} @state
+             {:search/keys [] :keys [dragging]} @state
```
shanberg, line 222: Kondo said this `drag-target` was unused so I removed it, but I'm not sure that was the right thing to do.

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -97,9 +105,15 @@
...
+                            :grid-area "refs"
                             :z-index (:zindex-dropdown style/ZINDICES)
                             :visibility (when-not (pos? count) "hidden")})
-   [buttons/button {:primary true :on-click #(rf/dispatch [:right-sidebar/open-item uid])} count]])
+   [buttons/button {:primary true
+                    :on-click (fn [e]
+                                (doall
+                                  (.. e stopPropagation)
+                                  (rf/dispatch [:right-sidebar/open-item uid])))}
+    count]])
```
shanberg, line 116: I tried to cover all the cases where we'd need to catch clicks and stop propagation. Not 100% sure I got them all.

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -217,35 +232,36 @@
...
           [presence/presence-popover-info uid {:inline? true}]
 
-          [drop-area-indicator/drop-area-indicator #(when (= drag-target :above) {:opacity "1"})]
+          (when (= (:drag-target @state) :above) [drop-area-indicator/drop-area-indicator {:grid-area "above"}])
 
           [:div.block-body
-           [:button.block-edit-toggle
-            {:on-click (fn [e]
-                         (when (false? (.. e -shiftKey))
-                           (rf/dispatch [:editing/uid uid])))}]
```
shanberg, line 239: This button was causing some crazy reagent updates. I replaced it with a conditional `on-click` on the parent.

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -255,13 +271,12 @@
                      (or (and (true? linked-ref) (:linked-ref/open @state))
                          (and (false? linked-ref) open)))
             (for [child children]
-              [:div {:key (:db/id child)}
+              [:<> {:key (:db/id child)}
```
shanberg, line 274: Totall unnecessary div here

src/cljs/athens/views/blocks/toggle.cljs
```diff
@@ -44,18 +45,18 @@
...
+                               :on-click (fn [e]
+                                           (doall
+                                             (.. e stopPropagation)
                                              (if (true? linked-ref)
                                                (swap! state update :linked-ref/open not)
-                                               (toggle [:block/uid uid] open)))})
-     [:> KeyboardArrowDown {:style {:font-size "16px"}}]]
-    [:span (stylefy/use-style block-disclosure-toggle-style)]))
+                                               (toggle [:block/uid uid] open))))})
+   [:> KeyboardArrowDown {:style {:font-size "16px"}}]])
```
shanberg, line 61: This is an example of how we were leaving a placeholder element to make the styling work right. That styling logic is all handled by the container's grid settings now.

joelhans:
Tested this out locally and can confirm it fixes #1221! Love the fast turnaround.

src/cljs/athens/views/blocks/content.cljs
```diff
@@ -335,19 +334,18 @@
...
-                                                                                           :opacity 1})]]))))
+         [parse-and-render (:string/local @state) (or original-uid uid)]]))))
```
shanberg, line 350: I'm not sure why the drop-area-indicator was in the `content` as well as in `core`.

tangjeff0: Good question. `drop-area-indicator` shows up here as a target for dropping a block to the child of another block. This part of Athens is not done very well, logically shouldn't be part of `content`, but it shouldn't be removed in this PR.

shanberg: I've restord the child drop indicator and it’s up in `core` now.

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -255,13 +271,13 @@
...
                [block-el child
                 (assoc linked-ref-data :initial-open (contains? parent-uids (:block/uid child)))
                 opts]]))
 
-          [drop-area-indicator/drop-area-indicator #(when (= drag-target :below) {;; :color "red"
-                                                                                  :opacity "1"})]])))))
+          (when (= (:drag-target @state) :child) [drop-area-indicator/drop-area-indicator {:style {:grid-area "below"} :child true}])
+          (when (= (:drag-target @state) :below) [drop-area-indicator/drop-area-indicator {:style {:grid-area "below"}}])])))))
```
tangjeff0, line 280: Two drop indicators? `:child` drop indicator is no longer in `block-contents`?!? Nice!!!

src/cljs/athens/views/blocks/toggle.cljs
```diff
@@ -44,18 +45,18 @@
...
-                                 :on-click (fn [_]
+  [{:block/keys [open uid]} state linked-ref]
+  [:button (stylefy/use-style block-disclosure-toggle-style
+                              {:class    (if (or (and (true? linked-ref) (:linked-ref/open @state))
+                                                 (and (false? linked-ref) open))
+                                           "open"
+                                           "closed")
+                               :tab-index 0
+                               :on-click (fn [e]
+                                           (doall
```
tangjeff0, line 56: Why `doall`?

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -217,35 +232,36 @@
...
+           (when (:context-menu/show @state)
+             [context-menu/context-menu-el uid-sanitized-block state])
            [bullet/bullet-el block state linked-ref]
            [tooltip/tooltip-el uid-sanitized-block state]
            [content/block-content-el block state]
 
-           (when-not (:block-embed? opts)
-             [block-refs-count-el (count _refs) uid])]
+           (when (> (count _refs) 0)
+             (when-not (:block-embed? opts)
```
tangjeff0, line 263: Combine into one `when`?

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -97,9 +105,15 @@
 (defn block-refs-count-el
   [count uid]
   [:div (stylefy/use-style {:margin-left "1em"
+                            :grid-area "refs"
                             :z-index (:zindex-dropdown style/ZINDICES)
                             :visibility (when-not (pos? count) "hidden")})
-   [buttons/button {:primary true :on-click #(rf/dispatch [:right-sidebar/open-item uid])} count]])
+   [buttons/button {:primary true
+                    :on-click (fn [e]
+                                (doall
```
tangjeff0, line 113: Do we need `doall`?

src/cljs/athens/views/blocks/content.cljs
```diff
@@ -335,19 +334,18 @@
...
-                             ;; :auto-focus     true
-                             :id             (str "editable-uid-" uid)
-                             :on-change      (fn [e] (textarea-change e uid state))
-                             :on-paste       (fn [e] (textarea-paste e uid state))
-                             :on-key-down    (fn [e] (textarea-keydown/textarea-key-down e uid state))
-                             :on-blur        (fn [_] (db/transact-state-for-uid (or original-uid uid) state))
-                             :on-click       (fn [e] (textarea-click e uid state))
-                             :on-mouse-enter (fn [e] (textarea-mouse-enter e uid state))
-                             :on-mouse-down  (fn [e] (textarea-mouse-down e uid state))}]
+         (when (or (:render-editable @state) editing?)
```
tangjeff0, line 337: Clever

src/cljs/athens/views/blocks/core.cljs
tangjeff0: You are using the key `:drag-target` instead of binding a variable for `drag-target`. Both work.

src/cljs/athens/views/blocks/core.cljs
tangjeff0: If div is unnecessary, is fragment also unnecessary?

src/cljs/athens/views/blocks/content.cljs
tangjeff0: Wait can you actually explain this a little? And what is happening on `mouse-enter` and `mouse-leave`?

src/cljs/athens/views/blocks/drop_area_indicator.cljs
```diff
@@ -24,21 +24,37 @@
...
+                                       :border-bottom-left-radius 0}]
+                     ["&.child:before" {:position "absolute"
+                                        :content "''"
+                                        :width "0.3rem"
+                                        :height "0.3rem"
+                                        :border-radius "10em"
+                                        :left "var(--indent)"
+                                        :top "50%"
+                                        :transform "translateY(-50%) translateX(calc((0.3rem + 2px) * -1))"
+                                        :border [["2px solid " (style/color :link-color :opacity-high)]]}]]})
```
tangjeff0, line 53: Not sure where it happens but the little white dot for child drop is such a nice detail.

tangjeff0 (changes requested):
Some questions and requests. This is a hugeeee PR! Amazing work @shanberg. Before this, we were creating 3 drop-area-indicators elements on the DOM for each block.

CSS and component decoupling is also much more readable.

Grid is magic.

src/cljs/athens/views/blocks/core.cljs
shanberg: The fragment includes the key, which is necessary.

src/cljs/athens/views/blocks/core.cljs
shanberg: We could get away with a single drop-indicator if we wanted to be more clever about it, but I didn't think it was worth the complexity.

src/cljs/athens/views/blocks/content.cljs
shanberg: I've renamed this var and added some comments around where it's used.

src/cljs/athens/views/blocks/core.cljs
tangjeff0: Got it. Suppose we could add a key somehow without a fragment -- Do you think there would be a performance boost removing the fragment?

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -217,34 +231,38 @@
...
-            {:on-click (fn [e]
-                         (when (false? (.. e -shiftKey))
-                           (rf/dispatch [:editing/uid uid])))}]
-
-           [toggle/toggle-el uid-sanitized-block state linked-ref]
-           [context-menu/context-menu-el uid-sanitized-block state]
+           (when (seq children)
+             [toggle/toggle-el uid-sanitized-block state linked-ref])
+           (when (:context-menu/show @state)
+             [context-menu/context-menu-el uid-sanitized-block state])
```
tangjeff0, line 260: Do you think it is faster to use the `when` outside of the element rather than inside?

src/cljs/athens/views/blocks/core.cljs
```diff
@@ -217,34 +231,38 @@
...
            [bullet/bullet-el block state linked-ref]
            [tooltip/tooltip-el uid-sanitized-block state]
            [content/block-content-el block state]
 
-           (when-not (:block-embed? opts)
+           (when (and (> (count _refs) 0) (not= :block-embed? opts))
```
tangjeff0, line 265: I personally prefer `(pos? (count _refs))` over `(> (count _refs) 0)`

tangjeff0 (approved):
Great job @shanberg :) Small comments, but no requests.

src/cljs/athens/views/blocks/core.cljs
shanberg: I think the fragment is the way we add the key, and the fragment exists only in the Reagent tree, not the DOM, so shouldn't result in any performance hit.

src/cljs/athens/views/blocks/core.cljs
shanberg: I'd assume it’s faster outside because it doesn't have to evaluate anything about the element itself, just the outer `when`, but this isn't my strong suit.

src/cljs/athens/views/blocks/core.cljs
shanberg: Preferences are allowed. Fixed.

Issue #607, opened 2021-02-03
rfct(keybindings): rewrite `update-query`


tangjeff0:
- query value should always derive directly from value within (()), [[]], or following # or /
- results should always derive from query
- when caret position changes (i.e. via mouse or arro keys), app should know if dropdown should be open or not

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 300.0 USD (300.0 USD @ $1.0/USD)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/athensresearch/athens/607/100024995).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/athensresearch/athens/607/100024995).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
* $2,655,940.31 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

ssjoleary:
Hey @tangjeff0, off the top of your head, can you remember what the behaviour of those weird edge cases was? I think I'll give this issue a shot next.

tangjeff0:
The trouble with the current implementation is that the trigger for auto-complete relies on `textarea-keydown` rather than the location of the caret. There are 4 kinds of auto-complete, we'll just go with `[[` for now. See end of this comment for more background on others.

[There are two triggers to open link auto-complete.](https://sourcegraph.com/github.com/athensresearch/athens/-/blob/src/cljs/athens/views/blocks/textarea_keydown.cljs?subtree=true#L558-602)

Trigger 1: no selected text. Example problem: If user moves left or right when auto-complete is open, auto-complete closes. Auto-complete should always be open if caret is within [[]], e.g. [[_x_y_z_]] where `_` is any caret position.

Trigger 2: selected text: Example problem: If user selects text than presses [[, update-query searches with that query, e.g. [[asd]] searches for pages with "asd". Pressing backspace should empty the query to "". Instead, it currently searches for "as". That's because `update-query` is updated by either `handle-backspace` or `write-char`. `update-query` should really only be updated when auto-complete is open, and should derive directly from the text. Right now, it is detected character-by-character.

**more background on auto-complete and `update-query`**

### State

[...]

PR #1219, opened 2021-05-23
fix(auto-complete): pressing [[,(( and enter works better. close #1214, #1220

### LINKED ISSUE #1214: Block search after wrapping `((` around selection is buggy
### Discussed in https://github.com/athensresearch/athens/discussions/1204

<div type='discussions-op-text'>

<sup>Originally posted by **datokrat** May 21, 2021</sup>
# Problem

Select some text, type `((`. This opens the block search panel. Select a block. One would expect that this results in a block reference to the selected block. However, in BETA-80, nothing happens (except that some characters are unselected), while in BETA-82, instead of a block reference, we get just `((`.

As an aside (which should probably get a separate bug report), after selecting text and then typing `((`, sometimes no item is selected at all, so that pressing ENTER leads to an error message that the selected index is still `nil`. It seems to be reproducible in the following way: Create a new block and then do the above as the first thing you do. I suspect that there is some problem with the local state of the block. After re-opening the page, this state is flushed and the bug occurs at pre-existing nodes again.

I believe the former bug to be a bit more severe.

# Demo

Here is a recording demonstrating the two bugs (first the former, then the latter) in the [github.io instance](https://athensresearch.github.io/athens/).

[https://www.loom.com/share/97e1ab3b9d8c40348b5ae1b661a0aa84](https://www.loom.com/share/97e1ab3b9d8c40348b5ae1b661a0aa84)

# Athens version and environment

Tested mainly in the browser (BETA-80 and BETA-82), but this also occured to me in my native BETA-80 running on Linux.

# A side remark

My main use case for "write text, select text, type `((`" stems from the performance problems with the block search (see #756). I noticed that searching lots of blocks is only slow when the resulting list of blocks is very long. However, when I write a very specific search query and then surround it with `((`, the search is very fast, even in older versions of Athens. This puzzles me a bit, since the #756 performance problems seem to be fixed by optimizing the search queries. I'm puzzled why the very specific search queries have been blazing fast even before this fix.

# Fix

I have a [fix](https://github.com/datokrat/athens/tree/fix-autocomplete) for the first bug, which also contains some refactoring of `auto-complete-inline`. Since this isn't an issue yet, I did not create a PR. </div>

datokrat:
See #1214.

Already editing `auto-complete-inline`, I noticed that the arity-overloaded contains duplicated code, one for each signature. The duplication was removed.

There is more code duplication in the arity-overloaded methods of the same file, which I can refactor, too, if this is wanted.

(Edit: I replaced "multimethod" by "arity-overloaded method")


tangjeff0:
Hi @datokrat, I love how this PR is largely documentation and refactor of existing code.

A small correction: [multimethod](http://clojure-doc.org/articles/language/polymorphism.html#extending-protocols-for-core-clojure-data-types) and [arity-overloading](https://riptutorial.com/clojure/example/11372/parameters-and-arity) are conceptually similar, but in Clojure there are in fact two separate constructs.

datokrat:
I added a bit more analogous refactoring and fixed a minor bug I found when I stumbled upon a discrepancy of the two implementations of `auto-complete-slash` I wanted to merge. The bug was that when `Block Embed` was selected using mouse, no block search panel would appear.

datokrat:
I noticed that there are some more places in `textarea_keydown` where undo might fail. I'll update them, too. Until then, I make this a draft.

datokrat:
The `deps-check` was fixed in #1220, I hope this is fine :)

tangjeff0 (changes requested):
Hey @datokrat I'm not sure this works. Am I demo'ing this right? https://www.loom.com/share/fe8950585d6f4650ba493e1c576ff0c0

datokrat:
Yes, that's right. I'm sorry. I did some further refactoring an forgot to test this edge case.

So, just for protocol, these cases should work now, I tried it again:

- hashtag completion
- ``[[`` and ``((`` completion without selected text
- ``[[`` and ``((`` completion with selected text
- `/` completion
- undo in all of these cases.

datokrat:
I merged PR #1234 into this one. Selection and autocompletion both work now:

https://www.loom.com/share/7cd8671b3d2b4cbb8d29ac0ac78d04f4

What's left is another update of the re-frame-10x dependency from 1.1.0 to 1.1.1 :rocket: I'm currently doing a full rebuild to ensure the version bumb doesn't break anything.

tangjeff0:
Don't worry about versioning. Clojure deps never breaks and it's a patch!

tangjeff0 (approved):
You are a refactor beast! 

Deleted two `println`s for you.

Next up, if you want to continue in this area, is fixing `update-query` so it's defined on caret position rather than keybindings #607

PR #1229, opened 2021-05-24
Find in page like chrome

### LINKED ISSUE #813: feat(cmd-f): electron doesn't have this by default
https://github.com/vector-im/element-web/issues/2764
https://github.com/electron/electron/issues/2037

arkRedM:
I was studying some notes in Athens and was surprised CMD F doesn't work by default and hence pr

Steps

- [x] Checked libs -> https://github.com/rhysd/electron-in-page-search not ideal to use(old and not cljs)
- [x] Using webContents from electron API https://www.electronjs.org/docs/api/web-contents
- [x] Search with cmd/ctrl f and escape to stop highlight
- [x] Refactor to new window
- [x] Next, prev
- [x] Theme, style - cleanup.
- [x] Simplifying code (self review)
- [x] Extra review comments

close #813


sid597:
Is this same as https://github.com/athensresearch/athens/issues/813#issuecomment-846935572

datokrat:
> * [ ]  Next, prev

I think this will be very hard. It seems that refocusing the search panel after search also resets the search, so that the next run of `findInPage` yields tha first match again.

[This](https://github.com/TheoXiong/electron-find) package uses a very dirty hack: The search panel is set `visibility: hidden` before `findInPage` and `visibility: visible` on `found-in-page`. However, this solution has lots of glitches. For example, the search panel flickers intensely during typing a query.

Another solution would be to put all the searchable content (everything except app toolbar?) into a separate `<webview>` component and run `findInPage` on this `webview` instead of the total app. Since the search panel is not part of the `webview`, next and prev should work more easily.

datokrat:
My Athens just warned me that the Electron `remote` module ist deprecated and that one should use this one instead: https://github.com/electron/remote

arkRedM:
> However, this solution has lots of glitches

Creating a new window for search and handling it with events. Similar to chrome.
> all the searchable content (everything except app toolbar?) into a separate <webview>

"find in page" is the only text. It should be ok(unless it looks extremely bad). I would probably want it everywhere. Settings, all pages, etc. 

>  warned me
Already changed in local.

Update: refactoring it to new window(new target) and managing with events. Using find inside this main dom is bad. Following behaviour of chrome(new window)

arkRedM:
Demo time: https://www.loom.com/share/0ff4294afa42456ab41b78036a8a4d5e 

Update: Found some edge cases. Might need more refactoring(simplifying some code). Please do comment on the demo in the meanwhile

datokrat:
Very nice!

This looks really good. The only problem I can see are the textareas in the background, so that you have to press Enter twice to move to the next block, but as far as I understand, #1217 will remove most textareas, so that this hopefully won't be a problem in the future.

arkRedM:
Final demo: https://www.loom.com/share/c8d207a9e1014814b8419e4cb3a46058

tangjeff0:
@shanberg please review

tangjeff0 (changes requested):
Code looks great. This is what I get on Linux:

image

Mac:

Screen Shot 2021-05-26 at 5 13 36 PM

shanberg (changes requested):

src/cljs/athens/views/find_in_page.cljs
```diff
@@ -0,0 +1,196 @@
...
+
+
+(defn get-find-win-pos
+  ([] (get-find-win-pos
+        (window-id->window @!main-window-id)))
+  ([win]
+   (let [[x y] (-> win (. getPosition) js->clj)]
+     [(+ x 20)
+      ;; + y 45(below toolbar) 25(title bar)
+      (+ y 45 (when-not (. win isFullScreen) 25))])))
```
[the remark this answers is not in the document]
arkRedM: nice catch

resources/public/find-in-page.html
```diff
@@ -0,0 +1,17 @@
...
+    <title>Find in page window</title>
+</head>
+<body>
+<div id="app"></div>
+<script>
+      src = "js/compiled/find-in-page.js"
+      var script = document.createElement("script");
+      script.setAttribute("src", src);
+      document.getElementsByTagName("head")[0].appendChild(script);
+</script>
```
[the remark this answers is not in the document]
arkRedM: implicit blindness to all files except .cljs 😅

arkRedM:
@tangjeff0 
> This is what I get on Linux:

why is the window empty? I think you might have started the dev earlier and directly checked out to the branch. There's a new target for find-in-page.js. It needs a lein dev restart

src/cljs/athens/views/find_in_page.cljs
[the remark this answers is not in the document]
shanberg: Or we could get the bounding box of the actual app_toolbar element and use that to place the window.

project.clj
```diff
@@ -50,7 +50,7 @@
                              :linux   "xdg-open"}}}
 
   :aliases {"dev"          ["with-profile" "dev" "do"
-                            ["run" "-m" "shadow.cljs.devtools.cli" "watch" "main" "renderer"]]
+                            ["run" "-m" "shadow.cljs.devtools.cli" "watch" "main" "renderer" "find-in-page"]]
             "compile"        ["with-profile" "dev" "do"
```
tangjeff0, line 54: @arkRedM I see why my build was failing. The reason is I use `yarn dev` more often than `lein dev`. Having two shadow-cljs wrappers from yarn and lein is confusing and may lead to errors. They do subtlely different things, though, I think. May be good to address this issue in a later PR.

If this is only in the `dev` build, we need to make sure the new `:find-in-page` is also released with the production build.

src/cljs/athens/style.cljs
```diff
@@ -1,59 +1,13 @@
...
-   :background-minus-2  "#EFEDEB"
-   :graph-control-bg    "#f9f9f9"
-   :graph-control-color "black"
-   :graph-node-normal   "#909090"
-   :graph-node-hlt      "#0075E1"
-   :graph-link-normal   "#cfcfcf"
-
-   :error-color         "#fd5243"})
-
-
```
tangjeff0, line 56: Do we like having a separate theme file @shanberg ?

src/cljs/athens/views/find_in_page.cljs
```diff
@@ -0,0 +1,243 @@
...
+        {:clear-current-selection _}))))
+
+
+(rf/reg-event-fx
+  :next-result
+  (fn [_ _] (find-in-page!)))
+
+
+(rf/reg-fx
+  :prev-result
```
tangjeff0, line 91: Can you namspace these events?

src/cljs/find_in_page/core.cljs
```diff
@@ -0,0 +1,229 @@
...
+    {:db       (update db :find-in-page-info
+                       merge (select-keys
+                               init-data [:index-info :is-dark?]))
+     :dispatch [:set-find-in-page-text text true]}))
+
+
+(rf/reg-event-db
+  :set-find-in-page-index
+  (fn [db [_ index-info]]
+    (assoc-in db [:find-in-page-info :index] index-info)))
```
tangjeff0, line 111: Namespace events?

src/cljs/athens/views/find_in_page.cljs
```diff
@@ -0,0 +1,243 @@
...
+         [_ y-content-size] (-> win (. getContentSize) js->clj)]
+     [(+ x 20)
+      ;; + y 45(below toolbar) 25(title bar)
+      (+ y 45 (when-not (. win isFullScreen)
+                (- y-win-size y-content-size)))])))
+
+
+(defn init!
+  []
+  (let [browser-win (.. electron -remote -BrowserWindow)
```
tangjeff0, line 147: Are there any solutions that don't create a new BrowserWindow?

I get that this solution let's us use native `electron/webContents` lib, but it introduces some cruft: another file for keybinding listeners, another html file with font imports, IPC usage, another shadow-cljs build target (see separate comment on this).

A separate window may also introduce more surface area for OS-specific edge cases.

For example, I use a tiling manager.

image

image

src/cljs/athens/theme.cljs
```diff
@@ -0,0 +1,60 @@
...
+   :graph-node-hlt                   "#0075E1"
+   :graph-link-normal                "#cfcfcf"
+
+   :error-color                      "#fd5243"
+
+   ;; find in page styles
+   :find-in-page-background          "rgb(245 245 245 / 75%)"
+   :find-in-page-text-color          "black"
+   :find-in-page-icon-hover-bg-color "rgb(67 63 56 / 10%)"
+   :find-in-page-text-color-light    "#3e3e3e"})
```
tangjeff0, line 60: Anyway to reuse more existing styles rather than creating new ones? @shanberg

tangjeff0 (changes requested):

src/cljs/athens/style.cljs
arkRedM: One main thing is keeping the find-in-page compiled file small. Hence separated it out. Separating it out to a file seemed like a decent abstraction

src/cljs/athens/views/find_in_page.cljs
arkRedM: Tiling manager is able to control the find in page window? Is it not changing with position of the parent as mac does? is it not changing when window resized?


Webcontents api doesnt have the exclusion of what not to highlight so each highlight will also highlight the text field. You would have to deal with refocusing logic on input and could have more issues.

It's basically mimics chrome window

project.clj
arkRedM: Need to look into what exactly is diff between yarn and lein dev builds. But yeah, should add it to prod.

arkRedM:
Update: fixed reload-related edge cases, compile and prod targets, namespace events

src/cljs/athens/style.cljs
shanberg: - How meaningful is the size difference?
- Do we get any reduction/minification for free with tree-shaking?

All else being equal, it’s easier to manage the styles when they're all in the same place, so I'm trying to understand the tradeoff.

arkRedM: > How meaningful is the size difference?

~35MB vs ~12MB. It makes sense  because config, deps and all the stuff that's needed from athens style will get compiled. 

> tree shaking

style uses all these deps one way or another. Afaiu this won't help

src/cljs/athens/views/find_in_page.cljs
arkRedM: Need more info on this.

src/cljs/athens/theme.cljs
arkRedM: Like changing the existing color opacities? 
I've tried existing color color-minus-n and plus-n, didn't look great. 
Made these with trail and er

src/cljs/athens/views/find_in_page.cljs
tangjeff0: Is this another valid solution? https://discord.com/channels/708122962422792194/708124156113321985/847631235320905738

arkRedM: > Is this another valid solution? https://github.com/glafche/electron-find-in-page

There were already other deps that did this? This is not really different.
https://github.com/rhysd/electron-in-page-search did similar(more robust)
All these are literal wrappers on webContents API and provide a window(that's it). Shallow modules and not much value. Constrained API for the wrapper itself. Why not depend on the robust browser API of webContents and avoid these unnecessary deps?

> What is wrong with highlight.js?

it modifies elements in place. Will have unintended conseq on rendering logic.
Way too slow if there are too many matches.

> Sometimes the findWindow does not update position with parent.

Maybe old code, if reloaded main window loses findWindow that's already open. 0e3bfa8 commit prevents reloads in such cases. If it's happening after than commit, need more info.

> Tiling manager

I don't see it happening in my mac. I don't have access to a ubuntu machine. Might need some help here. 
I doubt adding an npm dep will fix this because it's just a window at the end.

arkRedM: > Creating a separate window is not mimicking Chrome in a pretty big way?

This is probably the best electron offers. Not sure if there are any other ways for "attached window" to be present in electron. Functionality is the one that's getting mimicked. What do you think is limiting in terms of functionality?

src/cljs/athens/style.cljs
arkRedM: Any more suggestions here? mark it resolved?

project.clj
arkRedM: Looks like it got fixed in the recent merge https://github.com/athensresearch/athens/pull/1246.  
mark it resolved?

src/cljs/athens/views/find_in_page.cljs
tangjeff0: Is it possible to back up and understand the tradeoffs between using webContents API versus creating an element within the original window?

I think this is a good architectural question that merits a design document. There are both engineering and product considerations between the two approaches (maybe there are even more than 2 approaches).

Is it possible to back up and understand the tradeoffs between using webContents API versus creating an element within the original window?

I think this is a good architectural question that merits a design document. There are both engineering and product considerations between the two approaches (maybe there are even more than 2 approaches).

Obsidian seems to create DOM elements directly with their find-in-page. They are probably not the only ones doing this so maybe we can learn from them.

tangjeff0: Is it possible to back up and understand the tradeoffs between using webContents API versus creating an element within the original window?

I think this is a good architectural question that merits a design document. There are both engineering and product considerations between the two approaches (maybe there are even more than 2 approaches).

Is it possible to back up and understand the tradeoffs between using webContents API versus creating an element within the original window?

I think this is a good architectural question that merits a design document. There are both engineering and product considerations between the two approaches (maybe there are even more than 2 approaches).

Obsidian seems to create DOM elements directly with their find-in-page. They are probably not the only ones doing this so maybe we can learn from them.

arkRedM: 41ab9e0

tangjeff0: Is it possible to back up and understand the tradeoffs between using webContents API versus creating an element within the original window?

I think this is a good architectural question that merits a design document. There are both engineering and product considerations between the two approaches (maybe there are even more than 2 approaches).

Is it possible to back up and understand the tradeoffs between using webContents API versus creating an element within the original window?

I think this is a good architectural question that merits a design document. There are both engineering and product considerations between the two approaches (maybe there are even more than 2 approaches).

Obsidian seems to create DOM elements directly with their find-in-page. They are probably not the only ones doing this so maybe we can learn from them.

arkRedM: tangjeff0-bot?

tangjeff0: lol bad internet

arkRedM: Let me know if the reasoning in adr sounds ok. That's one blocker along with the tiling issue. Let's debug that if possible(I don't have access to an ubuntu machine)

arkRedM: Tiling is reproducible in mac. Explored modal views(basically there's no good customization to it in electron -- can't even change the backdrop or position).

added code to force find window back to the position but, that doesn't stop tiling manager to create space.
https://www.loom.com/share/cad9928b62394f1bb713f1714a429392 

I think this is the best possible with electron + window. It will server most users(not sure how many use a tiling manager). Others alternatives are highlighted in adr.

shanberg (commented):
On Windows and Ubuntu a scrollbar appears as soon as I enter text to begin searching: 
CleanShot 2021-06-02 at 21 10 47@2x

And the Ubuntu instance seems to be missing some spacing on the right. It also appears offset to the left slightly, relative to the Windows version:
CleanShot 2021-06-02 at 21 18 41@2x

I'm surprised that the find-on-page window isn't automatically closed when I navigate to a different page. I can imagine situations where that's useful, but it’s surprising behavior. If we do want that behavior rather than something typical of a web browser, I think that should be a separate UX discussion.

I'm not sure what's to be done about the shadows and corner radius and such. There may be some trickery we can do with transparent Electron window backgrounds and additional styling, but that strikes me as a separate unit of work.

tangjeff0:
Do you have any deeper concerns about separate window vs same window for search? @shanberg @neotyk

shanberg:
> Do you have any deeper concerns about separate window vs same window for search? @shanberg @neotyk

Same window is easier for styling to deal with but if there's no technical pathway to achieving that, then I think we can make do with a separate window.

neotyk:
> Do you have any deeper concerns about separate window vs same window for search? @shanberg @neotyk

Yes I do have concerns with separate window, as tiling window manager user - apps that spawn new windows for popup kind of content, are broken.

@arkRedM Great to see you've put ADR for it 👍 
I think we're missing one available option to us, that is to leverage the work on Athena search.
We could run fast search on content that we know we display, store result coordinates in re-frame, render `span.text-search-result` for each result.

As for Search input, have we considered maybe just text-input field, styled nicely and fixed position or something.
I see window as big overhead compared to normal re-frame component.

What do you think of this approach of leveraging Athena Search and re-frame component?

arkRedM:
> We could run fast search on content that we know we display, store result coordinates in re-frame, render span.text-search-result for each result.

I think the use case is different. Athena is the whole search and find-in-page is what is currently on-screen search(works on all pages -- useful in more than just block search -- page list page/settings etc). Content and navigating through it is more important in case of find in page, so having a big input blocking the content like Athena is bad. Also if we are doing the highlighting(highlight.js) it's slower(more in adr)

> As for Search input, have we considered maybe just text-input field, styled nicely and fixed position or something.
I see window as big overhead compared to normal re-frame component.

As I've said in adr, if the input is inside the same window, webContents.findInPage API (browser's find api) will also effect the input we are typing in as well and more issues with that approach are in ADR.

tangjeff0:
Is it possible to continue using `findInPage` but with `ipcRenderer` so it's on the `renderer` thread, not `main`?

neotyk:
I didn't mean to reuse Athena for doing in-page search.
What I meant was to reuse the search functionality of it, and just take it to the next level by allowing it to search on narrower dataset.

I agree that search query getting highlighted as search result is meh.

We could achieve same async behaviour by using WebWorkers, though Athena is not using it and it's searching whole db.

Issue #1213, opened 2021-05-22
Clicking inside a block with a hyperlink does not work sometimes


sid597:
https://www.loom.com/share/e5bd2f75589249079593bf0d131dc670

sid597:
I would like to work on this issue 
- [ ] Other cases where this might arise
- [ ] What is causing the issue 
- [ ] How can this be solved

sid597:
This is what I could gather by just using the chromium dev tools, have not looked at the code : 

First, let's check out how a block with normal text in HTML looks like :
There is some more HTML under the `<div class="block-container"`  but that is not of importance, here focus on `<span class="block">`.

![s1](https://user-images.githubusercontent.com/22994594/119244008-bbc38a00-bb89-11eb-8004-160131f041d8.png)




Now here is the same for a buggy block, focus on  `<span class="block">` :

![s2](https://user-images.githubusercontent.com/22994594/119244037-07763380-bb8a-11eb-8d46-874a3cf47ce6.png)


As you can see this `<span class="block">` has another nested `span` which contains both the block text i.e `A block with link in it is like this ` and the hyperlink. 

From all the various combinations I tried by inserting bold text, italics text, block embeds etc. in a block, I found whenever we have the case of `<span>` containing both the text and hyperlink **and** the hyperlinks `<span>` is not nested in another tag this bug will arise, checkout the video for a few cases.
https://www.loom.com/share/3ddc825e0fd4404ebc2f0a870158565b

tangjeff0:
> I found whenever we have the case of <span> containing both the text and hyperlink and the hyperlinks <span> is not nested in another tag this bug will arise, checkout the video for a few cases.

This is very promising if this is reproducibly true. What I don't understand is why clicking a few times eventually makes it possible to click in to the textarea.

sid597:
> What I don't understand is why clicking a few times eventually makes it possible to click into the textarea.
Yeah me too, I am looking into it.

sid597:
I think there are 2 ways to solve this issue and I am not sure which one would be the right way to go because I am currently not fully aware of the consequences they might have. I will post both the approaches I have: 

1. Make some `z-index` changes : 
     - The current implementation has `z-index`  for ( this is all for parent `div` with  `class = "block-content"`) :
         - `textarea`  : not set i.e 0
         - inside a `span` : all the `span` and `<a>`  have `z-index` `2` 
[...]

PR #1236, opened 2021-05-26
Fix(click): Clicking inside a block with hyperlink works on first attempt

sid597:
PR  for #1213

### DIFF
--- a/test/athens/parser/compatibility_test.clj
+++ b/test/athens/parser/compatibility_test.clj
@@ -157,7 +157,7 @@
 ;; as a link. Those mostly test the URL parser.
 (deftest parser-lone-raw-url-tests
   (are [url] (= [:block [:paragraph [:textRun [:link {:text   url
-                                                   :target url}]]]] (sut/staged-parser->ast url))
+                                                      :target url}]]]] (sut/staged-parser->ast url))
     "https://example.com"
     ;; URL with path set to /.
     "https://example.com/"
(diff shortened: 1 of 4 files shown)


sid597:
There is a bug in the current implementation due to which block-ref styles are not applied. This PR is more of a draft to show my current approach.

sid597:
Fixed the issue by using pointer-events, tested on a few custom blocks which contain all type of elements.

test/athens/parser/compatibility_test.clj
```diff
@@ -269,7 +269,7 @@
 
     [:block
      [:paragraph
-      [:span
+      [:testRun
```
neotyk, line 272: it should be `textRun`

neotyk (changes requested):
Using CamelCase is not very idiomatic Clojure, please use kebab-case for var, fn, keyword names.
In this particular instance `:textRun` -> `:text-run`.

Apart from it LGTM, Thank you :-)

tangjeff0 (changes requested):
see Alex's requests

test/athens/parser/compatibility_test.clj
sid597: Aha!, Thank you, Alex.

sid597:
Thank you Alex for reviewing this, more PR's coming your way from my side :)

tangjeff0 (approved):

PR #1242, opened 2021-05-27
perf(right-sidebar): fix memory+time leak with proper GC of sorted-map #1239

datokrat:
Bug discussed in #1239 

Prevent expensive and unnecessary recursive calls to previous sidebar item maps

### DIFF
--- a/src/cljs/athens/events.cljs
+++ b/src/cljs/athens/events.cljs
@@ -258,14 +258,18 @@
   (fn [{:keys [db]} [_ uid is-graph?]]
     (let [block     (d/pull @db/dsdb '[:node/title :block/string] [:block/uid uid])
           new-item  (merge block {:open true :index -1 :is-graph? is-graph?})
-          new-items (assoc (:right-sidebar/items db) uid new-item)
+          ;; Avoid a memory leak by forgetting the comparison function
+          ;; that is stored in the sorted map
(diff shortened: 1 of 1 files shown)


tangjeff0:
I don't understand the memory leak :|

datokrat:
Well yes, the explanation was a bit convoluted. I hope a picture will help.

image

So my claim is that old values of `(:right-sidebar/items db)` will never be disposed by the Garbage Collector. Over time (when lots of pages are added and removed in the sidebar), memory usage will steadily increase.

However, I think that the time leak is the greater (but connected) problem: Whenever a new page is opened, some sidebar operations (such as closing a page) become exponentially slower. The comparison function calls the comparison function of the previous sidebar state twice. The latter, in turn, calls the previous comparison function twice. This leads to exponential processing time (comparable to a very naive recursive implementation of Fibonacci numbers).

I'm not sure whether this helps clarifying the topic. Otherwise, we may also meet in Discord, maybe tomorrow? Of course, this could wait if there are higher-priority tasks at the moment.

tangjeff0:
Hi @datokrat, actually your earlier video posted here https://github.com/athensresearch/athens/discussions/1239#discussioncomment-791038 actually was the best explanation, because it used video + had print statements.

You are very good at teaching :)

PR #1244, opened 2021-05-28
chore: downgrade re-frame-10x so web version works, comment out deps-check

tangjeff0:
Issue with re-frame-10x on web :app build

Screenshot:
image

Screenshot on local (which has more error logs):
image

[Brought up by @shanberg ](https://discord.com/channels/708122962422792194/802652338619285594/847647941623676958)

The error comes from re-highlight. @neotyk updated a lot of deps including re-frame-10x at https://github.com/athensresearch/athens/commit/181ad52286d982586a9c1f5016d37553915b6b05

Version goes from 0.6.0 to 1.0.2

re-frame-10x changelog switches to re-highlight at 1.0.2 https://github.com/day8/re-frame-10x/blob/master/CHANGELOG.md#102---2021-03-23

Commenting out `deps-check` for now so that athensresearch.github.io/athens works again (and embedded demo on athensresarch.org)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,16 +59,17 @@ jobs:
        - name: Lint
          run: script/lint
 
-   deps-check:
-     runs-on: ubuntu-18.04
-     steps:
-       - name: Git checkout
-         uses: actions/checkout@v1
-         with:
-           fetch-depth: 1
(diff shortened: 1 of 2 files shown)


tangjeff0:
Doesn't seem to work. Weird

PR #1251, opened 2021-05-30
fix(cursor pos, undo, #1250) on shift+enter, cmd+enter, pair character removal

### LINKED ISSUE #1250: Adding a todo doesn't place the caret at the end of the component
**Problem**
Creating a todo `cmd+enter`, places the carent at the start of the block (before the {{[[TODO]]}}) which requires to move the caret at the end of the block to start writting

**Screenshots/Demo**
https://www.loom.com/share/9cddbc24384e457485829ef8659cf958

**Athens Version**
- Beta 85. Was not happening in 82 (I think)

datokrat:
This can be seen as an amendment to #1219, since I seem to have overlooked some occurrences of the evil `(swap! state assoc :string/local new-value)`.

This fixes #1250 on my machine.

### DIFF
--- a/src/cljs/athens/views/blocks/textarea_keydown.cljs
+++ b/src/cljs/athens/views/blocks/textarea_keydown.cljs
@@ -392,7 +392,7 @@
 
 (defn handle-enter
   [e uid state]
-  (let [{:keys [shift ctrl meta value] :as d-key-down} (destruct-key-down e)
+  (let [{:keys [shift ctrl meta value start] :as d-key-down} (destruct-key-down e)
         {:search/keys [type]} @state]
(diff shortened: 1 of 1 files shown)


julionav:
Code looks good, this solves the issue, but seems to introduce another: 

https://www.loom.com/share/f01c41df62d3443f872244b98d1d2e79

datokrat:
> Code looks good, this solves the issue, but seems to introduce another:

Oh no. :facepalm:

I added a line that moves the cursor to the end of the block.

However, I can't reproduce locally why the style check fails...

tangjeff0 (approved):

PR #1256, opened 2021-05-30
Refactor keybindings to use a library and introduce a POC of keybindings settings

julionav:
Solution to athensresearch#625

Previously on: https://github.com/julionav/athens/pull/1

- **Very short version:**
    - Most of the keybindings event handlers that were checking manually for keys were replaced by Mousetrap. This is used in a react component to "listen" events locally, and also for global events.
    - Not all the events listener were replaced, some of them were not possible, nor had much sense, like the event to listen for "queries" with the slash command.
    - Some hotkeys were added in a keymap that lives in `re-frame`, when the keymap changes in reframe, the events are rebinded. This allows to change hotkeys at runtime.
    - Since the the hotkeys can be changed at runtime, an UI was built to other developers to change them. **THIS IS NOT MEANT TO BE USED BY END USERS**. To do so, run `toggleUnstableKeymapSettings`. Changes in the keymap are persisted in localstorage, and also whether the unstable feature is active or not.
    - Settings UI: https://drive.google.com/file/d/12wu4lTa7MbaS5qf3oU60vCrmZ_5NJsSC/edit
- **Changes summary:**
    - Installed [mousetrap](https://craig.is/killing/mice)
    - Added a keymap (clojure map) (hotkey aliases and hotkeys pairs) that allow to bind event handlers to hotkeys using an alias (like :athena/toggle)
    - Created a couple of helper functions
        - mousetrap-bind to bind a hotkey (or a hotkey alias) to an event.
        - mousetrap-bind-all to bind a map of hotkeys (or aliases) with events.
    - Created a react component that creates a mousetrap instance, and listen for events made in the childs passed to the react component.
        - This react component will rebind the events in the case that the keymap changes by connecting with `re-frame`.
        - It also cleans up the event listeners when it unmounts.
    - Global keybindings:
        - Added "changeable" global keybindings (hotkey + event handler pair) that are re-binded when the keymap changes.
            - These are attached to the browser `window` in the `boot/desktop` and `boot/web` events.
                - in the `boot/desktop` it is added to the `async-flow`
        - Added "unchangeable" global keybindings.
            - These are attached to the browser`window` object when the app starts using an exposed `init` function.
        - "changeable" and "unchangeable" global keybindings are pretty much the same, the main difference is that the "changeable" keybindings are re-binded if the keymap changes (using a re-frame effect).
        - 
    - Changes to the keymap are enabled using re-frame events.
        - There is an event to update the keymap.
            - This returns an effect that saves the update in localstorage. This only saves the hotkey that was changed, not the entire thing. This allows to keep expanding the default keymap without having issues with user-stored keymaps.
        - There is an event to restore a keymap.
            - This sets the hotkey to the default, and removes the saved value from localstorage.
    - Athena keybinds were replaced.
        - It was also slightly refactored, extracting the functionality to "listen" for clicks out side of it to another component. This removed the need of using a reagent component class.
        - the "create-page" event (when pressing enter, or shift-enter) was moved to a re-frame event. This follows the rule of dispatching one event per user event (I read this somewhere I cannot remember, from which I didn't take notes like a loser)
    - Block content element keybindings were replaced.
        - arrow key event handler was split in 3 handlers: one when pressing shift, another when pressing the mod key, and another when there isn't a modifier. 

### Tasks
- ☑︎ Review current keybindings
    - List use cases
- ☑︎ Review available libraries
    - Pick one that allows to change keybindings at runtime (or allow to easily do it) so that we can eventually allow the user to change their keybindings.
    - Libs:
        - https://dmauro.github.io/Keypress/
        - https://github.com/jaywcjlove/hotkeys
        - https://github.com/piranha/keymage
        - https://craig.is/killing/mice
        - CLJS/re-frame: https://github.com/gadfly361/re-pressed
- ☑︎ What do we want to be able to do?
    - Cross platform keybindings
    - Global listeners
    - Local (component level) listeners
    - Reset the binding for a listener somehow (unbinding and binding a listener is good enough)
    - Record user keypresses so that we can update the keysbindings of a listener.
- ☑︎ Decide library for POC
    - Decided for `react-hotkeys` after playing with `hotkeys` and `mousetrap`
- ☑︎ Replace current implementation with `react-hotkeys`
    - ☑︎ Global listeners `core.listeners.cljs`
    - ☑︎ Replace local listeners on Athena
        - While working on this I saw that Athena used a class component just to handle the close on "click" outside of it; so I went ahead and extracted that functionality into a component that can be composed in Athena render function. This meant that the **class component** was no longer needed, so I stripped that part.
        - ☑︎ Clean up event handling on athena
    - ☑︎ Replace local listeners on `block-content-el`
        - ☑︎ Arrow keys
            - ☑︎ Replace mousetrap implementation to use a reagent react class
            - Allows to navigate through blocks
            - ☑︎ Replaced
            - ☑︎ Tested
            - ☑︎ Retested
        - ☑︎ Tab event
            - Indents a block
            - ☑︎ Replaced
            - ☑︎ Tested
                - ☑︎ Retested
        - ☑︎ Enter event 
            - Creates a new block after, with `shift`, it creates a line break within the same block.
            - ☑︎ Replaced
            - ☑︎ Tested
            - ☑︎ Retested
        - ☑︎ Backspace event
            - ☑︎ Replaced
            - ☑︎ Tested
            - ☑︎ Retested 
        - ☑︎ Delete event
            - "Delete has the same behavior as pressing backspace on the next block."
            - ☑︎ Replaced
            - ☑︎ Tested
            - ☑︎ Retested
        - ☑︎ Escape event
            - Stops the edition of a block
            - ☑︎ Replaced
            - ☑︎ Tested
            - ☑︎ Retested
        - ☑︎ Slash command handling /
            - ☑︎ Tested that it didn't break
            - **May not be replaceable**
            - **This cannot be replaced since it listens to any key, and moustrap cannot do that, so this was left as it it (in a keydown event handler)**
        - ☑︎ Shortcuts
            - **Issues**
                - ☑︎ After writing some text, the bindings stop working.
                    - Solved by migrating to `mousetrap`
            - ☑︎ MOD + A
                - Selects all text
                - ☑︎ Replaced
                - ☑︎ Tested
                - ☑︎ Retested
            - ☑︎ MOD + Z
                - Does a datascript UNDO if there aren't more `undos` to do in the textarea.
                - ☑︎ Replaced
                - ☑︎ Tested
                - ☑︎ Retested
            - ☑︎ MOD + B
                - Surrounds text in **double * to make it bold**
                - ☑︎ Replaced
                - ☑︎ Tested
                - ☑︎ **Retested**
            - ☑︎ MOD + I
                - Surrounds text in *a single * to make it italic*
                - ☑︎ Replaced
                - ☑︎ *Tested*
                - ☑︎ *Retested*
            - ☑︎ MOD + Y
                - Surrounds text in a ~~double ~ to add a strikethrough~~
                - ☑︎ Replaced
                - ☑︎ Tested
                - ☑︎ ~~Retested~~
            - ☑︎ MOD + U
                - Surounds text in doble -- to do **nothing I guess**
                - ☑︎ Replaced
                - ☑︎ --Tested--
                - ☑︎ --Retested--
            - ☑︎ MOD + H
                - Surounds text in ^^double ^ to highlight^^
                - ☑︎ Replaced
                - ☑︎ Tested
                - ☑︎ ^^Retested^^
            - ☑︎ MOD + O
                - ;; if caret within brackets or brackets, navigate to that page
                - ;; if caret within ((uid)), navigate to that uid
                - ;; otherwise zoom into current block
                - ;; if caret on a #hashtag, navigate to that page
                - ☑︎ Replaced
                - ☑︎ Tested
                - ☑︎ Retested
    - ☑︎ Replace keybindings from multi-block-selection
        - ☑︎ Enter
        - ☑︎ Delete/backspace
        - ☑︎ Tab / Shift + Tab 
        - ☑︎ Shift + up or down
        - ☑︎ up or down
    - ☑︎ Allow to change keybindings at runtime
        - My plan here is relatively simple; allow the bindings to be set with literal bindings (strings, like "mod+b"), or using keywords (like :toggle-athena), when a keyword is used, it is looked up in a map of keywords that live in re-frame that can be changed at runtime.
        - When a hotkey changes, the key bindings will react to the hotkey to it and re-bind the events. 
        - ☑︎ Use re-frame effect or a react component to set global bindings
        - ☑︎ Move the global bindings to a reagent component to allow them to react to re-frame hotkeys changes.
        - ☑︎ Update `mousetrap` helper functions to read from re-frame hotkeys config map if the binding is a keyword
        - ☑︎ block-content-el shortcuts
        - ☑︎ Global shortcuts
- ☑︎ Replace react-hotkeys for moustrap
    - ☑︎ Replace Athena bindings
    - ☑︎ Replace global bindings
    - ☑︎ Replace current block-content-el replacements with `react-hotkeys`
- ☑︎ Deploy app so that people can test this.
    - **Netlify seems to deploy clojure apps**
    - Build command: `lein run -m shadow.cljs.devtools.cli --npm release app`
    - Public directory: `/resources/public`
- ☐ Document process to add new keybinding in code 
- ☑︎ Fix review from [[Alex]]
    - ☑︎ Expose configuration of hotkeys in an init function
        - https://github.com/julionav/athens/pull/1/files#r637405966
    - ☑︎ Add tasks for rest of [[Alex]] comments
- ☑︎ Refactor `mousetrap` component with the help of [[Alex]]
    - **Update: ** I went ahead and refactored this on my own. Alex can check it on the code review
- ☑︎ Review https://github.com/athensresearch/athens/issues/1214
- ☑︎ Mousetrap component is not clearing up events on unmount.
- ☑︎ mousetrap component enchancements:
    - ☑︎ Look for a way to only reset the events when the bindings truly change. Right now they are reset on any keypress.
        - This may be caused by the functions on the callbacks. They are technically changing on each render. 
        - **A good solution here is to go with the `react-hotkeys` approach and include a prop "allowChanges" that is false by default, but when true it listens to do re-bindigns.**
        - **Bindings do not change when the callback changes. Only when the keymap change**
- ☑︎  Write a summary of the changes
- ☑︎ Test and fix Athena.
    - **Getting "index out of bound" on athena when going up**
- ☑︎ Persist keymap
    - Only persist changed hotkeys.
    - When restoring, remove the persistence of the hotkey.
    - **This allows to modify the default keymap in the future and have the users pick up the changes**
- ☑︎ The hotkey bindings configuration is all over the place. I need to consolidate it in a single place, if it is possible.
- ☑︎ The language is a bit confusing right now. I'm using too many terms keybinding, binding, keys, etc.
    - Definitions.
        - Hotkey: combination of keys `ctrl+k`
        - Hotkey alias: keyword that is mapped to a hotkey
        - Keymap: map between a hotkey aliases (`athena/toggle`) and a hotkey
        - Keybinding: the actual binding of the hotkey and the event handler.
- ☑︎  Build temporary UI to change keymap
    - ☑︎ Record user keys and update keymap
    - ☑︎ Allow to restore "default" hotkey in the keymap 
    - ☑︎ Hide new UI by default unless the user executes a function on the console: `_activateKeybindingsAlphaPreview()` or something like that.
- ☐ Get code review from [[Alex]] or [[Jeff]]
- ☐ Fix caret position saving when jumping between blocks using arrow keys.
- ☑︎ Retest bindings 
    - ☑︎ CMD+Enter binding doesn't work
        - ☑︎ Testing... IT WORKS
- **Open questions from a talk with [[Shanberg]]**
    - ☑︎ Restoring default hotkey
        - **Added!**
    - ☐ Platform specific keymaps (mac and Windows/linux)


src/cljs/athens/events.cljs
```diff
@@ -175,6 +177,17 @@
...
 
 
+(reg-event-fx
+  :athena/create-page
+  (fn [_ [_ title uid open-on-sidebar]]
+    {:fx [[:dispatch [:page/create title uid]]
+          [:dispatch [:athena/toggle]]
+          (if
+            open-on-sidebar
+            [:dispatch-later {:ms 300 :dispatch [:right-sidebar/open-item uid]}]
```
neotyk, line 187: Not a huge fan of artificial delays, what will happen when DB is big enough for it not to complete in 300ms?

src/cljs/athens/views.cljs
```diff
@@ -62,22 +62,22 @@
 
 (defn main
   []
-  (let [loading    (rf/subscribe [:loading?])
-        modal      (rf/subscribe [:modal])]
+  (let [loading (rf/subscribe [:loading?])
+        modal (rf/subscribe [:modal])
+        {:keys [msg type]} @(rf/subscribe [:db/snack-msg])]
```
neotyk, line 67: What's happening here?
Why do we need it with keybinding?

src/cljs/athens/views/athena.cljs
```diff
@@ -272,76 +229,121 @@
...
-                                                   [:span.link-leader (use-sub-style result-style :link-leader) [(r/adapt-react-class ArrowForward)]]])))]))]])))})))
+  (let [s (r/atom {:index   0
+                   :query   nil
+                   :results []})
+        search-handler (create-search-handler s)]
+    (fn []
+      (let [open? @(subscribe [:athena/open])]
+        (when open?
+          [view-utils/track-outside-click #(dispatch [:athena/toggle])
+           [athena-keybindings s
```
neotyk, line 297: Why keybindings get involved with state here?

src/cljs/athens/views/blocks/content.cljs
```diff
@@ -316,6 +317,39 @@
...
+  (let [event-wrapper
+        ;; Only handle event if the currently editing uid is the same
+        ;; as this block; and there aren't any selected item.
+        (fn [callback-or-config _]
+          (let [wrap-callback
+                (fn [callback]
+                  (fn [event]
+                    (when
+                      (and
+                        (= uid @(rf/subscribe [:editing/uid]))
```
neotyk, line 332: Please don't `deref` inside the helper like that.
Subscribe in `let` and so on.

src/cljs/athens/views/blocks/textarea_keydown.cljs
```diff
@@ -380,7 +387,7 @@
 
 (defn handle-escape
   "BUG: escape is fired 24 times for some reason."
```
neotyk, line 389: Is this still the case?

src/cljs/athens/views/utils.cljs
```diff
@@ -0,0 +1,23 @@
...
+    [reagent.core :as r]))
+
+
+;; Calls `callback` when a click is done outside the provided `children` nodes.
+;; Note: This won't reattach a document event listener if the callback changes.
+;; So you need to make sure you don't need that, and if you need that; this can
+;; be rewritten using an `useEffect` to listen/unlisten for clicks on `callback`
+;; changes.
+(defn track-outside-click
+  [callback children]
```
neotyk, line 13: Confusing naming, it's `child` not `children`

neotyk (changes requested):

src/cljs/athens/effects.cljs
```diff
@@ -387,3 +405,16 @@
...
+(def unbind-global-keybindings (atom nil))
+
+
+(reg-fx
+  :keybindings/bind!
+  (fn []
+    (when (not (nil? @unbind-global-keybindings))
+      (@unbind-global-keybindings))
+    (reset! unbind-global-keybindings
+            (bind-changeable-global-keybindings))))
```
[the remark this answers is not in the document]
julionav: `bind` returns a function to cleanup the `bindings`, so yeah, `bind` returns `unbind`

src/cljs/athens/events.cljs
julionav: I was unable to solve this though. How can I wait until `:page/create` finishes? I can look more into it though

src/cljs/athens/views.cljs
julionav: Not related to keybindings. You asked me in another PR to move the subscription from line 70 to the let bindings at the start of the component. This removes the need of the function on line 70, and it can be just hiccup

src/cljs/athens/views/athena.cljs
julionav: They need access to change the currently selected element in Athena. That is part of the current implementation

src/cljs/athens/views/blocks/textarea_keydown.cljs
julionav: It is not :+1:

src/cljs/athens/events.cljs
julionav: The solution here is to rework `:right-sidebar/open-item` to only update the `db`, and create a subscription that does the `datascript` pulling and element sorting. 

That way the subscription will be updated once the `:page/create` transactions is done. Thoughts?

julionav: I could do it here, or in another PR

neotyk:
This PR uncovers another problem, that needs fixing before we can revisit keybindings.

Whole shenanigans with using js library is related to a need of binding keys locally in respect to DOM.

Key bindings could be solved easily with use of https://github.com/piranha/keybind/blob/master/src/keybind/core.cljs
if we wouldn't need to have local DOM bindings, and we need them because of local ratom state.

Local ratoms seem to me like a bit of premature optimization, maybe we could optimeze shape of data that is stored in re-frame at the same time.

Basically having global state in re-frame will allow us to have global keybindings, because we'll be able to resolve any context of key binding callback.

neotyk:
I suggest to Close It 🛑

julionav:
> This PR uncovers another problem, that needs fixing before we can revisit keybindings.
> 
> Whole shenanigans with using js library is related to a need of binding keys locally in respect to DOM.
> 
> Key bindings could be solved easily with use of https://github.com/piranha/keybind/blob/master/src/keybind/core.cljs
> if we wouldn't need to have local DOM bindings, and we need them because of local ratom state.
> 
> Local ratoms seem to me like a bit of premature optimization, maybe we could optimeze shape of data that is stored in re-frame at the same time.
> 
> Basically having global state in re-frame will allow us to have global keybindings, because we'll be able to resolve any context of key binding callback.

Makes sense @neotyk. Lets revisit this in the future

src/cljs/athens/effects.cljs
```diff
@@ -279,6 +281,22 @@
...
+          new-value (assoc old-value key value)]
+      (js/localStorage.setItem storage-key new-value))))
+
+
+(reg-fx
+  :local-storage/dissoc!
+  (fn [[storage-key key]]
+    (let [old-value (or (edn/read-string (js/localStorage.getItem storage-key)) {})
+          new-value (dissoc old-value key)]
+      (js/localStorage.setItem storage-key new-value))))
```
tangjeff0, line 297: This is a good effect

src/cljs/athens/keybindings.cljs
```diff
@@ -0,0 +1,231 @@
...
+;; Wrap what is passed as "child" into a DOM node that listen for hotkeys that
+;; are bound to event handlers.
+;; If the application keymap changes, this component will re-bind the event
+;; handlers with the new hotkeys.
+;; keybindings is a pair of aliases-or-hotkeys and event handler pairs
+;; e.g: {:athena/toggle some-function1 "mod+z" some-function2}
+(defn mousetrap
+  [keybindings child]
+  (let [keymap @(subscribe [:keymap])]
+    [mousetrap-binder keymap keybindings child]))
```
tangjeff0, line 110: Are these the js `shenanigans`?

tangjeff0 (commented):
Overall, I think there are two points of contention.

1. It wasn't clear if we wanted to continue having both local+global keybindings vs. all global. I initially implemented both local+global, so it made sense to continue that pattern for @julionav.
2. I and @julionav were not aware of the library @neotyk suggested https://github.com/piranha/keybind.

These are two fairly large architectural decisions - local+global vs global and the keybinding library. On these two points, I would need more data as to why we should just refactor to all global and why we piranha/keybind is better than the other libraries evaluated by @julionav. Would love to hear more here @neotyk, probably good place for an architecture doc.

It seems like some of this could've been avoided if we had more design discussion beforehand. At the same time, hindsight is 20/20, and I believe this was a very good PR in terms of learning for @julionav. Going forward, we should try to recognize when a PR might require a broader architectural decision and have that discussion earlier. Happy to give a bounty to @julionav for the work here.

src/cljs/athens/keybindings.cljs
neotyk: Maybe `shenanigans` was too harsh, but yes.

PR #1264, opened 2021-06-02
fix(blocks): child drop-area-indicator shouldn't squish or be misplaced

### LINKED ISSUE #1262: Block drop-as-child indicator is not styled properly
**Problem**
The indicator's circle element is squished and off-center relative to the bar element

**Screenshots/Demo**
CleanShot 2021-06-01 at 17 26 06@2x

**Athens Version**
Beta.85

shanberg:
Fixes #1262

### DIFF
--- a/src/cljs/athens/views/blocks/drop_area_indicator.cljs
+++ b/src/cljs/athens/views/blocks/drop_area_indicator.cljs
@@ -1,6 +1,6 @@
 (ns athens.views.blocks.drop-area-indicator
   (:require
-    [athens.style :as style :refer [OPACITIES]]
+    [athens.style :refer [color OPACITIES]]
     [stylefy.core :as stylefy]))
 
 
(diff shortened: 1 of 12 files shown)


src/cljs/athens/views/blocks/drop_area_indicator.cljs
```diff
@@ -1,6 +1,6 @@
 (ns athens.views.blocks.drop-area-indicator
   (:require
-    [athens.style :as style :refer [OPACITIES]]
+    [athens.style :refer [color OPACITIES]]
```
shanberg, line 3: Reducing the requires here

src/cljs/athens/views/blocks/drop_area_indicator.cljs
```diff
@@ -24,34 +24,39 @@
...
+   :color (color :link-color)
    :position "relative"
    :transform-origin "left"
    :z-index 3
    :width "100%"
+   ;; Can't animate this, because the animation
+   ;; fires at too many states. If drop-area-indicator
+   ;; location is determined by order in the current
+   ;; page or block then we can reactivate animation.
+   ;; :animation "drop-area-appear 0.2s both"
```
shanberg, line 36: Might be better to just delete this

src/cljs/athens/views/blocks/drop_area_indicator.cljs
neotyk: It's more readable in code when we use `:as alias`, you know straight away that it's from outside local namespace.

neotyk (changes requested):

Issue #1263, opened 2021-06-02
roam-import: dates imported from Roam without zero padding


lwrogersjr:
**Problem**
Dates with single digit days in Roam are not padded to conform to the Athens format of Month DD, YYYY. For example, June 1st, 2021 in Roam is imported as May 2, 2021 instead of the Athens standard May 02, 2021. Also, the incorrectly formatted date page title cannot be edited to correct it.

On the daily page view, scrolling down from May 2, 2021 does not display "Earlier" and does not allow scrolling to May 01, 2021, etc.

Related to #1171 

**Screenshots/Demo**

**Athens Version**
1.0.0-BETA.85

PR #1268, opened 2021-06-02
fix(roam-import): single digit roam dates are padded with a left zero.

julionav:
Solves: https://github.com/athensresearch/athens/issues/1263

The issue was caused by `athens.patterns/replace-roam-date` only replacing the "cardinal" part of the date (th, st, etc); but not adding the 0 padding at the start of the number.


agentydragon:
Please fix style (run `script/style`). Also consider adding a test.

tangjeff0:
Yeah tests would be nice here (even though there aren't any for this yet).

julionav:
Added tests here, but when running script/style it says that there are 3 files incorrectly formatted, but it doesn't format them.

Getting this at the end:

```
Error encountered performing task 'run' with profile(s): 'base,system,user,provided,dev,cljstyle'
Suppressed exit
```

PR #1286, opened 2021-06-05
feat: add appearance preferences menu

shanberg:
What
- Adds appearance picker
- Adds content-width setting and options
- Moves theme switch into new appearance dropdown

Why
- We want to have more settings available than just swapping the theme dark/light
- The on/off toggle was never a great icon for the theme toggle
- Placeholder for future likely settings like content density, font, and zoom level

Help Wanted:
- [ ] Save content-width setting to presist beyond reload

image


src/cljs/athens/views/appearance_settings/core.cljs
```diff
@@ -0,0 +1,154 @@
+(ns athens.views.appearance-settings.core
```
tangjeff0, line 1: Does this need to be its own namespace?

src/cljs/athens/subs.cljs
```diff
@@ -143,3 +143,12 @@
       (:db/remote-graph-conf db)
       {})))
 
+
+;; Appearance Preferences
```
tangjeff0, line 147: Could begin putting some re-frame constructs like subs and events inside the view namespace, e.g. `src/cljs/athens/views/appearance_settings/core.cljs` in this case.

@neotyk do you have preference on colocating re-frame stuff vs keeping in separate events, i.e. all events in `events.cljs`, all subs in `subs.cljs`

src/cljs/athens/views/buttons.cljs
```diff
@@ -55,7 +55,7 @@
                              :text-align "left"}]
                      [:kbd {:margin-inline-start "1rem"
                             :font-size "85%"}]
-                     [:svg button-icons-style
+                     ["> svg" button-icons-style
```
tangjeff0, line 58: What's the difference here?

tangjeff0 (changes requested):
What's blocking making width setting persist?

src/cljs/athens/views/buttons.cljs
shanberg: `>` allows us to target only an SVG that's a direct descendant of the current context, and not any deeper ones.

src/cljs/athens/views/appearance_settings/core.cljs
shanberg: Up to you. Personally I prefer files much smaller than this, with more atomic concerns and less complexity.

shanberg:
> What's blocking making width setting persist?

@tangjeff0 It’s a low priority help-wanted. Fell off my radar when I moved to higher priority tasks.

src/cljs/athens/views/appearance_settings/core.cljs
```diff
@@ -0,0 +1,154 @@
...
+                     :key (:id option)
+                     :active (= (:id option) current)
+                     :on-click (:fn option)}
+             (:content option)]))])
+
+
+(def theme-settings
+  [{:content [:<> [:> Brightness7] [:span "Light"]]
+    :id "theme-light"
+    :fn #(dispatch [:theme/set-light])}
```
agentydragon, line 109: maybe you could have just one event, and give it an argument saying which theme to use?

e.g.: `(dispatch [:theme/set :dark])`

also, maybe it might make sense to rename the event & local storage keys to `appearance/...`, theme is an aspect of appearance.

src/cljs/athens/views/appearance_settings/core.cljs
```diff
@@ -0,0 +1,154 @@
...
+(def width-settings
+  [{:content [:<> normal-width-icon [:span "normal"]]
+    :id :normal
+    :fn #(dispatch [:appearance/set-width :normal])}
+   {:content [:<> large-width-icon [:span "large"]]
+    :id :large
+    :fn #(dispatch [:appearance/set-width :large])}
+   {:content [:<> full-width-icon [:span "full"]]
+    :id :full
+    :fn #(dispatch [:appearance/set-width :full])}])
```
agentydragon, line 124: might be neat to make this a loop

src/cljs/athens/views/appearance_settings/core.cljs
agentydragon: oh, I just remembered Athens already has themes so changing the key for themes would be annoying. disregard that idea.

sid597:
This so close to getting merged, I will take up the task of making this persistent. As of now, it is not straightforward to get and set local storage values to reframe db, the reason being re-frame gets reset whenever we start Athens, open or add a new db. Also, the situation: _we need to load values from local storage to re-frame db_ would arise again when a new 
- db is selected from db-picker
- A backup is selected in case of current db corruption

So I think it would be better to make _load data from local storage to re-frame_ a separate component, this would be one place where any new features which needs _load data from local storage to re-frame_  feature can be added. 

What do you think @shanberg @tangjeff0?
Will post progress updates.

sid597:
- Now can persist width to local storage 
- Now it would be easier to add more data that needs to set from local storage to rfdb
- Width is serialized and de-serialized while storing and retrieving from local storage

@tangjeff0 On a related note to (de)serialize we dont do that for `theme-dark`, this is stored and retrieved as a string, and when `theme-dark` needs to be `assoc` we do:

``` 
(let [is-dark (= "true" local-storage)
          theme   (if is-dark style/THEME-DARK style/THEME-LIGHT)]
      {:db          (assoc db :theme/dark is-dark)
```
This is not idiomatic and can be refactored, I did not do this in this PR because this can be done another PR with all other values that are set and retrieved from local storage.

src/cljs/athens/electron.cljs
```diff
@@ -315,6 +317,42 @@
             mtime (or mtime1 (.. fs (statSync filepath) -mtime))]
         (assoc db :db/mtime mtime))))
 
+  (reg-event-fx
+    :local-storage/set-rfdb
+    [(inject-cofx :local-storage-map {:ls-key "theme/dark"
+                                      :key    :theme-dark?})
+     (inject-cofx :local-storage-map {:ls-key "appearance/width"
+                                      :key    :appearance-width})]
```
tangjeff0, line 325: Overall, using cofx with localStorage here isn't a bad idea, but having two keys (one for local-storage and which more typical edn keyword) doesn't seem like a good idea. I know you didn't write `:local-storage-map`, but I wonder if there's a better way to have the same key for both localStorage and clojure.

src/cljs/athens/electron.cljs
```diff
@@ -315,6 +317,42 @@
             mtime (or mtime1 (.. fs (statSync filepath) -mtime))]
         (assoc db :db/mtime mtime))))
 
+  (reg-event-fx
+    :local-storage/set-rfdb
```
tangjeff0, line 321: Do you think `:local-storage/set-rfdb` should just replace `:init-rfdb`?

tangjeff0 (changes requested):

src/cljs/athens/electron.cljs
```diff
@@ -315,6 +317,42 @@
...
+      -Need to keep track of what type of value coeffect has
+      and de-serialize that before associating
+
+      Type of coeffect:
+      theme-dark? : boolean
+      appearance-width : keyword
+      "
+      (let [d-theme-dark?      (cljs.reader/read-string theme-dark?)
+            d-appearance-width (cljs.reader/read-string appearance-width)]
+        (if (and d-theme-dark? d-appearance-width)
```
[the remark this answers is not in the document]
sid597: Yeah, you are right I did not update this line, in my first commit I was not serializing `theme-dark?`  hence its value remained a `str`  and this worked, but after serializing I should have moved the `let` bindings inside the `if` true block

src/cljs/athens/electron.cljs
```diff
@@ -315,6 +317,42 @@
...
+
+          {:dispatch-n  [[:associate-key-val :appearance/width (:appearance/width db)]
+                         [:associate-key-val :theme/dark       (:theme/dark db)]]}))))
+
+
+
+  (reg-event-fx
+    :associate-key-val
+    (fn [{:keys [db]} [_ k v]]
+      {:db (assoc db k v)}))
```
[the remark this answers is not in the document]
sid597: @tangjeff0  Here we have to `assoc` both `d-appearance-width` and `d-theme-dark?` and I don't know how to do that. With 
```
{:db (assoc db :appearance/width d-appearance-width}
``` 
only `d-appearance-width` would be updated I don't know how do I update `d-theme-dark`? So I used another event to `assoc`  key values in db. Please guide me here.

src/cljs/athens/electron.cljs
sid597: I Will update after looking into this.

src/cljs/athens/events.cljs
```diff
@@ -609,14 +609,27 @@
...
+     :local-storage/set! ["theme/dark" true]
+     :stylefy/tag        [":root" (style/permute-color-opacities style/THEME-DARK)]}))
+
+
+(reg-event-fx
+  :theme/set-light
+  (fn [{:keys [db]} _]
+    {:db                 (assoc db :theme/dark false)
+     :local-storage/set! ["theme/dark" false]
+     :stylefy/tag        [":root" (style/permute-color-opacities style/THEME-LIGHT)]}))
```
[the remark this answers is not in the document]
sid597: @shanberg

src/cljs/athens/electron.cljs
sid597: Yeah now that I think about this my first implementation for this can be seen below, but in the first iteration, I did not (de)serialize `theme/dark` or `appearance/width` hence the UI misbehaved and I thought I was doing something wrong therefore to debug by printing values in the console log I took the re-frame approach. But I think following would be the better solution for setting `theme/dark` and other such values from local storage. What do you think @tangjeff0 ?
```
 :theme/dark    (or (js/localStorage.getItem (cljs.reader/read-string "theme/dark"))
                     false)
```

src/cljs/athens/events.cljs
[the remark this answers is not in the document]
sid597: Actually `:theme/toggle` is not split into 2 but it was removed from the codebase and `set-dark` and `set-light` took its place. Because `:theme/toogle` is not used in the codebase now.

martin-braun:
Honestly, I like this idea and I don't understand why something like this would be hidden into a settings menu. It would also be the perfect place to solve #2092.

PR #1287, opened 2021-06-05
fix(#1277): compare string without dangerous regex injection for slash cmd

### LINKED ISSUE #1277: Regex error when using parenthesis with slash command open
**Problem**
If you write something in parenthesis when the slash commands are open, a `invalid` regex error pops up.
Something like: `/(t)`

**Screenshots/Demo**
https://www.loom.com/share/1491573689f24de4b9fcda87d5aabebc

**Athens Version**
Beta 85

arthurbarroso:
Fixes #1277. `(re-pattern "invalid-pattern")` throws errors and Athens wouldn't handle those, prompting the user with javascript alerts.

### DIFF
--- a/src/cljs/athens/views/blocks/textarea_keydown.cljs
+++ b/src/cljs/athens/views/blocks/textarea_keydown.cljs
@@ -111,12 +111,14 @@
 
 (defn filter-slash-options
   [query]
-  (if (blank? query)
-    slash-options
-    (filterv (fn [[text]]
-               (re-find (re-pattern (str "(?i)" query)) text))
-             slash-options)))
-
+  ((try
+     (if (blank? query)
+       slash-options
+       #(filterv (fn [[text]]
+                   (re-find (re-pattern (str "(?i)" query)) text))
+                 slash-options))
+     (catch js/Error _
+       slash-options))))
 
 (defn update-query
   "Used by backspace and write-char.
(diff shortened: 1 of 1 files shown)


src/cljs/athens/views/blocks/textarea_keydown.cljs
```diff
@@ -111,40 +111,43 @@
...
-        new-query       (str (subs head query-start-idx) key)
-        results         (query-fn new-query)]
-    (if (and (= type :slash) (empty? results))
-      (swap! state assoc :search/type nil)
-      (swap! state assoc
-             :search/index 0
-             :search/query new-query
-             :search/results results))))
-
+  (try
```
agentydragon, line 129: is changing this function necessary? I don't see it creating a regex from a user-supplied string.

src/cljs/athens/views/blocks/textarea_keydown.cljs
```diff
@@ -111,40 +111,43 @@
...
-             slash-options)))
-
+  ((try
+     (if (blank? query)
+       slash-options
+       #(filterv (fn [[text]]
+                   (re-find (re-pattern (str "(?i)" query)) text))
+                 slash-options))
+     (catch js/Error _
+       slash-options))))
```
agentydragon, line 121: this just silences the problem and does not really fix it :/

also this will silence *any* JavaScript error, which is not a good idea.

this is basically a query for "does `text` contain `query`, case-insensitive", right? could we maybe just do `(str/includes? (st/lower-case text) (st/lower-case query))`?

agentydragon (changes requested):

src/cljs/athens/views/blocks/textarea_keydown.cljs
agentydragon: (see: https://clojuredocs.org/clojure.string/includes_q, https://clojure.github.io/clojure/clojure.string-api.html)

agentydragon: to expand: silencing broad classes of errors is a bad idea, because it can mask other problems and make them harder to notice and debug

arthurbarroso: Thank you for your review :-)

I completely agree that silencing errors is not the greatest approach. My main concern here was that typing an invalid Regex shouldn't result in a js/alert being shown to the end user.

I also agree the `str/includes?` approach would probably solve this problem in a fancier way and don't really ser any downsides in using `str/includes?` instead of `re-find` here. If we decide on going this way I can make a new commit and fix it!

agentydragon:
looks good to me. if you feel like it, might be nice to have a test.

agentydragon:
by the way, looks like there's a bunch of instances of the same thing in Athens:

```
$ ag '\(\?i\)'
(... irrelevant occurrences omitted ...)

src/cljs/athens/events.cljs
1799:  (let [ignore-case-title (re-pattern (str "(?i)" title))

src/cljs/athens/views/filters.cljs
152:                                     (re-pattern (str "(?i)" (:search @s)))

src/cljs/athens/views/blocks/textarea_keydown.cljs
117:               (re-find (re-pattern (str "(?i)" query)) text))

src/cljs/athens/db.cljs
402:  (re-pattern (str "(?i)" (escape-str query))))
```

Might be good to go through those and see if they could also be replaced with plain string operations. And maybe a shared function, like `includes-lowercase?`.

tangjeff0 (approved):
Nice first PR @arthurbarroso :sunglasses:

tangjeff0:
@agentydragon good suggestion. @arthurbarroso do you want to take that on?

arthurbarroso:
> @agentydragon good suggestion. @arthurbarroso do you want to take that on?

Yeah. I'll work on that.

PR #1292, opened 2021-06-07
Feat: find in page like chrome

arkRedM:
Prev pr: https://github.com/athensresearch/athens/pull/1229/files

Using [browserView](https://github.com/electron/electron/blob/v12.0.4/docs/api/browser-view.md) doesn't create a new  window


arkRedM:
Open comments from prev:

@tangjeff0  >> Anyway to reuse more existing styles rather than creating new ones? @shanberg
@shanberg >> Styling issues in win and ubuntu 
@tangjeff0  >> Additionally, should we decide to add additional styling and functionality to ctrl-f search, this would also be accessible on web.
@neotyk  >> https://github.com/athensresearch/athens/pull/1229#issuecomment-855300854

> @shanberg >> Styling issues in win and ubuntu 

I think they should be gone now(Could use some help in testing).

>  Additionally, should we decide to add additional styling and functionality to ctrl-f search, this would also be accessible on web.

web already has find in page and that is more familiar for users. Not sure if this is the best way. We can add this later if there is a use case for custom window in browser for find in page. 

> https://github.com/athensresearch/athens/pull/1229#issuecomment-855300854

I think this is not needed for now?

tangjeff0 (changes requested):
1. Any way to not make a new shadow build and html file?
2. Any way to use `findInPage` without `ipc`?

Demo https://www.loom.com/share/d3f4ac790965436d913e496866e8b690
- `t` shows 17 results, I see one on page
- Using browserView makes it so that `mod-+` or `mod--` changes zoom for that window
- searching for `222222` arbitary number of times gives orange and yellow highlighting in a weird way
- There is a delay on input (near end of demo). If main is blocked, renderer is also blocked
- searching for `z` shows has two results, I see on on page. That's because we have two DOM elements for each block - one for render, one for raw edit mode, but only one is visible at a time.

arkRedM:
> Any way to not make a new shadow build and html file?

Electron/findInPage doesn't really give much wiggle room here. I wish it to be a simple modal and findInPage takes in some selectors to ignore(but it doesn't provide that API) 
All we have from electron is windows(views, webviews etc) and anything we do in it is HTML + js(cljs).
*We can write plain js for the build and it means no separate build target(HTML will still be there)  but that's not what anyone wants.

> Any way to use findInPage without ipc?

Both the windows are isolated and have separate processes and doms for themselves. ipc is built for communicating across windows https://www.electronjs.org/docs/api/ipc-main Are there any reservations against it? 

> t shows 17 results, I see one on page

It could be the things inside left and right sidebar | even the empty state text. To be fair it shows similar in athens web demo(behaviour is same -- find in page not just visible context)

> Using browserView makes it so that mod-+ or mod-- changes zoom for that window

Should disable that behavior 

> searching for 222222 arbitary number of times gives orange and yellow highlighting in a weird way

It's matching correctly. "2222" and search string "22" will give two highlights(more are hidden with textarea stuff) 

> There is a delay on input (near end of demo). If main is blocked, renderer is also blocked

I'll experiment and get back.

>  searching for z shows has two results ... one for render, one for raw edit mode, but only one is visible at a time.

Visible here means display none or visibility hidden. Only then the matching is skipped. But we hide the textarea with opacity. I remember a patch which was being worked on which eliminates textarea rendering. 

Just to iterate, this works in all places(settings, all pages etc.) and it's focus is to find content on page(whichever page it is) and search for text(not notes).

tangjeff0:
Will re-open when we can focus on frontend improvements again.

PR #1317, opened 2021-06-09
refactor: use MUI/Dialog for Athena

juniusfree:
Main changes made:
1. used MUI/Dialog https://material-ui.com/api/dialog/
2. removed the key-handler for `esc` key because the Dialog component can handle that already.

### DIFF
--- a/src/cljs/athens/views/athena.cljs
+++ b/src/cljs/athens/views/athena.cljs
@@ -1,5 +1,6 @@
 (ns athens.views.athena
   (:require
+    ["@material-ui/core/Dialog" :default Dialog]
     ["@material-ui/icons/ArrowForward" :default ArrowForward]
     ["@material-ui/icons/Close" :default Close]
     ["@material-ui/icons/Create" :default Create]
@@ -14,7 +15,6 @@
     [cljsjs.react.dom]
     [clojure.string :as str]
     [garden.selectors :as selectors]
-    [goog.dom :refer [getElement]]
     [goog.events :as events]
     [re-frame.core :refer [subscribe dispatch]]
     [reagent.core :as r]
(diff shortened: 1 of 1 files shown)


src/cljs/athens/views/athena.cljs
```diff
@@ -274,76 +270,68 @@
...
+  (let [s (r/atom {:index   0
+                   :query   nil
+                   :results []})]
+    [:> Dialog
+     {:fullWidth          true
+      :maxWidth           "md"
+      :onClose            #(dispatch [:athena/toggle])
+      :open               @(subscribe [:athena/open])
+      :transitionDuration {:enter 0}}
+     [:div.athena (use-style container-style)
```
shanberg, line 282: I think we need to respecify the text color here, because the MUI dialog sets a text color and isn't aware of the current Athens theme.

CleanShot 2021-06-12 at 09 21 11@2x

shanberg (changes requested):
One styling bugfix. Otherwise this is good from a UX standpoint.

src/cljs/athens/views/athena.cljs
```diff
@@ -163,19 +163,18 @@
 
 
 (defn create-search-handler
```
neotyk, line 165: Please change the name from `create-search-handler` to `search-handler` since it's not creating a handler anymore but just handling.

src/cljs/athens/views/athena.cljs
```diff
@@ -1,5 +1,6 @@
 (ns athens.views.athena
   (:require
+    ["@material-ui/core/Dialog" :default Dialog]
```
neotyk, line 3: That's a bit too early.
We have not made the decision if we'll go with MUI or TailwindCSS, we have to test both with Steward.
https://github.com/athensresearch/athens/discussions/1316#discussioncomment-862581

neotyk (changes requested):

src/cljs/athens/views/athena.cljs
tangjeff0: It's "Stuart" btw :)

juniusfree:
Closing this for now. I'll re-open once there's a decision on which component library to use.  https://github.com/athensresearch/athens/discussions/1316

PR #1351, opened 2021-06-14
fix: deleting the page on the right sidebar doesn't remove the page

### LINKED ISSUE #1350: fix: deleting the page on the right sidebar doesn't remove the page
To reproduce:
1. Open a page on the right sidebar
2. Delete the page on the right sidebar

Expected output: The page will be deleted **and** removed from the right sidebar
Current output: The page is deleted but it's not removed from the right sidebar
image

Tested in 1.0.0.Beta.89

juniusfree:
fix #1350

**Problem 1**
- When deleting a page, Athens do not check and remove the page if it's on the right sidebar.

**Solution 1**
- Check if the page that's being deleted is in the right sidebar. If it is on the right sidebar, remove it.

**Problem 2**
- After deleting a page, Athens navigates to `:pages`. This is undesirable if the page that's being deleted on the right sidebar is different from the `current-route-uid`.

**Solution 2**
- Check if the page that's being deleted on the right sidebar is the same as the `current-route/uid`. If it's the same, then Athens will navigate to `:pages`, if not it'll stay at the same `current-route/uid`.


filipesilva (approved):
Thanks for fixing both of these problems @juniusfree!

juniusfree:
@filipesilva 🙇

PR #1361, opened 2021-06-15
feat(rtc): add presence toolbar

tangjeff0:
image

# Todos

- [X] implement UI
- [x] hook up to re-frame
  - [x] subscription
  - [x] event
  - [x] mock db data
- [ ] hook up to backend
  - [ ] creating event and dispatches for `:presence/hello` / `:presence/viewing` / `:presence/editing` -> update `:presence/users`
- [ ] Daily Notes edge case - in this case, the user on a page, but the route is `/home` rather than than `/page/{block-uid}`
  - use https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
  - use React Hooks https://non-traditional.dev/a-beginners-guide-to-dealing-with-refs-in-react-7dac6a355964
  - React hooks in cljs/reagent https://cljdoc.org/d/reagent/reagent/1.1.0/doc/tutorials/react-features
  - could probably refactor the current `db-scroll-daily-notes` as well

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -58,7 +58,7 @@
                :zoom-level          1
                :graph-conf          default-graph-conf
                :presence/users      [{:username "Zeus", :color "#DDA74C", :block/uid ""}
-                                     {:username "Poseidon", :color "#C45042", :block/uid "51c3580f5"}
+                                     {:username "Poseidon", :color "#C45042", :block/uid "94272f778"}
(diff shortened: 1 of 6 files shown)


tangjeff0:
Merged

sid597:
@tangjeff0 a minor issue, when screen width is below 900px the `find or create page` and the `toolbar` overlap 
Screenshot from 2021-06-24 16-01-00

PR #1458, opened 2021-07-29
Port `:paste` event

sid597:
Currently this is WIP, `paste`, I don't understand fully what problem this event is solving, also cannot quickly figure out looking at the current implementation what the problem is and why there is ~120 lines of code for this.


sid597:
Oops wrong branch

Issue #1455, opened 2021-07-28
in `transact-string` use `:block/save` event


filipesilva:

neotyk:
transact-string should be gone after Linkmaker 😁

PR #1466, opened 2021-08-04
resolves #1452 #1453 #1454 #1455 #1456

### LINKED ISSUE #1452: in `todo-on-click` fn use `:block/save` event

sid597:
WIP Doing local testing for
- [x] #1452 
- [x] #1453 
- [x] #1454 
- [x] #1456

### DIFF
--- a/src/cljc/athens/common_events/resolver.cljc
+++ b/src/cljc/athens/common_events/resolver.cljc
@@ -153,6 +153,7 @@
   [db {:event/keys [args]}]
   (let [{:keys [parent-uid
                 new-uid]} args
+        ;; Why do we set :db/id here, should this not be handled db?
         new-child         {:db/id        -1
                            :block/uid    new-uid
                            :block/string ""
@@ -1181,4 +1182,16 @@
                           new-str           (string/replace string ignore-case-title (str "[[" title "]]"))]
                       {:db/id [:block/uid uid] :block/string new-str}))
                   unlinked-refs)]
+    tx-data))
+
+
+(defmethod resolve-event-to-tx :datascript/block-open
+  [_db {:event/keys [args]}]
+  (println "resolver :datascript/block-open args:" (pr-str args))
+  (let [{:keys [block-uid
+                open?]}      args
+        new-block-state      {:db/add     [:block/uid block-uid]
+                              :block/open open?}
+        tx-data              [new-block-state]]
+    (println "resolver :datascript/block-open tx-data:" (pr-str tx-data))
     tx-data))
--- a/src/cljc/athens/common_events/resolver.cljc
+++ b/src/cljc/athens/common_events/resolver.cljc
@@ -122,10 +122,14 @@
 (defmethod resolve-event-to-tx :datascript/block-save
   [_db {:event/keys [args]}]
   (let [{:keys [uid
-                new-string]} args
+                new-string
+                add-time?]} args
         new-block-string     {:db/id        [:block/uid uid]
                               :block/string new-string}
-        tx-data              [new-block-string]]
+        block-with-time      (if add-time?
+                               (assoc new-block-string :edit/time (now-ts))
+                               new-block-string)
+        tx-data              [block-with-time]]
     (println ":datascript/block-save" (pr-str args)
              "=>" (pr-str tx-data))
     tx-data))
@@ -152,14 +156,18 @@
 (defmethod resolve-event-to-tx :datascript/add-child
   [db {:event/keys [args]}]
   (let [{:keys [parent-uid
-                new-uid]} args
+                new-uid
+                add-time?]} args
         ;; Why do we set :db/id here, should this not be handled db?
         new-child         {:db/id        -1
                            :block/uid    new-uid
                            :block/string ""
                            :block/order  0
                            :block/open   true}
-        reindex           (concat [new-child]
+        child-with-time   (if add-time?
+                            (assoc new-child :edit/time (now-ts))
+                            new-child)
+        reindex           (concat [child-with-time]
                                   (common-db/inc-after db [:block/uid parent-uid] -1))
         new-block         {:block/uid      parent-uid
                            :block/children reindex}
--- a/src/cljc/athens/common_events/resolver.cljc
+++ b/src/cljc/athens/common_events/resolver.cljc
@@ -158,7 +158,6 @@
   (let [{:keys [parent-uid
                 new-uid
                 add-time?]} args
-        ;; Why do we set :db/id here, should this not be handled db?
         new-child         {:db/id        -1
                            :block/uid    new-uid
                            :block/string ""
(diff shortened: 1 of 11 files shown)


sid597 (commented):
Some questions for @neotyk

sid597:
More questions for @neotyk :
1. What should be tested in the `block/open` event, we can est the value for a block to be open or close but that might still not be true if UI is not updated and we cannot check the UI in tests
2. What to do with `:transact/string`, you said we don't need it after linkmaker and I don't know how to move this to linkmaker

src/cljc/athens/common_events/resolver.cljc
```diff
@@ -153,6 +153,7 @@
   [db {:event/keys [args]}]
   (let [{:keys [parent-uid
                 new-uid]} args
+        ;; Why do we set :db/id here, should this not be handled db?
         new-child         {:db/id        -1
                            :block/uid    new-uid
```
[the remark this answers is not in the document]
neotyk: When inserting and providing negative ids, new concrete ids are going to be generated.
This allows you to say things like parent is a new block (-1) and it has a child (-2) and you can refer them from other places via these negative ids (within transaction).

src/cljs/athens/components.cljs
```diff
@@ -14,13 +14,14 @@
 
 (defn todo-on-click
   [uid from-str to-str]
-  (let [current-block-content (:block/string (db/get-block [:block/uid uid]))]
-    (dispatch [:transact [{:block/uid    uid
-                           :block/string (clojure.string/replace
-                                           current-block-content
-                                           from-str
-                                           to-str)
-                           :edit/time    (now-ts)}]])))
```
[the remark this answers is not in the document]
neotyk: `:edit/time` is our own property, it's not automatically managed by DB.
There is no DEFAULT of SQL in Datascript.

neotyk:
> More questions for @neotyk :
> 
> 1. What should be tested in the `block/open` event, we can est the value for a block to be open or close but that might still not be true if UI is not updated and we cannot check the UI in tests

We should test for state of `:block/open`.
Agree we can't test UI for now.

> 2. What to do with `:transact/string`, you said we don't need it after linkmaker and I don't know how to move this to linkmaker [in `transact-string` use `:block/save` event #1455](https://github.com/athensresearch/athens/issues/1455)

I don't see where `transact-string` is used here, please explain.

src/cljs/athens/components.cljs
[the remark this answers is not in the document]
sid597: Ah, I see `:edit/time` can be attached to pages and then it can be used to sort the pages by time, don't know how TODO uses this will check this and update.

sid597:
> I don't see where transact-string is used here, please explain.

Yeah `trnasact-string` is not used here because I did not make any changes to it. The function is In file `src/cljs/athens/views/pages/block_page.cljs` and the code related to it is : 

```
;; Helpers

(defn transact-string
  "A helper function that takes a `string` and a `block` and datascript `transact` vector
  ready for `dispatch`. Used in `block-page-el` function to log when there is a diff and `on-blur`."
  [string block]
  [:transact [{:db/id        [:block/uid (:block/uid block)]
               :block/string string
               :edit/time    (now-ts)}]])


(defn persist-textarea-string
  "A helper fn that takes `state` containing textarea changes and when user has made a text change dispatches `transact-string`. "
  [state block]
  (let [diff? (not= (:string/local state)
                    (:string/previous state))]
    (when diff?
      (dispatch (transact-string (:string/local state) block)))))
```

neotyk:
> > I don't see where transact-string is used here, please explain.
> 
> 
> 
> Yeah `trnasact-string` is not used here because I did not make any changes to it. The function is In file `src/cljs/athens/views/pages/block_page.cljs` and the code related to it is : 
> 
> 
> 
> ```
> 
> ;; Helpers
> 
> 
> 
> (defn transact-string
> 
>   "A helper function that takes a `string` and a `block` and datascript `transact` vector
> 
>   ready for `dispatch`. Used in `block-page-el` function to log when there is a diff and `on-blur`."
> 
>   [string block]
> 
>   [:transact [{:db/id        [:block/uid (:block/uid block)]
> 
>                :block/string string
> 
>                :edit/time    (now-ts)}]])
> 
> 
> 
> 
> 
> (defn persist-textarea-string
> 
>   "A helper fn that takes `state` containing textarea changes and when user has made a text change dispatches `transact-string`. "
> 
>   [state block]
> 
>   (let [diff? (not= (:string/local state)
> 
>                     (:string/previous state))]
> 
>     (when diff?
> 
>       (dispatch (transact-string (:string/local state) block)))))
> 
> ```
> 
> 
> 
> 

I confused it with walk-transact.

It seems to be straightforward, am I missing something?

sid597:
> It seems to be straightforward, am I missing something

No, I got confused when you said this is not needed now that we have linkmaker 😅

src/cljs/athens/events.cljs
```diff
@@ -945,7 +945,7 @@
 
 (reg-event-fx
   :block/save
-  (fn [_ [_ {:keys [uid old-string new-string callback] :as args}]]
+  (fn [_ [_ {:keys [uid old-string new-string callback add-time?] :as args}]]
```
sid597, line 948: I did not add the `add-time?` key to all the other places from where this event is called because this value will be `nil` if not passed and that would also work fine for this event. Is this right?

src/cljs/athens/views/pages/block_page.cljs
```diff
@@ -102,7 +98,8 @@
   (let [state (r/atom {:string/local    nil
                        :string/previous nil})]
     (fn [block parents editing-uid refs]
-      (let [{:block/keys [string children uid]} block]
+      (let [{:block/keys [string children uid]} block
+            block-uid                           (:block/uid block)]
```
neotyk, line 102: `block-uid` isn't needed, you have this value already in `uid` from line above.

neotyk (changes requested):
Left few comments, nothing big.
Well done.

src/cljs/athens/views/pages/block_page.cljs
sid597: 😅

src/cljs/athens/events.cljs
sid597: What if I did this : 

```
{:keys [block new-uid embed-id add-time?]
 :or   {add-time? false}
 :as args}
```
then I think we don't need to add multi-arity in the common event and remote event because `add-time?` value is always passed as boolean and not `nil`

filipesilva (approved):
I think everything @neotyk asked for is done, maybe with the exception of https://github.com/athensresearch/athens/pull/1466#discussion_r684107030. I don't quite get what Alex would like in this case but it doesn't sound like a blocker.

@neotyk if that item is important can you clarify a bit when you come back and we make the changes in a separate PR?

src/cljs/athens/events.cljs
neotyk: This will also work as goo

neotyk:
> I think everything @neotyk asked for is done, maybe with the exception of https://github.com/athensresearch/athens/pull/1466#discussion_r684107030. I don't quite get what Alex would like in this case but it doesn't sound like a blocker.
> 
> 
> 
> @neotyk if that item is important can you clarify a bit when you come back and we make the changes in a separate PR?

Commented more

PR #1472, opened 2021-08-05
common events paste

neotyk:
Porting `:paste` event to common events, only locally for now.
#1413


src/cljs/athens/events.cljs
```diff
@@ -1607,6 +1609,21 @@
                                         embed-id (str "-embed-" embed-id)) n]))]})))
 
 
+#_(reg-event-fx
```
filipesilva: I think this can be removed?

src/cljc/athens/common_events/resolver.cljc
```diff
@@ -966,3 +967,115 @@
                       {:db/id [:block/uid uid] :block/string new-str}))
                   unlinked-refs)]
     tx-data))
+
+
+(defn text-to-blocks
```
filipesilva, line 972: I'm a bit confused here, I see a lot of code was added but no code was removed. But the paste functionality already existed, so shouldn't its old code be removed in this PR?

src/cljc/athens/common_events.cljc
```diff
@@ -524,14 +524,18 @@
...
+  [last-tx uid text start value]
   (let [event-id (gen-event-id)]
     {:event/id      event-id
      :event/last-tx last-tx
      :event/type    :datascript/paste
-     :event/args    {:uid  uid
-                     :text text}}))
+     :event/args    {:uid   uid
+                     :text  text
+                     :start start
```
filipesilva, line 537: Will `:start` and `:value` be used later or something? They don't seem to be used now.

filipesilva (changes requested):
Couple of comments, and also there's merge conflicts that need to be resolved before this is merged.

src/cljc/athens/common_events.cljc
neotyk: Resolver uses these values.

src/cljs/athens/events.cljs
neotyk: It's removed.

src/cljc/athens/common_events/resolver.cljc
neotyk: Indeed, it wasn't removed from `events.cljs`

neotyk: Now it's gone from `events.cljs`.

filipesilva (approved):

PR #1485, opened 2021-08-12
feature/rtc db picker

filipesilva:
- fix: deref selected-db subs for open-local-comp
- rfct: support self-hosted dbs in all-dbs
- feat: connect to RTC via db-picker
- fix: don't send editing events with nil block-uid
- feat: allow resetting settings

### DIFF
--- a/README.self-hosted.md
+++ b/README.self-hosted.md
@@ -2,21 +2,7 @@
 
 ## Client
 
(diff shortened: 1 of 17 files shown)


filipesilva:
@neotyk can you take a look at these changes when you're back?

PR #1498, opened 2021-08-16
Fix: uncalled for `apply conj`

neotyk:


filipesilva (approved):
Nice, didn't even think of using just `into` instead of into+concat.

PR #1559, opened 2021-08-18
ci: only deploy to gh-pages on master

filipesilva:


.github/workflows/build.yml
```diff
@@ -126,7 +126,7 @@ jobs:
 
    deploy-gh-pages:
      needs: [test, lint, style, carve]
-     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
+     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v') && github.ref == 'refs/heads/master'
```
neotyk, line 129: Should it be just master?

filipesilva: I think so, yes. gh-pages is a singleton that should reflect the latest version on master only. We could deploy a web build version for other releases elsewhere, but not on gh-pages.

filipesilva: Actually, should be `main`, not master, as that's the right default branch.

filipesilva:
@neotyk updated PR content and title

PR #1564, opened 2021-08-19
build: support tsx components and storybook

filipesilva:


src/cljs/athens/devcards/right_sidebar.cljs
```diff
@@ -1,13 +1,13 @@
...
-    [athens.views.buttons :refer [button]]
+    ["/components/Button/Button" :refer [Button]]
     [athens.views.right-sidebar :as right-sidebar]
     [devcards.core :refer [defcard-rg]]
     [re-frame.core :refer [dispatch]]))
 
 
 (defcard-rg Toggle
-  [button {:primary true :on-click-fn #(dispatch [:right-sidebar/toggle])} "Toggle"])
+  [:> Button {:isPrimary true :on-click-fn #(dispatch [:right-sidebar/toggle])} "Toggle"])
```
filipesilva, line 10: Reagent will convert props in dash-chase into camelCase. So we can still keep using idiomatic property names in cljs.

filipesilva: Instead of using `:>` everywhere, we can also declare these components with `reagent/adapt-react-class` somewhere. Not sure if it's worth the effort though.

src/js/components/Button/button.css
```diff
@@ -0,0 +1,29 @@
+.storybook-button {
```
filipesilva, line 1: I don't think this file is used. Need to delete it.

src/cljs/athens/views/buttons.cljs
```diff
@@ -78,13 +78,13 @@
 (stylefy/class "button" buttons-style)
 
 
-(defn button
+#_ (defn button
```
filipesilva, line 81: This is the component that the PR moves to a JS component.

.storybook/preview.js
```diff
@@ -0,0 +1,146 @@
...
+  actions: { argTypesRegex: "^on[A-Z].*" },
+  controls: {
+    matchers: {
+      color: /(background|color)$/i,
+      date: /Date$/,
+    },
+  },
+}
+
+const cssVars = `
```
filipesilva, line 14: Discussed with Stuart how this is duplication of code in `athens.style`, and how we could put it in a JS file and use it as the source of truth for both setups. Left it as is for now to reduce scope of the PR.

filipesilva:
Note for reviewers: most of the file changes in this PR are changing the require and usage of Button.

.github/workflows/build.yml
```diff
@@ -238,6 +244,9 @@ jobs:
            mkdir -p ~/private_keys/
            echo '${{ secrets.api_key }}' > ~/private_keys/AuthKey_${{ secrets.api_key_id }}.p8
 
+       - name: Compile components
```
neotyk, line 247: What is the reason we compile components 3x?
Can't we cache the result of 1st complication and reuse in other 2 places?

neotyk (changes requested):
1 question about build workflow.

.github/workflows/build.yml
filipesilva: Roughly the same reason we compile the app in those 3 places too: the are being used as a build step prior to the app build step, so whenever the app gets built the components should be built as well.

It should be possible to cache the results of the compilation, especially since the component compilation is simpler than the app compilation. But caching brings its own set of problems, like determining the cache invalidation policy (yarn.lock+all ts files for now, maybe different later), and I can't imagine it will bring any benefit since compiling the components is faster than restoring cache+storing cache for now.

So I don't really see a reason to use cache at the moment.

tangjeff0 (changes requested):
Please take a look at questions.

Should we just get rid of devcards now?

Great work!

.github/workflows/build.yml
filipesilva: They aren't served anywhere now, no. I think they can be built with `yarn build-storybook` in this fashion. Afterwards they would need to be deployed.

For contributors to show them they'd also have to somehow deploy them to github pages though. How was that done before?

.storybook/preview.js
filipesilva: Not sure about the whole of athens.style, but at least part of it yes.

package.json
```diff
@@ -10,11 +10,16 @@
...
-    "prod": "shadow-cljs release main renderer app",
-    "clean": "rm -rf resources/public/**/*.js && rm -rf target && rm -rf .shadow-cljs",
-    "dist": "electron-builder -p always"
+    "dev": "concurrently \"yarn components:watch\" \"yarn client:watch\"",
+    "client:watch": "shadow-cljs watch main renderer app",
+    "components": "yarn tsc",
+    "components:watch": "yarn tsc --watch",
+    "compile": "yarn components && shadow-cljs compile main renderer app",
+    "prod": "yarn components && shadow-cljs release main renderer app",
+    "clean": "rm -rf resources/public/**/*.js && rm -rf target && rm -rf .shadow-cljs && rm -rf ./src/stories/**/*.js",
```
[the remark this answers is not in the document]
filipesilva: Good catch, done.

package.json
```diff
@@ -10,11 +10,16 @@
...
+    "dist": "electron-builder -p always",
+    "storybook": "start-storybook -p 6006",
+    "build-storybook": "build-storybook"
```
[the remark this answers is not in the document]
filipesilva: done

project.clj
```diff
@@ -58,7 +58,7 @@
 
   :min-lein-version "2.5.3"
 
-  :source-paths ["src/clj" "src/cljs" "src/cljc" "src/js"]
+  :source-paths ["src/clj" "src/cljs" "src/cljc" "src/js" "src/stories"]
```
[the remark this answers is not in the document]
filipesilva: good catch, fixed

.github/workflows/build.yml
filipesilva: Another note: athensresearch.github.io is essentially a singleton for the master branch deployment, we can't deploy there until rtc is merged into master.

tangjeff0: > For contributors to show them they'd also have to somehow deploy them to github pages though. How was that done before?

https://handbook.athensresearch.org/community/get-involved/setup#deploying-athens-and-devcards

> Another note: athensresearch.github.io is essentially a singleton for the master branch deployment, we can't deploy there until rtc is merged into master.

That's a good point. Right now, Github pages are only updated when main is deployed via a tagged release. 

In the past, I setup Actions so that contributors would update their GitHub pages whenever they pushed to any branch, not just when they pushed to their main. The problem here was that if they pushed to another branch, it would update. I even dynamically included their Git hash and linked to the GitHub commit in their GitHub pages Athens build. Actually, none of this worked out that well, and the easiest ended up being that I just added their fork as a remote and built it locally every time I was reviewing their PR. This should probably still be the workflow going forward for PR reviews.

.gitignore
```diff
@@ -31,3 +31,6 @@
 
 # electron build
 dist
+
+# design system ts output
+/src/js/components/**/*.js
```
tangjeff0, line 36: For some reason, I my git is recognizing unstaged changes in `/src/js/components/Button/Button.js` and `/src/js/components/Button/Button.stories.js`. The diff is just different slightly differenent transpiled TS=>JS.

doc/adr/000x-js-components.md
```diff
@@ -0,0 +1,45 @@
...
+* Better development and testing environment for components
+* Clearer defined deliverable scope and context for components
+* Component documentation
+* Smaller and simpler application codebase 
+* Looser coupling between engineering and design
+* Design can deliver work on separate cadence, units, and timeline, from engineering
+* Easier to enforce discipline on components as pure functions
+* CLJS proficiency is not necessary for design work
+* Easier for contributors to contribute to components
+* Reduced bus factor in design
```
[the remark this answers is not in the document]
filipesilva: Added

.gitignore
filipesilva: What's your version of git? Recursive glob support was only added in 1.8.2.

filipesilva:
@tangjeff0 removed devcards, which removed a LOT of other stuff via `carve`.

src/cljs/athens/devcards/right_sidebar.cljs
tangjeff0: I prefer dash-case.

I prefer `:>`.

.storybook/preview.js
tangjeff0: Yeah just the part that is all the CSS properties.

.github/workflows/build.yml
tangjeff0: The problem with GH pages is it would only be able to serve one branch at a time. I feel like a service like Vercel could have a deploy for an arbitrary number of git hashes. This is too much overhead for most contributors, but I could seem it being more valuable when we have several people iterating on frontend code.

.gitignore
tangjeff0: 2.25.1. Can look into it and add my repro steps.

src/cljs/athens/devcards.cljs
```diff
@@ -27,8 +27,6 @@
     [athens.effects]
```
tangjeff0, line 27: Delete the main devcards file?

src/cljs/athens/devcards/style_guide.cljs
```diff
@@ -1,116 +0,0 @@
-(ns athens.devcards.style-guide
-  (:require
-    [athens.db]
-    [athens.style :refer [color THEME-LIGHT THEME-DARK OPACITIES]]
-    [cljsjs.react]
-    [cljsjs.react.dom]
-    [devcards.core :refer-macros [defcard-rg]]
-    [stylefy.core :as stylefy :refer [use-style]]))
```
tangjeff0, line 8: @shanberg do we have analogs for this in storybook yet? Should make sure we do before deleting this file and have to go looking for it in the future.

src/cljs/athens/devcards/styling_with_stylefy.cljs
```diff
@@ -1,136 +0,0 @@
-(ns athens.devcards.styling-with-stylefy
-  (:require
-    [athens.db]
-    [cljsjs.react]
-    [cljsjs.react.dom]
-    [devcards.core :refer-macros [defcard-doc]]))
```
tangjeff0, line 6: @shanberg Similar question that I had in style_guide.cljs. Though it's less relevant now that we can use styled-components, it took a while to figure out the intricacies of Stylefy, and I want that knowledge to be easily accessible in the future. Not sure we need it to be in storybook, but it should be somewhere. The code here could be a handy, evergreen reference.

src/js/components/Button/Button.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+export interface ButtonProps {
+  /**
+   * Is this the principal call to action on the page?
+   */
+  isPrimary?: boolean;
+  /**
+   * Is this the principal call to action on the page?
+   */
+  isPressed?: boolean;
```
tangjeff0, line 98: Can we explicitly define what conventions we use, and eventually enforce them to the extent we can with build tools? There are 4 if I'm not mistaken.

CLJS: `primary?`
CSS: `is-primary`
JS: `isPrimary`
TS: `isPrimary?`

src/cljsjs/create_react_class.cljs
```diff
@@ -1,8 +0,0 @@
-(ns cljsjs.create-react-class
-  (:require
-    ["create-react-class" :as create-react-class]
-    ["react" :as react]))
-
-
-(js/goog.object.set react "createClass" create-react-class)
-(js/goog.exportSymbol "createReactClass" create-react-class)
```
tangjeff0, line 8: Why aren't cljsjs namespaces necessary anymore?

src/cljs/athens/views/buttons.cljs
```diff
@@ -1,90 +0,0 @@
...
-(defn button
-  "Keep button interface as close to vanilla hiccup as possible.
-  Dissoc :style :active and :class because we don't want to merge them in directly.
-  Can pass in a :key prop to make react happy, as a :key or ^{:key}. Just works"
-  ([children] [button {} children])
-  ([{:keys [style active primary class] :as props} children]
-   (let [props- (dissoc props :style :active :primary :class)]
-     [:button (use-style (merge buttons-style style)
-                         (merge props- {:class (vec (flatten [(when active "is-active") (when primary "is-primary") class]))}))
-      children])))
```
tangjeff0, line 90: RIP light-wrapper `button` over `:button` @shanberg

src/cljs/athens/views/filters.cljs
```diff
@@ -1,230 +0,0 @@
...
-    ["@material-ui/icons/FilterList" :default FilterList]
-    ["@material-ui/icons/Sort" :default Sort]
-    [athens.style :refer [color OPACITIES]]
-    [athens.views.buttons :refer [button]]
-    [athens.views.textinput :refer [textinput]]
-    [cljsjs.react]
-    [cljsjs.react.dom]
-    #_[re-frame.core :as re-frame :refer [dispatch]]
-    [reagent.core :as r]
-    [stylefy.core :as stylefy :refer [use-style #_use-sub-style]]))
```
tangjeff0, line 15: Filters never made it from devcards into the app, but I'm not sure sure we should delete this file as we may implement some of this UI again: dropdown + search + filter buttons.

Should we create storybook analogy first? Could also save a permalink to this file somewhere, just need to make sure we can find it. @shanberg

.gitignore
tangjeff0: Seems fine now. But `storybook-static/` code is being created. Do we want all these files?

tangjeff0 (changes requested):
We should add a project-level JS linter now that we are adding javascript, but in a separate PR. I'm noticing some inconsistencies regarding semi-colons and whitespace in the new JS files.

src/js/components/Button/Button.tsx
filipesilva: On `TS` `isPrimary?` means the prop is optional. For it to have `?` in the prop name it would have to be a string, e.g. `"isPrimary?"`.

I think the only distinction we should have is between camelCase and dash-case. CLJ tools usually convert camelCase to dash-case on interop.

src/cljsjs/create_react_class.cljs
filipesilva: They never were really necessary with `shadow-cljs`. But devcards was a bit broken and required them as a workaround for https://github.com/bhauman/devcards/issues/165.

.storybook/preview.js
shanberg: This has been implemented in my current branch https://github.com/shanberg/athens/commit/dca06885cc21c086010eacd45c9d3fb8bb492310

src/cljs/athens/devcards/style_guide.cljs
shanberg: I'll add it to my list of TODOs

src/js/components/Button/Button.tsx
shanberg: I prefer is/has/should prefixes for booleans in TS, since the "?" isn't available to distinguish them.

src/cljs/athens/views/buttons.cljs
shanberg: We made a better one

src/cljs/athens/views/filters.cljs
shanberg: It’s still in the GitHub history even if we remove it from the latest code. Do we need to hang on to it?

src/cljs/athens/devcards/styling_with_stylefy.cljs
shanberg: The only additions this documentation contains over the actual Stylefy documentation are the convention of appending `-style` to component style definition names and the preference for using the SCSS-like `stylefy/manual` mode rather than sub styles.

I'm not sure it’s worth keeping around if we extract Stylefy from the codebase, which I see as the natural final step in the transition.

src/cljs/athens/devcards.cljs
filipesilva: Urf durf done now

.github/workflows/build.yml
filipesilva: WRT https://github.com/athensresearch/athens/pull/1564#discussion_r698394605, talked to Alex on standup and he prefer the faster build than the caching for now.

neotyk (approved):

.gitignore
filipesilva: Ignored

src/js/components/Button/Button.tsx
filipesilva: Stuart thinks there's no action here, since it seems we're just following CLJS->JS naming conversion, is there something you'd like to be done here @tangjeff0?

src/cljs/athens/devcards/style_guide.cljs
filipesilva: I would prefer to not leave the files in because:
- the files themselves are not enough to remember to do this
- we'd need to configure out linting/etc to ignore all these broken files (or to leave all the unused libs and deps in)

Instead I propose to leave these as deferred work items in the ADR (see https://github.com/athensresearch/athens/pull/1564/commits/8c668d2c8b6f673adadd1f539d879ccd8b41590b). WDYT?

src/cljs/athens/devcards/styling_with_stylefy.cljs
filipesilva: See https://github.com/athensresearch/athens/pull/1564#discussion_r704565717

src/cljs/athens/views/filters.cljs
filipesilva: See https://github.com/athensresearch/athens/pull/1564#discussion_r704565717

src/js/components/Button/Button.tsx
tangjeff0: Ok

src/cljs/athens/devcards/style_guide.cljs
tangjeff0: Ok

src/cljs/athens/electron/db_modal.cljs
```diff
@@ -12,11 +13,8 @@
     [athens.style :refer [color]]
     [athens.subs]
     [athens.util :refer [js-event->val]]
-    [athens.views.buttons :refer [button]]
     [athens.views.modal :refer [modal-style]]
     [athens.views.textinput :as textinput]
-    [cljsjs.react]
-    [cljsjs.react.dom]
```
juniusfree, line 19: The removal of this results in this error when adding a db
image

filipesilva: Fix in https://github.com/athensresearch/athens/pull/1641, thanks for the catch!

PR #1565, opened 2021-08-19
fix: problems in the daily page scroll

### LINKED ISSUE #1492: Problem with the daily pages scroll (ALPHA.RTC.1 - Local DB)
**Problem**
When using the new ALPHA.RTC.1 with a local DB, there is a problem with the daily pages scrolling down :
* If the daily page already exists, it will act as usual
* If the daily page doesn't exist, it will create it then jump out of the "daily mode"

And when I click to the "daily page" button, it will jump to the latest opened daily page (not today's page)

**Screenshots/Demo**
https://user-images.githubusercontent.com/24638389/129581245-03edaa74-a340-4ada-b732-ab28dc902c81.mov

**Athens Version**
1.0.0-ALPHA.RTC.1 - Intel OSX

juniusfree:
fixes #1492

From the issue:
- [x] If the daily page already exists, it will act as usual
- [x] If the daily page doesn't exist, it should create it but it shouldn't open that date's node-page.
- [x]  (feature request) ~~when I click on the "daily page" button, it will jump to the latest opened daily page (not today's page).~~ (I decided to maintain the current behavior)

Additional check:

- [x] ~~In remote, after creating a new daily page via scrolling, the new page should be displayed immediately. The delayed response from the server is the issue.~~ CORRECTION: The problem is that the `:daily-notes/items` in app-db is updated first before the page is created.

### DIFF
--- a/src/cljs/athens/events/remote.cljs
+++ b/src/cljs/athens/events/remote.cljs
@@ -5,6 +5,7 @@
     [athens.common-events.resolver :as resolver]
     [athens.common-events.schema   :as schema]
     [athens.db                     :as db]
+    [athens.util                    :as util]
     [malli.core                    :as m]
     [malli.error                   :as me]
     [re-frame.core                 :as rf]))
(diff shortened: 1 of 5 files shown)


src/cljs/athens/events/remote.cljs
```diff
@@ -193,9 +194,13 @@
...
                   block-uid]} (:event/args event)]
       (js/console.log ":remote/followup-page-create, page-uid" page-uid)
-      {:fx [[:dispatch-n [(if shift?
+      {:fx [[:dispatch-n [(cond
+                            shift?
                             [:right-sidebar/open-item page-uid]
+
+                            (not (util/is-daily-note page-uid))
                             [:navigate :page {:id page-uid}])
+
```
neotyk, line 203: To make it fully work you'll need to add to daily-pages.
Something like this:
```clojure
(util/is-daily-note page-uid)
[:daily-note/add page-uid]
```

PR #1567, opened 2021-08-20
Feature: Drafting ADR for Atomic/Composite Graph Operations.

### LINKED ISSUE #1557: Create ADR for Atomic/Composite graph operations

neotyk:
Closes #1557


sid597 (commented):
Some questions

doc/adr/0010-atomic-composite-grapth-operations.md
```diff
@@ -0,0 +1,267 @@
...
+            - ((⚛️ Atomic Graph Ops))
+            - *Input*
+                - `page-uid` - `:block/uid` of page to be renamed
+                - `new-name` - Page should have this name after operation
+                - `old-name` - ^^To Remove^^ This is accidental, and shouldn't be provided
+        - ⚛️**`:page/merge`**
+            - ((⚛️ Atomic Graph Ops))
+            - *Input*
+                - `page-uid` - `:block/uid` of page to be merged into `new-page`
+                - `new-page` - page name of a page we'll merge contents of `page-uid` page into
```
filipesilva, line 109: This should be `target-page-uid`, I think. It's a bit weird how page identity seems split between name and uid. Some operations take both because they really needed, but if they just need to identify the page then there should be a single unique identifier, and that sounds like the uid.

doc/adr/0010-atomic-composite-grapth-operations.md
```diff
@@ -0,0 +1,267 @@
...
+        - Like `:block/save` when new link is discovered, should produce also `:page/create` event
+
+- List of Atomic Graph Operations:
+    - Page Ops
+        - ⚛️**`:page/new`**
+            - ((⚛️ Atomic Graph Ops))
+            - *Input*
+                - `title` - Page title page to be created
+                - `page-uid` - `:block/uid` of page to be created
+                - `block-uid` - `:block/uid` of 1st block to be created in page to be created
```
filipesilva, line 98: Needing a block to create a page sounds more like an incidental requirement of the web client than an integral part of page creation.

doc/adr/0010-atomic-composite-grapth-operations.md
```diff
@@ -0,0 +1,267 @@
...
+        - ⚛️**`:page/rename`**
```
filipesilva, line 99: I think creating a new page and renaming a page are actually the same operation. Partly because they take the same args (since I don't think the first block is actually part of page/new), and partly because the only way to distinguish in their usage is by verifying whether or not the page exists before.

If we make them the same atomic op we have a couple of extra benefits that are very desirable in a concurrent and partial information environment:
- idempotency
- doesn't require knowing if the page exists or not already

doc/adr/0010-atomic-composite-grapth-operations.md
```diff
@@ -0,0 +1,267 @@
...
+            - ((⚛️ Atomic Graph Ops))
+            - *Input*
+                - `page-uid` - `:block/uid` of page to be renamed
+                - `new-name` - Page should have this name after operation
+                - `old-name` - ^^To Remove^^ This is accidental, and shouldn't be provided
+        - ⚛️**`:page/merge`**
```
filipesilva, line 105: Page merge sounds a lot like a bunch of block/move, block/save (because of refs), and a page/remove.

doc/adr/0010-atomic-composite-grapth-operations.md
```diff
@@ -0,0 +1,267 @@
...
+                - `page-uid` - `:block/uid` of the page to be deleted
+    - Block Ops
+        - ⚛️**`:block/new`**
+            - ((⚛️ Atomic Graph Ops))
+            - *Input*
+                - `parent-uid` - `:block/uid` of parent block (or page)
+                - `block-uid` - `:block/uid` of new block to be created
+                - `block-order` - `:block/order` of new block to be created
+                    - Currently it's only `int`
+                    - We could extend it to allow `int` and 2 keywords `:first` & `:last` (to say that we want this new block to be 1st among the children of `parent-uid` or last)
```
filipesilva, line 123: `:after` and `:before` would also make it much more semantically meaningful in concurrent scenarios.

doc/adr/0010-atomic-composite-grapth-operations.md
```diff
@@ -0,0 +1,267 @@
...
+            - *Input*
+                - `page-uid` - `:block/uid` of page to be merged into `new-page`
+                - `new-page` - page name of a page we'll merge contents of `page-uid` page into
+                - `old-name` - ^^To Remove^^ This is accidental, and shouldn't be provided
+        - ⚛️**`:page/remove`**
+            - ((⚛️ Atomic Graph Ops))
+            - *Input*
+                - `page-uid` - `:block/uid` of the page to be deleted
+    - Block Ops
+        - ⚛️**`:block/new`**
```
filipesilva, line 116: Block new, as defined here, is the same as block/move WRT the args. I imagine the difference is that it also sets an empty string or something. But then it's a composite between save and move.

This distinction is meaningful is conflict resolution. You can succeed at editing the content but fail on setting the position, and other combinations. Composing these two operations also ensures idempotency.

doc/adr/0010-atomic-composite-grapth-operations.md
```diff
@@ -0,0 +1,267 @@
...
+    - (({{[[TODO]]}} ⚛️**`:shortcut/move`**))
+
+
+((⎄ Composite Graph Ops)) are a way to group ((⚛️ Atomic Graph Ops)).
+
+We'll only need one type of Composite Graph Ops, that is Consequence Event.
+
+Consequence Event is 2 things:
+- Trigger Event (like `:paste`)
+- Consequence Events List
```
filipesilva, line 60: Nit: it's a bit weird to have names like `trigger` and `consequence` to describe what's just a nested tree structure.

filipesilva (approved):
LGTM besides some non-blocking discussion points about idempotency.

It was very useful to see the listing of all ops.

tangjeff0 (commented):
So with composite graph events, did we essentially just reimplement re-frame events and effects?

doc/adr/0010-atomic-composite-grapth-operations.md
neotyk: How can you save or move block that doesn't exist?
IMHO Block creation is atomic, and isn't composition of save and move.

doc/adr/0010-atomic-composite-grapth-operations.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Decision
+
+We have two kinds of events to modify graph:
+- ⚛️ Atomic Graph Ops
+    - Not divisible Graph Ops
+    - Operations like create new block, create page, save block
+- ⎄ Composite Graph Ops
+    - Collection of events to be executed on the graph
+    - Like `:block/save` when new link is discovered, should produce also `:page/create` event
```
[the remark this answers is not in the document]
neotyk: It's one and the same.

doc/adr/0010-atomic-composite-grapth-operations.md
neotyk: Please propose better names.

doc/adr/0010-atomic-composite-grapth-operations.md
neotyk: We always create page with a block.
Maybe because it's always being done like that.
But I'm not sure what is assuming that pages always have blocks.

I see it as page creation will create 1st block in this page.

doc/adr/0010-atomic-composite-grapth-operations.md
neotyk: Page creation does create 1st block in that page as well.
I can't see how page/create and page/rename are the same thing.

doc/adr/0010-atomic-composite-grapth-operations.md
neotyk: Yes, thought this would be next to order, so additional attribute.
Maybe 🤔

PR #1568, opened 2021-08-20
fix: Some bullets are not round

### LINKED ISSUE #1495: Some bullets are not round ? (ALPHA.RTC.1 - Local DB)
**Problem**
Some bullets are not round.
Couldn't manage to find a systematic way to create these "not round" bullets

**Screenshots/Demo**
image

**Athens Version**
1.0.0-ALPHA.RTC.1 - Intel OSX

juniusfree:
fix #1495

### DIFF
--- a/src/cljs/athens/views/blocks/bullet.cljs
+++ b/src/cljs/athens/views/blocks/bullet.cljs
@@ -28,7 +28,7 @@
    ::stylefy/manual [[:&:after {:content "''"
                                 :background "currentColor"
                                 :transition "color 0.05s ease, opacity 0.05s ease, box-shadow 0.05s ease, transform 0.05s ease"
-                                :border-radius "100px"
+                                :border-radius "50%"
                                 :box-shadow "0 0 0 0.125rem transparent"
                                 :display "inline-flex"
                                 :margin "50% 0 0 50%"
--- a/src/cljs/athens/views/blocks/bullet.cljs
+++ b/src/cljs/athens/views/blocks/bullet.cljs
@@ -27,8 +27,8 @@
    :align-items "center"
    :justify-content "center"
    :color (style/color :body-text-color :opacity-low)
-   ::stylefy/manual [[:svg {:transform "scale(1.001)" ;; Prevents the bullet being squished
-                            :overflow "visible" ;; Prevents the stroke from being cropped
+   ::stylefy/manual [[:svg {:transform "scale(1.001)" ; Prevents the bullet being squished
+                            :overflow "visible" ; Prevents the stroke from being cropped
                             }]
                      [:circle {:fill "currentColor"
                                :transition "color 0.05s ease, opacity 0.05s ease, box-shadow 0.05s ease, transform 0.05s ease"}]
--- a/src/cljs/athens/views/blocks/bullet.cljs
+++ b/src/cljs/athens/views/blocks/bullet.cljs
@@ -45,8 +45,9 @@
                      [:&:hover:before
                       :&:focus-visible:before {:opacity 1}]
                      [:&.closed-with-children [:circle {:stroke (style/color :body-text-color)
-                                                        :r "8"
-                                                        :stroke-width "4"
+                                                        :fill (style/color :body-text-color :opacity-low)
+                                                        :r "22"
+                                                        :stroke-width "12"
                                                         :opacity (:opacity-med style/OPACITIES)}]]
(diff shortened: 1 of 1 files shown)


filipesilva (commented):
I'm a bit confused as to why this is the right fix, the border radius has not changed in a while.

Could effecting this change actually be hiding the real problem, where a bullet might be rendered twice?

neotyk:
> I'm a bit confused as to why this is the right fix, the border radius has not changed in a while.
> 
> Could effecting this change actually be hiding the real problem, where a bullet might be rendered twice?

I've checked it looks like there is just 1 element.
Not sure how we got here, I'll add it to result of css-garden approach.

filipesilva:
@neotyk I defer to you, you have a much better understanding of this.

juniusfree:
I might have rushed this one because I can see a subtle difference after the update (or am I? 😵). I'll try to investigate if this has deeper issues.

shanberg:
I think the answer here is we should be using SVG circles instead of pseudo-elements.

If the DOM is telling the bullet to be taller than it is wide or wider than it is tall, the only change we've made is that it'll be elliptical rather than squarish, and we haven't told the browser to keep it always the same width/height. A very simple SVG should have an easier time staying at a 1:1 aspect ratio.

shanberg:
It looks like I misunderstood the nature of this problem.

I had assumed that something was squishing the bullet and we just needed to use a stronger method of defining its shape but I'm still seeing cases where some bullets are elliptical, even with the SVG.

What seemed to solve all cases is to add `transform: scale(1.01)` to the bullet SVG. I'm not absolutely sure why, but given that the calculated size of the bullets has always been squared, I'm guessing it changes the way the element treats subpixels during rendering. Regardless, this is the fix I recommend at this point. It’s essentially a hack, but not anything fundamentally crazy or likely to cause further issues.
- - -
Note: Switching from `em` units to `px` units in `bullet-style` fixed the shape, but the `em` units allow the bullets (and everything else in blocks) to scale down in the sidebar, so switching to `px` units isn't a magic wand and I don't recommend it.

src/cljs/athens/views/blocks/bullet.cljs
```diff
@@ -24,17 +23,12 @@
...
-                                :transition "color 0.05s ease, opacity 0.05s ease, box-shadow 0.05s ease, transform 0.05s ease"
-                                :border-radius "100px"
-                                :box-shadow "0 0 0 0.125rem transparent"
-                                :display "inline-flex"
-                                :margin "50% 0 0 50%"
-                                :transform "translate(-50%, -50%)"
-                                :height "0.3125em"
-                                :width "0.3125em"}]
+   ::stylefy/manual [[:circle {:fill "currentColor"
+                               :transition "color 0.05s ease, opacity 0.05s ease, box-shadow 0.05s ease, transform 0.05s ease"}]
```
shanberg, line 31:
```suggestion
   ::stylefy/manual [[:svg {:transform "scale(1.001)" ;; Prevents the bullet being squished
                            :overflow "visible" ;; Prevents the stroke from being cropped
                            }]
                     [:circle {:fill "currentColor"
                               :transition "color 0.05s ease, opacity 0.05s ease, box-shadow 0.05s ease, transform 0.05s ease"}]
```

src/cljs/athens/views/blocks/bullet.cljs
```diff
@@ -47,9 +41,11 @@
                      [:&:hover {:color (style/color :link-color)}]
                      [:&:hover:before
                       :&:focus-visible:before {:opacity 1}]
-                     [:&.closed-with-children [:&:after {:box-shadow (str "0 0 0 0.125rem " (style/color :body-text-color))
-                                                         :opacity (:opacity-med style/OPACITIES)}]]
-                     [:&:hover:after {:transform "translate(-50%, -50%) scale(1.3)"}]
+                     [:&.closed-with-children [:circle {:stroke (style/color :body-text-color)
+                                                        :r "8"
+                                                        :stroke-width "4"
```
shanberg, line 46: Update the style to look closer to the previous bullets

```suggestion
                                                        :fill (style/color :body-text-color :opacity-low)
                                                        :r "22"
                                                        :stroke-width "12"
```

shanberg (changes requested):
I was wrong about SVG fixing it, but switching to SVG is better for a bunch of reasons anyway. 

I've suggested a new fix for the bullet squishing.

shanberg (approved):

PR #1570, opened 2021-08-20
Multi-platform client

### LINKED ISSUE #1497: Fix the web build
Currently karma tests are failing because of node dependencies leak to web build.

filipesilva:
Fix #1497

### DIFF
--- a/doc/adr/0008-local-storage.md
+++ b/doc/adr/0008-local-storage.md
@@ -1,4 +1,4 @@
-# x. Local Storage
+# 8. Local Storage
 
 Date: 2021-07-01
 
--- a/doc/adr/000x-multi-platform-client.md
+++ b/doc/adr/000x-multi-platform-client.md
@@ -1,0 +1,81 @@
+# x. Multi-platform Client
+
+Date: 2021-08-20
+
+
+## Status
+
+Pending
+
+
+## Context
+
+The Athens client is built for two platforms, Electron and Web, using the same compilation unit.
+This is possible in CLJS due to the dynamic nature of using Electron dependencies via `js/require`, allowing compilation to finish successfully.
+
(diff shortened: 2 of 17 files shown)


filipesilva:
I favor the combination of:
- using Clojure conditionals for platform specific requires
- adding explicit errors if the require isn't available
- not using Closure Defines because it's extra mental overhead
- using platform namespaces
- using database types that define how they should be stored

neotyk:
- Initial #[[ADR]] feedback
    - Reader conditional using targets `#{:node :browser}`
        - I like it because we can have:
            - We can do target specific require
                - So only on `:node` require node specific namespaces
            - ```clojure
              (rf/reg-event-fx :event
              #?(:node (fn ..)
                 :browser (fn ..)))
              ```
                - Where in `:node` it actually does something like file I/O,  and in `:browser` we might dispatch an platform feature unavailable message
    - We could maybe have runtime feature availability flags
        - Like `:local-fs` or things like that
            - ```clojure
              (defn feature-available? [feature-key]
              ; runtime check if `feature-key` is available
              )
              ```
            - So we know if we can display File System option in DB Picker

filipesilva:
Worth mentioning it's not quite `node` vs `browser` - both the electron and web builds are using the browser environment, but the electron build has a couple of extra exposed globals. We can also define the names ourselves, it doesn't have to be `node` and `browser`. I think the right ones are `electron` and `browser`.

The conditionals can only be used on `.cljc` files so we'd have to be a bit careful there, but I think it's doable.

I like your proposal that the event itself dispatches the error message. That's strictly better than having different ones with a name convention (e.g. boot-web and boot-electron).

The `feature-available` checks also SGTM. We can also use the conditionals there.

PR #1575, opened 2021-08-23
feat: replace block refs with content

### LINKED ISSUE #1448: Deleting a block/page should replace references to it with its string content
Alex and I discussed that case a while back. The real question being "what happens to refs in text when the ref'd thing is deleted". It applies to anything that can be reffed (pages and blocks right now).

The decision we came to is that the refs should be replaced with their content, and that it's the events responsibility to do so, which will then trigger linkmaker to remove the `:block/refs`. The idea is that the strings are the source of truth and now they changed.

So for instance if you have `here is ((block-uid))` and `block-uid` has `an orange` as `:block/string`, deleting `block-uid` would leave the other block with `here is an orange` as `:block/string`. If it was a title instead, the same would happen.

I don't think this happens right now, and we should do a pass on it for RTC.

juniusfree:
closes #1448 

- [x] replace block refs with content after merge-delete block - local
- [x] replace block refs with content after merge-delete block - remote
- [x] replace block refs with content after multiple selection delete - local
- [x] replace block refs with content after multiple selection delete - remote
- [x] replace multiple block refs with text content
- [x] add tests

### DIFF
--- a/src/cljc/athens/common_db.cljc
+++ b/src/cljc/athens/common_db.cljc
@@ -124,39 +124,6 @@
           eid))
 
 
-(defn replace-page-refs-tx
-  "For a given title, unlinks [[brackets]], #[[brackets]], #brackets."
-  [db title]
-  (let [pattern (patterns/linked title)]
-    (->> pattern
-         (get-ref-ids db)
-         (d/pull-many db [:db/id :block/string])
-         (mapv (fn [block]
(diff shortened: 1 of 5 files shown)


neotyk:
With test, FUIYOH!

fuiyoh!

This one has audio: <https://giphy.com/clips/mrnigelng-cooking-show-uncle-roger-ZYgJInfriKUufmFoZf>

PR #1587, opened 2021-08-24
docs: add rtc deployment ADR

filipesilva:


doc/adr/0009-rtc-deployment.md
```diff
@@ -0,0 +1,83 @@
+# 5. RTC deployment
```
neotyk, line 1: Why the `5.` here?

filipesilva: bad templating

PR #1595, opened 2021-08-25
fix: bullet too small

juniusfree:
fix https://discord.com/channels/708122962422792194/708124073544122399/880029228458975233

### DIFF
--- a/src/cljs/athens/views/blocks/bullet.cljs
+++ b/src/cljs/athens/views/blocks/bullet.cljs
@@ -30,8 +30,8 @@
    :color (style/color :body-text-color :opacity-low)
    ::stylefy/manual [[:svg {:transform "scale(1.001)" ; Prevents the bullet being squished
                             :overflow "visible" ; Prevents the stroke from being cropped
-                            :width "0.25em"
-                            :height "0.25em"}]
+                            :width "0.5em"
+                            :height "0.5em"}]
                      [:circle {:fill "currentColor"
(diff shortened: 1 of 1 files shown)


filipesilva:
Followup fix for #1495

src/cljs/athens/views/blocks/bullet.cljs
```diff
@@ -30,8 +30,8 @@
    :color (style/color :body-text-color :opacity-low)
    ::stylefy/manual [[:svg {:transform "scale(1.001)" ; Prevents the bullet being squished
                             :overflow "visible" ; Prevents the stroke from being cropped
-                            :width "0.25em"
-                            :height "0.25em"}]
+                            :width "0.5em"
+                            :height "0.5em"}]
```
shanberg, line 34:
```suggestion
                            :width "0.375em"
                            :height "0.375em"}]
```

shanberg (changes requested):
I had to go back to the mockups to recall the exact intended size after seeing it so many different ways. The size of the anchor on a regular block at the base zoom level is 6px. It looks like the first fix made the anchor too small, your fix made the anchor slightly too big, and the `0.375em` size should make it the correct size.

With my suggestion:
CleanShot 2021-09-02 at 08 18 41@2x

The mockup/source of truth:
CleanShot 2021-09-02 at 08 19 38@2x

I'm hopeful that certain unrelated front-end refactoring will make this kind of issue easier to identify and fix in the future.

shanberg (approved):

PR #1601, opened 2021-08-28
feat: notify when client is disconnected from server

### LINKED ISSUE #1598: Let user know that Lan-Party is over/suspended
**What?**

Communicate to user that Athens client is disconnected from Lan-Party.

**Why?**

When we loose connection with Lan-Party player is unaware of it.
We should let the player know that Athens client is disconnected from Lan-Party.

**How?**

*Find 3 Solutions*

juniusfree:
closes #1598 

The summary of the decisions made about this issue can be found here https://github.com/athensresearch/athens/issues/1598#issuecomment-907512172

### DIFF
--- a/src/cljs/athens/views/pages/core.cljs
+++ b/src/cljs/athens/views/pages/core.cljs
@@ -50,6 +50,10 @@
 
 (defn view
   []
+  ;; This is temporary
+  ;; TODO: create a UI to inform the player of the connection status
+  (when (= @(rf/subscribe [:connection-status]) :reconnecting)
+    (js/alert "Oops! Connection Lost. Reconnecting..."))
   (let [route-name (rf/subscribe [:current-route/name])]
     [:div (stylefy/use-style main-content-style
                              {:on-scroll (when (= @route-name :home)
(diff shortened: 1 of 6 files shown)


src/cljs/athens/views/pages/core.cljs
```diff
@@ -51,6 +51,9 @@
 (defn view
   []
   (let [route-name (rf/subscribe [:current-route/name])]
+    ;; TODO: create a UI to inform the player of the connection status
+    (when (= @(rf/subscribe [:connection-status]) :reconnecting)
+      (rf/dispatch [:alert/js "Oops! Connection Lost. Reconnecting..."]))
```
neotyk, line 56: will this display alert every few seconds when we try to reconnect?

neotyk: It's good for now.

neotyk (approved):
well done

src/cljs/athens/views/pages/core.cljs
juniusfree: > will this display alert every few seconds when we try to reconnect?

No, it's not.

PR #1606, opened 2021-09-01
refactor: deep merge config_edn

### LINKED ISSUE #1584: load athens config from env var
All parameters needed by the Athens server should be configurable via a `config.edn` file, whose location is specified through a env var for ease of use with docker.

juniusfree:
closes #1584

### DIFF
--- a/README.self-hosted.md
+++ b/README.self-hosted.md
@@ -18,7 +18,7 @@ This will create `target/athens-lan-party-standalone.jar`.
 ### Custom Datahike DB path
 
 You can customize where Datahike stores it's DB via
-environment variable `datahike_store_path`.
+environment variable `config_edn`.
 
 The default in `config.edn` is right now `/tmp/example`,
 so if you don't overwrite with above mentioned ENV var,
--- a/README.self-hosted.md
+++ b/README.self-hosted.md
@@ -15,15 +15,16 @@ lein uberjar
 
 This will create `target/athens-lan-party-standalone.jar`.
 
-### Custom Datahike DB path
+### Custom Config
 
-You can customize where Datahike stores it's DB via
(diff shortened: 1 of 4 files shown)


src/clj/athens/self_hosted/components/config.clj
```diff
@@ -13,8 +13,11 @@
 
   (start
     [component]
-    (log/info "Starting configuration component")
-    (assoc component :config (cfg/reload-env)))
+    (let [config        (cfg/reload-env)
+          merged-config (cfg/merge-maps config (:config-edn config))]
+      (log/info "Starting configuration component")
+      (log/info merged-config)
```
filipesilva, line 19: Nit: this isn't needed anymore, the final config is already printed elsewhere.

filipesilva (approved):
Lgtm, but would like @neotyk opinion on removing the previous convir var. Think this is an overall better approach.

README.self-hosted.md
```diff
@@ -18,7 +18,7 @@ This will create `target/athens-lan-party-standalone.jar`.
 ### Custom Datahike DB path
 
 You can customize where Datahike stores it's DB via
-environment variable `datahike_store_path`.
+environment variable `config_edn`.
```
neotyk, line 21: That's not just DB path anymore, maybe worth mentioning it.

neotyk (approved):

src/clj/athens/self_hosted/components/config.clj
juniusfree: There is no other `log/info` for config other than this.

filipesilva: Thought I had seen repeat prints while we were debugging, lgtm then

PR #1607, opened 2021-09-01
fix: error when deleting a block with block ref

### LINKED ISSUE #1603: fix: error when deleting a block with block ref
I'm expecting that the block should be deleted when backspace is pressed for a block with block ref but the whole text content is removed.

delete-block

It seems related to the linkmaker but this could also be just an effect. This is what linkmaker returned.
`[[:db/retractEntity 280] {:db/id [:block/uid "e2138185b"], :block/string "", :block/children []} {:db/id 225, :block/children #{}} [:db/add nil :block/refs [:block/uid "f43cf5ad6"]]]`

Removing `[:db/add nil :block/refs [:block/uid "f43cf5ad6"]` in the transaction fixes the issue.

juniusfree:
fixes #1603

### DIFF
--- a/src/cljc/athens/common_db.cljc
+++ b/src/cljc/athens/common_db.cljc
@@ -530,6 +530,16 @@
   input-tx)
 
 
+(defn update-refs
+  "Returns updated refs for eid.
+   Returns nil if eid is no longer in db."
+  [db [eid string]]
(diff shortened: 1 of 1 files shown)


filipesilva (approved):
LGTM, thanks Junius!

PR #1617, opened 2021-09-04
fix: broadcast username on update

### LINKED ISSUE #1514: Username change not updated (1.0.0-ALPHA.RTC.1 - Remote DB)
**Problem**
When I change my username in the settings, it won't be updated (for me and for every other user of the remote DB) until I disconnect then reconnect on the remote DB.

**Athens Version**
1.0.0-ALPHA.RTC.1 - Intel OSX

juniusfree:
- This PR fixes #1514.
- This will now allow the username to be updated to every client in the lan-party when one of the clients updated the user name.
- A new event to send the new user is created in the presence module on the client-side. This event will be triggered when the username on the settings page is updated.
- A new event is added on the server-side to broadcast the new user name.
- On the server-side, the atom, `athens.self-hosted.clients/clients` is also updated when the user name is changed. This is to ensure that the newly connected clients will get the latest list of user names.
- ~~On the server-side, I'm also thinking if the `editing-handler` and `username-handler` can be generalized but I'm unsure yet on how to do it.~~

### DIFF
--- a/src/clj/athens/self_hosted/web/presence.clj
+++ b/src/clj/athens/self_hosted/web/presence.clj
@@ -18,7 +18,7 @@
 (def supported-event-types
   #{:presence/hello
     :presence/editing
-    :presence/username
+    :presence/rename
     :presence/goodbye})
 
 
@@ -59,15 +59,15 @@
         (common-events/build-event-accepted id max-tx)))))
 
 
-(defn username-handler
+(defn rename-handler
   [datahike channel {:event/keys [id args]}]
   (let [{:keys
          [current-username
           new-username]}         args
         max-tx                   (:max-tx @datahike)
-        broadcast-username-event (common-events/build-presence-broadcast-username-event max-tx
-                                                                                        current-username
-                                                                                        new-username)]
+        broadcast-username-event (common-events/build-presence-broadcast-rename-event max-tx
+                                                                                      current-username
+                                                                                      new-username)]
 
     (clients/add-client! channel new-username)
     (clients/broadcast! broadcast-username-event)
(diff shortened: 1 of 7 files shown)


src/clj/athens/self_hosted/web/presence.clj
```diff
@@ -18,6 +18,7 @@
 (def supported-event-types
   #{:presence/hello
     :presence/editing
+    :presence/username
```
neotyk, line 21: `rename` might be a better name here.

src/clj/athens/self_hosted/web/presence.clj
```diff
@@ -58,6 +59,21 @@
...
+  [datahike channel {:event/keys [id args]}]
+  (let [{:keys
+         [current-username
+          new-username]}         args
+        max-tx                   (:max-tx @datahike)
+        broadcast-username-event (common-events/build-presence-broadcast-username-event max-tx
+                                                                                        current-username
+                                                                                        new-username)]
+
+    (clients/add-client! channel new-username)
```
neotyk, line 72: should we be adding a client?
This client already exists and it's renaming.

neotyk (changes requested):
2 comments, 1st about naming, 2nd about actual usage in backend.

src/clj/athens/self_hosted/web/presence.clj
juniusfree: It works but the function should be renamed.

juniusfree:
Other todos but not for this PR:
1. rename `clients/add-client!` function
2. consider how renaming will handle failures (e.g. name clash with other clients, server failure)

src/clj/athens/self_hosted/web/presence.clj
juniusfree: Updated to `rename`

PR #1629, opened 2021-09-08
fix: unlinked refs are lost from the view when one unlinked ref is linked

### LINKED ISSUE #1624: fix: unlinked refs are lost from the view when one unlinked ref is linked
Steps to reproduce
1. Go to a page with an unlinked ref or create an unlinked ref
2. Link the unlinked ref

Expectation: The unlinked ref will go to the linked refs after linking but the other unlinked refs should still be in the unlinked refs section

**Screenshots/Demo**
linking-unlinked-refs

**Athens Version**
1.0.0-BETA.96

juniusfree:
1. Fixes #1624
2. `unlinked-ref-el` uses `r/track` to sync with datascript. Before this update, the `node-page-el` is maintaining a separate `r/atom` for the unlinked references.
3. `unlinked-ref-el` is converted to form-2 component because of the update as mentioned in point no. 2.
4. Another considered solution is to fix the part where the `r/atom` for unlinked refs is getting updated but it'll add more lines of codes.


src/cljs/athens/views/pages/node_page.cljs
```diff
@@ -430,66 +430,51 @@
...
-                                                                                                             %))
-                                                                        seq)]
-                                             ;; ctrl-z doesn't work though, because Unlinked Refs aren't reactive to datascript.
-                                             (reset! unlinked-refs new-unlinked-refs)
-                                             (dispatch [:unlinked-references/link block title])))}
-                       "Link"])]))]))])])))
+  [_ title]
+  (let [unlinked-refs (r/track get-unlinked-references (escape-str title))]
+    (fn [state _]
+      (let [is-unlinked-ref? (:is-unlinked-ref? @state)]
```
neotyk, line 436: Clojure convention is to skip `is-` part, `?` makes it a predicate already.

neotyk (changes requested):
Just 1 comment about naming.

src/cljs/athens/views/pages/node_page.cljs
juniusfree: Updated
https://media3.giphy.com/media/cdNSp4L5vCU7aQrYnV/giphy.gif?cid=36b14fac1o34mrqauoe0alswpj5s8o3ku84ml5q6nhck7j9q&rid=giphy.gif&ct=g

filipesilva: @juniusfree I think Alex was asking for you to change it the other way around: from `is-unlinked-ref` to `unlinked-ref?`,

PR #1641, opened 2021-09-09
fix: import react-dom instead of using missing global

filipesilva:
Fix https://github.com/athensresearch/athens/pull/1564#discussion_r704963708


juniusfree (approved):
It's working. 👍

PR #1642, opened 2021-09-09
Fix: escape hatch when connecting to remote

neotyk:
And notify user with New DB creation, for now, until we have something
decent to present to a user.

### DIFF
--- a/src/cljs/athens/electron/db_modal.cljs
+++ b/src/cljs/athens/electron/db_modal.cljs
@@ -16,7 +16,6 @@
     [athens.util :refer [js-event->val]]
     [athens.views.modal :refer [modal-style]]
     [athens.views.textinput :as textinput]
-    [cljsjs.react.dom]
     [clojure.edn :as edn]
     [datascript.core :as d]
     [komponentit.modal :as modal]
--- a/src/cljs/athens/electron/db_modal.cljs
+++ b/src/cljs/athens/electron/db_modal.cljs
@@ -1,7 +1,6 @@
 (ns athens.electron.db-modal
   (:require
     ["/components/Button/Button" :refer [Button]]
-    [cljsjs.react.dom]
     ["@material-ui/icons/AddBox" :default AddBox]
     ["@material-ui/icons/Close" :default Close]
     ["@material-ui/icons/Folder" :default Folder]
@@ -16,6 +15,7 @@
     [athens.util :refer [js-event->val]]
     [athens.views.modal :refer [modal-style]]
     [athens.views.textinput :as textinput]
+    [cljsjs.react.dom]
     [clojure.edn :as edn]
     [datascript.core :as d]
     [komponentit.modal :as modal]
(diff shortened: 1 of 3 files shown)


src/cljs/athens/electron/db_modal.cljs
```diff
@@ -15,6 +15,7 @@
     [athens.util :refer [js-event->val]]
     [athens.views.modal :refer [modal-style]]
     [athens.views.textinput :as textinput]
+    [cljsjs.react.dom]
```
filipesilva, line 18: This isn't needed after https://github.com/athensresearch/athens/pull/1641 was merged. I think the approach in that PR is the correct one because it doesn't introduce side effects (i.e. this adds a global to window, and it's easy to rely on it by accident on other ns).

src/cljs/athens/events/remote.cljs
```diff
@@ -32,6 +32,17 @@
...
 
 
+(rf/reg-event-fx
+  :remote/connection-failed
+  (fn [_ _]
+    (js/console.log ":remote/connection-failed")
+    {:fx [[:dispatch-n [[:loading/unset]
+                        [:conn-status :disconnected]
+                        [:db/sync]
+                        ;; THIS was not recommendation from Stu
```
filipesilva, line 42: Bad UX leads to disapproving looks from @shanberg, disapproving looks like to mulling about a better UX, mulling leads to prototypes. Yes yes.

src/cljs/athens/self_hosted/client.cljs
```diff
@@ -38,6 +38,8 @@
 
 
 (def ^:private reconnect-timer (atom nil))
+(def ^:private MAX_RECONNECT_TRY 2)
+(def ^:private reconnect-counter (atom -1))
```
filipesilva, line 42: Nit: surprising initial value, expected it to be 0 or nil. -1 sounds like an artifact of the comparisons rather than the real value of the counter. e.g. if the impl changes this number stops making sense.

filipesilva (approved):
Only async non-blocking comments but would like the `cljsjs.react.dom` change to be removed.

src/cljs/athens/electron/db_modal.cljs
neotyk: removed

src/cljs/athens/events/remote.cljs
neotyk: Psst @shanberg

src/cljs/athens/self_hosted/client.cljs
neotyk: It's just so I didn't have to do `<=` comparison instead of `<` and allowed me to print amount of reconnects that have happened.

neotyk: @filipesilva are you ok with it staying like this?

filipesilva: Yeap am ok. I noticed it and wanted to mention only as a nit (nitpick). Whenever I put nits I mean it as something I noticed but that I think is up to the author.

PR #1644, opened 2021-09-10
fix: deleted daily page is blank to other clients

### LINKED ISSUE #1623: fix: deleting a page leave the other clients with a blank page
Steps to replicate:
1. Run the server
2. Connect at least 2 clients
3. Using one of the clients, delete a page
4. Observe the page where the other clients are

Screenshot
image
image

juniusfree:
Fixes #1623

Changes:
1. Added the `d/listen!` fn to let the daily pages know if the datascript db is updated.
2. Added the `:daily-note/remove` event to remove the deleted daily note from daily notes page.

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -6,7 +6,8 @@
     [clojure.string :as string]
     [datascript.core :as d]
     [posh.reagent :refer [posh! pull q]]
-    [re-frame.core :refer [dispatch]]))
+    [re-frame.core :refer [dispatch]]
+    [reagent.core :as r]))
 
 
 ;; -- Example Roam DBs ---------------------------------------------------
@@ -220,6 +221,12 @@
 
 
 (defonce dsdb (d/create-conn schema))
+
+
+(defonce tx-update (r/atom {}))
+
+
+(d/listen! dsdb :tx-update #(reset! tx-update %))
 
 
 ;; todo: turn into an effect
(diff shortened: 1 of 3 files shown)


src/cljs/athens/db.cljs
```diff
@@ -222,6 +223,12 @@
 (defonce dsdb (d/create-conn schema))
 
 
+(defonce tx-update (r/atom {}))
+
+
+(d/listen! dsdb :tx-update #(reset! tx-update %))
```
filipesilva, line 229: I'm not sure this is the right approach, because:
- it keeps a toplevel ratom to a tx-report, which will retain in memory the last db state
- it's not clear what the behaviour is when changing databases

src/cljs/athens/events.cljs
```diff
@@ -457,6 +457,12 @@
     (update db :daily-notes/items (comp rseq sort distinct conj) uid)))
 
 
+(reg-event-db
```
filipesilva, line 460: There's already a `:daily-note/delete`, is this different?

filipesilva (changes requested):
I think the right set of changes is not to add more external tracking, but rather to call the existing `:daily-note/delete` event when processing a page deletion event. 

The reason why this isn't a bug on local dbs is that the page deletion button does the behaviour I described:

https://github.com/athensresearch/athens/blob/0474e5aefe0bbef9e99b08efaa39278e1702f533/src/cljs/athens/views/pages/node_page.cljs#L359-L361

PR #1645, opened 2021-09-10
fix: all pages is slow when a page has 100+ blocks

### LINKED ISSUE #1638: Viewing all pages is slow if there is one page with 100's of biocks
One can experience this in our current Athens Public server

juniusfree:
Fixes #1638 

Changes:
1. Removed the "Body" column because it's the main cause of the slow performance. But @shanberg needs to look at this. 🕶
2. Moved the query fn to `common-db`
3. Removed the `preview-body` fn because it's not needed

### DIFF
--- a/src/cljc/athens/common_db.cljc
+++ b/src/cljc/athens/common_db.cljc
@@ -413,6 +413,15 @@
        (get-data db)))
 
 
+(defn get-all-pages
+  [db]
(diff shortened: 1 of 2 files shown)


filipesilva (approved):

filipesilva:
LGTM, thanks Junius!

PR #1654, opened 2021-09-17
Update style variables

shanberg:
Replaces the core styles with the new one from the TS library. Not a complete changeover yet but a unit of work.

Should appear exactly the same, except in the Presence menu where this change fixes some visual bugs.

- Removes cljs base style inclusions (style variables, html and body styles, etc.), replacing them with TS equivalents.
- Slightly changes the way the theme colors are applied to the DOM, mostly moving it into a classname applied near the root.
  - app-toolbar still subscribes to the theme, because it’s using an older form of styling that isn't able to respond to classes on higher components. Everything else just inherits the current theme from the DOM.

Considerations for later:
- Refactor out the `color` function to eliminate the code duplication between the remaining CLJ styles and TS styles, or only do that once we've replaced all other components?

### DIFF
--- a/src/cljs/athens/core.cljs
+++ b/src/cljs/athens/core.cljs
@@ -21,8 +21,7 @@
     [goog.dom :refer [getElement]]
     [goog.object :as gobj]
     [re-frame.core :as rf]
-    [reagent.dom :as r-dom]
-    [stylefy.core :as stylefy]))
+    [reagent.dom :as r-dom]))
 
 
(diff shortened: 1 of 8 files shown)


filipesilva (approved):

filipesilva:
@tangjeff0 I needed some of these changes for the block presence so reviewed it myself.

tangjeff0:
Np, I didn't need to review it, just wanted it considered

PR #1682, opened 2021-09-27
refactor: use Datalog Console library

groundedSAGE:
This removes the raw code to interface with the Datalog Console and replaces it with a library. It also places the Datalog Console behind a dev flag. We are currently doing a lot of work for the V1 release of the Datalog Console which includes better security, at which point we will consider it safe to turn this back on for end users.


filipesilva (approved):
Thanks for updating this @groundedSAGE, and sorry it took so long for us to take a look! It might be a while until it's reflected on https://athensresearch.github.io/athens/ since that domain is tracking the 1.0.0 release and `main` is releasing `2.0.0`, but it should be deployed there when 2.0.0 is the stable one.

PR #1690, opened 2021-09-28
feat(button): buttons have styled focus ring

shanberg:
- Add focus ring to buttons
- Focus ring only shows during keyboard interaction

https://user-images.githubusercontent.com/98312/135023115-5a1c9ced-68be-43bc-a616-bb197cf8a780.mp4


filipesilva (approved):
This looks very slick, keyboard navigation ftw

PR #1695, opened 2021-09-28
ci: ping alpha tester on new release (proposal)

MeerBiene:
The idea was to have a way to automatically ping users with a certain role from the github workflow.

Would need some beforehand setup in discord:
- create webhook url
- invite a reaction roles bot and set it up so users can give themselves different roles, e.g. "alpha tester", "beta tester" ... etc
- make a seperate channel that is only viewable by these roles. That way only those roles see the spam by this new webhook

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -209,7 +209,7 @@ jobs:
       #       curl \
       #       -H "Content-Type: application/json" \
       #       # how to ping roles manually: https://support.discord.com/hc/en-us/community/posts/360057835772-Role-Pinging-for-Bots
-      #       -d '{"username": "athens", "content": "Hey <@&${{ secrets.discord_alpha_role }}> a new release has been published \n\n[Go check it out!](https://github.com/athensresearch/athens/releases)"}' \
+      #       -d '{"username": "Athens", "avatar_url": "https://github.com/athensresearch/athens-assets/raw/main/athens-512x512.png", "content": "Hey <@&${{ secrets.discord_alpha_role }}> a new release has been published \n\n[Go check it out!](https://github.com/athensresearch/athens/releases)"}' \
       #       ${{ secrets.discord_webhook }}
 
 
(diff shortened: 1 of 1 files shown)


filipesilva:
This would be a bit similar to having a releases channel on discord with a github bot or something, right?

MeerBiene:
Exactly. 
My first approach was to hack a quick bot together that would receive post requests and then ping the required roles. Then i remembered webhooks exist so i tried to ping roles with a webhook and since it worked this was the simplest solution I came up with. 

I don't know if there are any github bots and how they handle releases but I'm afraid of a bot that would just  query the github API every now and then.

However if you would prefer a bot i could offer to hack something together but it would be in TS then :sweat_smile:

filipesilva:
I kinda expected there to be a super obvious free (hosted) github bot for discord around, but doesn't seem like it! Webhooks like you propose would be a much better approach than trying to host our own thing then.

Regarding the roles and messages, is it important to be a role+channel thing? I didn't quite get the importance of roles if the channel was already there. It sounded like roles+channels were a sort of double opt-in. Would it be enough to just post the message to the channel?

Another important thing to keep in mind is that the releases are created as drafts, and promoting them to a real release is a manual thing. So this action would need to be triggered on a release going from draft->non-draft, which I think is part of the github api anyway.

MeerBiene:
The roles+channel thing is for people like me with the memory of a goldfish. I simply forget to look for a new release so i really wanted to be pinged. That's why i thought about reaction roles, so only the ppl that explicitly opted in to getting ~~annoyed~~ pinged are getting ~~annoyed~~ pinged. The channel that's only visible to them is so the discord server doesn't get cluttered for the "normal" users.

About the draft-> release thing: I think the best way would be to use "release" as trigger then right?

filipesilva:
@MeerBiene this PR was closed automatically because we merged `feature/rtc-v1` into `main`. Can you open it again against `main` please?

PR #1698, opened 2021-09-29
build: use data env var for data, logs path

filipesilva:


docker-compose.yml
```diff
@@ -10,12 +10,13 @@ services:
     expose:
       - 3010 # Change this according to config.edn
     volumes:
-      - datahike-db:/srv/athens/db
-      - crash-logs:/srv/athens/logs
+      # Use bindings for volumes if provided.
+      - ${DATA_PATH:-datahike-db}:/srv/athens/db
```
neotyk, line 14: What happens with volumes defined at the end of this file if you customise here via env?

filipesilva: I thought they would still be created, but they actually aren't. If nothing uses them, they aren't created.

But if you used them once, and then stopped using them via the env file, they were already created and you'd have to remove them yourself.

neotyk: there is also this pattern of binding local directory to container.
It this scenario we'd have 2 folders where docker-compose is (db, logs).
and we would just mount these directories to inside container.

filipesilva: I would be a bit more restrictive but maybe a bit more intuitive. Maybe harder to use another drive for data? No strong opinion on it. We could have some `BIND_TO_LOCAL_DIR=true` flag I guess. 

I'm not sure how to do the relative paths though, so if you'd like to do this I'd appreciate an example.

neotyk: Like dat:
```yaml
- ./data/minio_root:/minio_root
```

filipesilva: Straightforward.

Maybe we should never use volumes at all then? And always bind to the dir. I'm having trouble seeing the case where we actually want the user to do anything at all with the volumes, and it's definitely useful to have the files somewhere you know.

Under this alternate setup, you download the docker compose, start it, and there's an `athens-data` folder or something with everything athens stores (db, logs, resources, a penguin, etc).

WDYT?

neotyk: Yeah, I like this setup too.
Only problem is that you need to have these folders available, they won't get created for you by docker compose.
Which creates friction.

filipesilva: They are created. So it's a go!

PR #1706, opened 2021-10-04
feat(notification): add notification component

shanberg:
Probably actually a 'toast' component but I can't bring myself to call it that. Once we get into OS-level notifications maybe I'll cave.

Preview: https://athens-git-fork-shanberg-toast-component-athens-research.vercel.app/?path=/story/components-notification--active

- Implements `react-hot-toast` library locally as `notify`.
- Adds NotificationContainer and NotificationItem components
- Extends `react-hot-toast` with some goodies:
    -  `isDismissable` provides `x` button to dismiss notification. Should usually be paired with `duration: Infinity`
    - `onUndo` accepts a fn and provides `Undo` button. Should usually be paired with an `undoMessage` to show when undo occurs.

Todos
- [ ] Properly extend module `notify` interface to avoid typescript complaints about `Toast` type


filipesilva (approved):
image

filipesilva:
err actually it's still a draft so no merge?

PR #1717, opened 2021-10-08
build: move from lein to clj+deps.edn

filipesilva:
Lein with deps.edn does not work on Windows, so we're moving to clj+deps.edn.

See RickMoynihan/lein-tools-deps#85


l2dy:
Docs need to be updated. https://github.com/athensresearch/docs/blob/main/docs/developer_guide/running.md

filipesilva:
@l2dy thanks for the heads up! Update inc https://github.com/athensresearch/docs/pull/4

Issue #560, opened 2021-01-21
fix: pressing delete twice on block incorrectly joins blocks together


tangjeff0:
**Problem**

Pressed `delete` on selected text twice joins the existing text together.

**Screenshots**

[ Particularly helpful for UI bugs. ]

https://user-images.githubusercontent.com/8952138/105416205-a85ddd80-5bee-11eb-86cd-dd5c398216ef.mp4

datokrat:
The cause of this bug is [here](https://sourcegraph.com/github.com/athensresearch/athens@9988417/-/blob/src/cljs/athens/views/blocks/textarea_keydown.cljs?subtree=true#L690:7): `handle-delete` merges the two blocks by joining their strings together. The current textarea state is ignored. One would somehow have to pass the local block state ([see](https://sourcegraph.com/github.com/athensresearch/athens@9988417/-/blob/src/cljs/athens/views/pages/node_page.cljs?subtree=true#L535:11) `node-page-el`) to `handle-delete`.

PR #1719, opened 2021-10-09
Fix the issue with merging blocks using the delete key

hustodemon:
Fixes https://github.com/athensresearch/athens/issues/560. I've got hit by it, so I'm proposing a fix.

To sum up:

The problem happened when:
1. there are 2 consecutive non-empty blocks
2. user changes the first one (e.g. deletes a character)
3. user moves the cursor to the end of the first block and hits the
   delete key

Instead of merging visible the blocks, the first block gets "reverted"
to the old state (as if the 2nd step didn't happen).

Fixed by transacting the block to the db before merging the blocks.

### DIFF
--- a/src/cljs/athens/views/blocks/textarea_keydown.cljs
+++ b/src/cljs/athens/views/blocks/textarea_keydown.cljs
@@ -704,6 +704,9 @@
         next-block-uid            (db/next-block-uid o-uid)]
     (when (and no-selection? end? next-block-uid)
       (let [next-block (db/get-block [:block/uid (-> next-block-uid db/uid-and-embed-id first)])]
+        ;; save the state of this block before merging with the next one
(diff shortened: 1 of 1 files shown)


filipesilva:
@hustodemon this PR was closed automatically because we merged `feature/rtc-v1` into `main`. Can you open the changes you had against `main` please?

hustodemon:
Sure thing! I'll try to rebase and will create a new PR later (same goes for my other PR).

filipesilva:
I think you don't need to rebase, you can just go to your branch on github and click `contribute`:

https://github.com/hustodemon/athens/tree/fix-merging-blocks-with-delete 

image

hustodemon:
Oh, I didn't know that feature. Thanks for the tip @filipesilva

Issue #1718, opened 2021-10-08
Demosite down


letharion:
**Problem**
Following the link to the demo (https://athensresearch.github.io/athens) from the main github page, initial page load just yields:

```
message=Error: No matching clause: 
source=https://athensresearch.github.io/athens/js/compiled/app.js
lineno=8398
colno=15
error=Error: No matching clause: 
```

hustodemon:
To be fair, I experienced this issue the 1st time I tried athens, too... So I created a PR proposal for basic warnings in case the browser is unsupported #1720.

aadilayub:
Is there a missing feature or something that prevents the app from working on Firefox?

D-side:
I second @aadilayub, it was working just fine in Firefox a couple weeks ago, and now with #1876 it now flat out refuses to.
Would it perhaps be a better idea to revert it for now and consider a less disruptive and more explanatory warning in its place?
With a message along the lines of:
> You are using a browser that might not fully support this live demonstration of Athens. Since the app is built with Electron, it's only expected to run well on Chrome and other Chromium-based browsers. But browser support is not currently a goal for Athens, it's meant to be used as a standalone application.

hustodemon:
Hey guys, for me it never worked with Firefox and since it's an electron app, I supposed it can't even work on non-chrome-based browsers. But I might be wrong. In any case, I think it'd be great if an Athens developer could shed more light on this. Perhaps @filipesilva ?

D-side-BL:
Well, it did 🙂
So I think we're looking at a slight misdiagnosis here at the very least.
Followed up by a good fix, just for an unconfirmed problem. 🤷

At a glance given this symptom I can only hypothesize that there's a `case` form somewhere that didn't produce a match ([here's where this is defined](https://github.com/clojure/clojurescript/blob/r1.10.914/src/main/clojure/cljs/core.cljc#L2368)), so at least the origin is in the realm of ClojureScript. But where and whether it's even in Athens, I didn't get much further, so ideally it would be great to reproduce this problem and investigate the stack trace. I'd love to try, I've done some Clojure before, but my browser was totally okay with the past versions of the live demo, so I don't think I'd get the error in the first place 🤔
…
What I didn't try back then though, is a clean Firefox with no extensions, but I don't think I'm using any that might make it look more like Chrome. And I doubt anything of that caliber can easily bridge browser incompatibilities, though that'd be cool 😅

[...]

Issue #1705, opened 2021-10-04
Demo linked to in github sidebar doesn't work for me.


pjboro:
Hi! I just wanted to try out the app, but it hasn't worked under Firefox 92.0 (64-bit) on Ubuntu so I wanted to let you know.

**Problem**
The demo mentioned in the github side panel does not open. 

**Screenshots/Demo**
![image](https://user-images.githubusercontent.com/42270745/135778077-1ef6de0e-4844-4020-95a7-15dc5a58dcfc.png)

**Athens Version**
It wasn't provided. I looked in the page source but also didn't find it - I may have missed it.

I can provide more information - just say what you need :)

PR #1720, opened 2021-10-10
Basic warnings for unsupported browsers

hustodemon:
1. Display a warning when JS is disabled in the browser (previously the  browser just displayed a blank page). Tested with the links browser as well with chromium with JS disabled.
2. Display a warning for browsers other than chrome/chromium (previously a JS error popup was displayed). See #1718 and #1705. Tested on Chromium Version 92.0.4515.159 (openSUSE Build), Opera 79.0 (warning not displayed) and Firefox 78.12 (warning displayed).

It's nothing fancy: there is no CSS styling applied to the warnings, but I believe it's still better than the current status.

Firefox:
ss1

Chrome w/o JS:
ss2

### DIFF
--- a/resources/public/index.html
+++ b/resources/public/index.html
@@ -10,6 +10,9 @@
       <link rel="stylesheet" href="./css/codemirror.css" type="text/css" />
       <style id="_stylefy-constant-styles_"></style>
       <style id="_stylefy-styles_"></style>
+      <noscript>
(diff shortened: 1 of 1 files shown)


filipesilva:
@hustodemon this PR was closed automatically because we merged `feature/rtc-v1` into `main`. Can you open the changes you had against `main` please?

PR #1733, opened 2021-10-14
Fixed regex patterns for block embed and block ref

sid597:
@neotyk @filipesilva How do I do this is a better way ?


filipesilva:
Hm I think I merged this to quickly, and left a carve error because I didn't understand what was happening. Following up on the rtc branch.

PR #1786, opened 2021-11-04
build: separate arm64 ci build

filipesilva:


filipesilva:
Also noticed the yarn cache is never hit, this happens because of https://github.com/actions/cache/issues/296#issuecomment-634780932.

> Two workflows triggered from tags won't be able to share caches (same scoping issue I mention above), but can read from caches created by your default branch.

Since we haven't updated our default branch in a while, our windows and mac yarn installs take about 5 minutes. 5 minutes seems like a lot still, but it really seems to be a cache problem.

neotyk (approved):

PR #1788, opened 2021-11-05
add-child using atomic op

sid597:
Cannot remove code related to `add-child`  semantic event because we are using it for `block-open-add-child` once we move this one to use atomic op then all the code should be removed.


neotyk:
Nice

PR #1789, opened 2021-11-05
enhance: navigate to pages by page title, not uid.

neotyk:


src/cljs/athens/views/pages/page.cljs
```diff
@@ -7,6 +7,18 @@
...
 
+(defn page-by-title
+  []
+  (let [title           (rf/subscribe [:current-route/page-title])
+        {:keys [node/title
+                block/string
+                db/id]} @(pull db/dsdb '[*] [:node/title @title])]
+    (cond
+      title  [node-page/page id]
+      string [block-page/page id]
```
filipesilva, line 18: This case should never happen - if there's no title, the pull fails.

src/cljs/athens/views/pages/page.cljs
```diff
@@ -7,6 +7,18 @@
     [re-frame.core :as rf]))
 
 
+(defn page-by-title
+  []
+  (let [title           (rf/subscribe [:current-route/page-title])
+        {:keys [node/title
+                block/string
+                db/id]} @(pull db/dsdb '[*] [:node/title @title])]
```
filipesilva, line 15: Pull will throw if the node title does not exist, and crash the app. This instead should do a e-by-av lookup, which does not throw.

filipesilva (approved):

PR #1792, opened 2021-11-05
Make `unlinked-references-link` use atomic block/save ops

sid597:


src/cljs/athens/events.cljs
```diff
@@ -1259,15 +1259,27 @@
   :unlinked-references/link
   (fn [{:keys [db]} [_ {:block/keys [string uid]} title]]
```
neotyk, line 1260: It doesn't look like we're still using `uid` here, correct?

src/cljs/athens/events.cljs
```diff
@@ -1259,15 +1259,27 @@
...
   (fn [{:keys [db]} [_ unlinked-refs title]]
     (log/debug ":unlinked-references/link:" title)
-    (let [event (common-events/build-unlinked-references-link-all (:remote/last-seen-tx db) unlinked-refs title)]
+    (let [block-save-ops      (mapv
+                                (fn [{:block/keys [string uid]}]
+                                  (let [ignore-case-title (re-pattern (str "(?i)" title))
+                                        new-str           (string/replace string ignore-case-title (str "[[" title "]]"))]
+                                    (graph-ops/build-block-save-op @db/dsdb uid string new-str)))
+                                unlinked-refs)
+          link-all-op         (composite-ops/make-consequence-op {:op-type :block/save}
```
neotyk, line 1279: Instead of `:op-type :block/save` use this `:op/type :block/unlinked-refs-link-all`.

neotyk (changes requested):

src/cljs/athens/events.cljs
neotyk: yes we do 🤦‍♂️

PR #1802, opened 2021-11-09
fixed indent not saving content of block

sid597:


src/cljs/athens/events.cljs
```diff
@@ -1045,26 +1045,42 @@
     [prev-block-uid target-rel]))
 
 
+(defn block-save-block-move-composite-op
+  [source-uid ref-uid relation string]
+  (let [block-save-op             (graph-ops/build-block-save-op @db/dsdb source-uid string)
+        block-move-op             (atomic-graph-ops/make-block-move-op source-uid
+                                                                       {:ref-uid ref-uid
```
neotyk, line 1052: for safety we probably should use `compat` thing.
@filipesilva am I right?

src/cljs/athens/views/blocks/textarea_keydown.cljs
```diff
@@ -392,16 +392,19 @@
         selected-items                 @(subscribe [::select-subs/items])
         editing-uid                    @(subscribe [:editing/uid])
         current-root-uid               @(subscribe [:current-route/uid])
-        [editing-uid embed-id]         (db/uid-and-embed-id editing-uid)]
+        [editing-uid embed-id]         (db/uid-and-embed-id editing-uid)
+        local-string                   (:string/local @_state)]
```
neotyk, line 396: don't leave `_` in `_state` if we're using it now.
`_` at the beginning means not used.

neotyk (approved):

src/cljs/athens/events.cljs
filipesilva: @neotyk you are right, as you often are, I would even say always.

filipesilva (approved):

PR #1803, opened 2021-11-09
build: gh pages, auto-updates only for v1.*

filipesilva:


.github/workflows/build.yml
```diff
@@ -195,12 +196,13 @@ jobs:
...
-           # Skip auto-update files on branches other than main.
+           # Only publish auto-update files on v1.* stable.
            # Ideally we could use electron-builder release channels, but they don't work for github publishing.
-           # Auto-update is also disabled in code inside athens.main.core.
+           # Instead we control whether the auto-update is published (here) and whether the app will auto-update
+           # from releases or releases+prereleases (in athens.main.core).
            # NB: GH actions doesn't have ternary operator, but can workaround using AND/OR logic.
            # see https://github.com/actions/runner/issues/409#issuecomment-752775072
-           args: ${{ github.ref != 'refs/heads/main' && '-c.publish.publishAutoUpdate=false' || '' }} ${{ matrix.electron-arch-overrides }}
+           args: ${{ startsWith(github.ref, 'refs/tags/v1.') && '-c.publish.publishAutoUpdate=true' || '-c.publish.publishAutoUpdate=false' }} ${{ matrix.electron-arch-overrides }}
```
neotyk, line 205: This doesn't take into account suffixes, like `*-beta-rtc`, does it?

src/cljs/athens/main/core.cljs
```diff
@@ -13,10 +13,13 @@
...
 
 
-;; NB: channels don't work on github releases, instead use use prereleases.
+;; Channels don't work on github releases, instead using prereleases.
+;; Also see .github/workflows/build.yml for more context and how auto-update data is published.
 ;; https://github.com/electron-userland/electron-builder/issues/1722#issuecomment-310468372
 ;; https://github.com/electron-userland/electron-builder/issues/4988
-(set! (.. autoUpdater -channel) "beta")
+;; https://www.electron.build/auto-update#appupdater-eventemitter
+(set! (.. autoUpdater -allowPrerelease) true)
```
neotyk, line 21: I though we've decided to not autoupdate prereleases.

neotyk (commented):
Need more clarity around it

.github/workflows/build.yml
filipesilva: Correct, it just determines if update data should be published. Here we are saying "publish update data if v1.*".

src/cljs/athens/main/core.cljs
filipesilva: It won't be updated if there's no data to update it (above). But now that I think about it more, this part is redundant:

> allowPrerelease = false Boolean - GitHub provider only. Whether to allow update to pre-release versions. Defaults to true if application version contains prerelease components (e.g. 0.12.1-alpha.1, here alpha is a prerelease component), otherwise false.

The auto updater will set it to true if there's any prerelease components to the version, which is what I was doing anyway. Might as well just not do anything with it there.

neotyk (approved):

PR #1805, opened 2021-11-09
feat: atomic `:page/rename`

neotyk:


src/cljc/athens/common_events/graph/atomic.cljc
```diff
@@ -79,12 +79,12 @@
...
   "Creates `:page/rename` atomic op.
-   - `page-uid` - `:block/uid` of page to be renamed
+   - `old-name` - Page name before rename,
    - `new-name` - Page should have this name after operation"
-  [page-uid new-name]
+  [old-name new-name]
   {:op/type    :page/rename
    :op/atomic? true
-   :op/args    {:page-uid page-uid
+   :op/args    {:old-name old-name
```
filipesilva, line 87: Interesting how with blocks we can just save by string + uid, but here we 100% need old and new name (because the content is the identity).

src/cljc/athens/common_events/graph/schema.cljc
```diff
@@ -7,14 +7,14 @@
...
-   :block/open ; ✓
+   :block/new    ; ✓
+   :block/save   ; ✓
+   :block/open   ; ✓
    :block/remove ; ✓
-   :block/move ; ✓
-   :page/new ; ✓
+   :block/move   ; ✓
+   :page/new     ; ✓
+   :page/rename  ; ✓
```
filipesilva, line 16: uhh fancy

filipesilva (approved):

PR #1815, opened 2021-11-11
fix: enable atomic `:page/rename` on protocol.

neotyk:


neotyk:
[trello card](https://trello.com/c/rOSjpd7u)

filipesilva (approved):

PR #1816, opened 2021-11-11
Atomic `shortcut/new` and `shortcut/remove` op

sid597:


src/cljc/athens/common_db.cljc
```diff
@@ -67,6 +67,50 @@
...
+        (:node/title)))
+
+
+#_(defn find-source-target-title
+    [db source-order target-order]
+    (let [sidebar-elements   (get-sidebar-elements db)
+          source-title (find-title-from-order sidebar-elements source-order)
+          target-title (find-title-from-order sidebar-elements target-order)]
+      [source-title target-title]))
+
```
sid597, line 94: Above 2 functions will be used in `shortcut/move`

src/cljc/athens/common_events/resolver/atomic.cljc
```diff
@@ -379,6 +379,36 @@
     tx-data))
 
 
+(defmethod resolve-atomic-op-to-tx :shortcut/new
+  [db {:op/keys [args]}]
+  (let [{page-name :name}    args
+        page-uid             (common-db/get-page-uid db page-name)
+        reindex-shortcut-txs (->> (common-db/get-sidebar-elements db)
```
filipesilva, line 386: Why is it necessary to reindex? These should not be out of order, and we're only adding at the end anyway.

src/cljc/athens/common_events/resolver/atomic.cljc
```diff
@@ -379,6 +379,36 @@
...
+                                  (sort-by :page/sidebar)
+                                  (map-indexed (fn [i m] (assoc m :page/sidebar i)))
+                                  vec)
+        add-shortcut-tx      {:block/uid    page-uid
+                              :page/sidebar (or (count reindex-shortcut-txs)
+                                                1)}
```
filipesilva, line 392: Shouldn't this be 0? E.g. if there's one existing sidebar element, count would be 1, and the new page would be order 1. So if there's none, it should be order 0, no?

filipesilva (approved):
The two comments I left are optional because the original implementation had the same issues and this task was mostly to migrate to atomic ops.

src/cljc/athens/common_events/resolver/atomic.cljc
sid597: yeah you are right no need for reindex  here

src/cljc/athens/common_events/resolver/atomic.cljc
sid597: Yeah, programmers will say this should be 0 other people will say it should be 1. I don't why this was originally made 1 so I let it be.

filipesilva: I don't think this is just an argument of starting number - if it is left at one, the next time a page is added it will also be 1, and the order will thus be incorrect.

PR #1825, opened 2021-11-12
embed presence

filipesilva:
- rfct: reduce log noise on non-verbose settings
- fix: editing/target should delegate to editing/uid
- fix: embeds show presence as well as the original block


neotyk (approved):
LFG

PR #1826, opened 2021-11-12
fix: group-title is now nested in a vector

filipesilva:


sid597 (approved):
Tested by checking out

PR #1829, opened 2021-11-13
don't publish auto update while rtc is in alpha

filipesilva:


filipesilva:
Closed in favor of #1831

filipesilva:
Actually still needed separately.

neotyk (approved):

PR #1834, opened 2021-11-15
rfct: review athens protocol

filipesilva:


src/cljc/athens/common_events/resolver/atomic.cljc
```diff
@@ -17,15 +17,14 @@
...
   [db {:op/keys [args]}]
-  (let [{:keys [block-uid
-                position]}              args
-        {:keys [ref-uid
-                ref-name
-                relation]}              position
+  (let [{:block/keys [uid position]}    args
+        {relation  :relation
+         ref-uid   :block/uid
+         ref-title :page/title}         position
```
sid597, line 23: What is the meaning of `ref`  in these cases? 

`ref` to me at first glance seems like we are talking about the `block ref`'s ref, are we not using `ref` in multiple scenerios?

sid597 (approved):
All good, I just have one question.

src/cljc/athens/common_events/resolver/atomic.cljc
filipesilva: It's the page/block referred to in position. Events at this level almost never talk about block-refs, and the athens protocol itself doesn't know about refs either.

neotyk (approved):

PR #1836, opened 2021-11-15
fix: health-check server before connecting to websocket

filipesilva:


src/cljs/athens/effects.cljs
```diff
@@ -149,17 +149,35 @@
...
+              (if (.-ok response)
+                (success-cb)
+                (failure-cb))))
+      (catch failure-cb)))
+
+
 (rf/reg-fx
   :remote/client-connect!
-  (fn [{:keys [ws-url] :as remote-db}]
+  (fn [{:keys [url ws-url] :as remote-db}]
```
neotyk, line 164: Might be better of if we derive `url` from `ws-url`

neotyk (approved):

src/cljs/athens/effects.cljs
filipesilva: It's the other way around really - ws-url is derived from url:

https://github.com/athensresearch/athens/blob/6a4eefe0c0331cdd960c3006c6fb696cf0852fbe/src/cljs/athens/electron/utils.cljs#L61-L70

PR #1849, opened 2021-11-17
fix: better logging.

neotyk:


sid597 (approved):

test/athens/common_events/utils_test.cljc
```diff
@@ -0,0 +1,30 @@
+(ns athens.common-events.utils-test
+  (:require
+    [athens.common-events                 :as sut]
```
filipesilva, line 3: `sut`?

src/clj/athens/self_hosted/clients.clj
```diff
@@ -70,24 +71,26 @@
...
-                   ", status:" (:event/status data))
+                   (if type
+                     (str ", type: " type)
+                     (str ", status: " status)))
         (http/send! channel (->transit data)))
       ;; TODO internal failure mode, collect in reporting
       (log/error "Not sending invalid event to username:" username
                  ", event-id:" (:event/id data)
-                 ", type:" (:event/type data)
+                 ", type:" (common-events/find-event-or-atomic-op-type data)
```
filipesilva, line 85: `find-event-or-atomic-op-type` doesn't seem to have defined behaviour or tests for invalid events, and events that fall here are invalid in one way or another.

filipesilva (approved):

test/athens/common_events/utils_test.cljc
neotyk: Suit Under Test

Issue #727, opened 2021-03-03
Option to turn spell check off 


panura:
**The problem: There's no option to turn spell check off**

When Athens Research doesn't detect the language of the text, the spell checker flags every word as incorrect.

The simplest solution would be an option to turn the spell check off.

jmartinezheras:
what about the option to select proofing language? This would be convenient for me as I write both in Spanish and English

panura:
That would be great as well. But I don't really expect to get spell checking for small languages like Finnish.

tangjeff0:
See also #564 and https://www.electronjs.org/docs/tutorial/spellchecker

sid597:
As mentioned by jeff this issue is discussed in #727, so closing this.

the-jb:
#564 is mainly about right-click action of spell check.
But I think this option is needed even if right-click action is added, because of appearance issue when reading.
Someone just want to disable this feature, leaving just misspelled words one intended.

So, I think this issue needed to be re-open, dealing separately from #564.

PR #1856, opened 2021-11-19
disable spellcheck

Malmik:
This is confirmed working on my local build
Would be a temproary workaround for #727

I've read code of conduct of this repo, also [this](https://handbook.athensresearch.org/community/get-involved/setup) and I think this pr is fine, but it's my first one so sorry for any mistakes.

### DIFF
--- a/src/cljs/athens/main/core.cljs
+++ b/src/cljs/athens/main/core.cljs
@@ -76,7 +76,8 @@
                                                    ;; Remove OverlayScrollbars and instances of `overflow-y: overlay`
                                                    ;; after `scollbar-gutter` is implemented in browsers.
                                                    :enableBlinkFeatures 'OverlayScrollbars'
-                                                   :nodeIntegrationWorker true}})))
+                                                   :nodeIntegrationWorker true
+                                                   :spellcheck false}})))
   ;; Path is relative to the compiled js file (main.js in our case)
   (.loadURL ^js @main-window (str "file://" js/__dirname "/public/index.html"))
   (.on ^js @main-window "closed" #(reset! main-window nil))


neotyk:
- This issue is not a simple as just turning spellchecker off for everyone.
- Users without TouchBar probably don't see spellcheck suggestions.
- On my Mac I get spellcheck suggestions on TouchBar, I'd still like to get those.
- Therefore this should be a user preference, if they want spell check or not
    - This requires editing settings.
- Changing spellcheck preferences will require restarting Athens to be activated.
    - This might be trickier then it sounds at first.
    - We manage application state in `rendering` part
    - `BrowserWindow` is started from `main` part of electron app
        - This part might not have easy access to data stored by `renderer`
        - If this is the case we'd have to communicate changes to spellchecker setting to main process, store it in main and read before we create `BrowserWindow`

Malmik:
I do understand that it isn't a fix for the issue and you are right with what you said.
I would find the decision of disabling spellcheck a compromise that will not solve current situation, but will improve it.

Now, you may say that majority of Athens users are english speaking people, and while that's true it doesn't mean they take personal notes in that language. I would be pretty sure that most of Athens users with their native language different from English do make notes in their own one, so it's apparent that spellchecking in it's current form is beneficial to few and detrimental for others, often to the such a degree that usage of Athens is impossible for them.

I understand you @neotyk would be in a dissadvantageous situation after merging this pr and I see your objections, but I hope that you will agree with me, especially that we didn't have any pr that tries to implement approach you propose, using preferences in settings, for at least 8 months ( #727 ) nor any other, even less viable solution.

PR #1857, opened 2021-11-19
fluree timeout

filipesilva:
- fix: save event to log inside the same lock as db transaction
- fix: don't print whole remote-db config
- fix: timeout on health check
- feat: handle errors and reconnect on add-event!
- build: add fluree dev script

### DIFF
--- a/src/clj/athens/self_hosted/components/fluree.clj
+++ b/src/clj/athens/self_hosted/components/fluree.clj
@@ -7,7 +7,7 @@
 
 
 (defrecord Fluree
-  [config conn]
+  [config conn-atom reconnect-fn]
 
   component/Lifecycle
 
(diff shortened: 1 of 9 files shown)


src/clj/athens/self_hosted/components/fluree.clj
```diff
@@ -21,10 +21,18 @@
...
         (do
           (log/info "Starting Fluree connection, servers" servers)
-          (let [conn (fdb/connect servers)]
+          (let [;; Wrap conn in an atom so we can reconnect without restarting component.
+                conn-atom    (atom nil)
+                reconnect-fn (fn []
+                               (when-some [conn @conn-atom]
+                                 (fdb/close conn))
+                               (reset! conn-atom (fdb/connect servers)))
+                comp         {:conn-atom    conn-atom
```
neotyk, line 30: `:conn-atom` & `:reconnect-fn` should be part `defrecord`.
for now it's still having `conn` in `defrecord`

PR #1860, opened 2021-11-20
build: linux-arm64 target supported and smaller images

neotyk:


.github/workflows/build.yml
```diff
@@ -225,6 +225,9 @@ jobs:
        - name: Build server JAR
          run: yarn server:uberjar
 
+       - name: Set up QEMU
```
filipesilva, line 228: Can you add a comment about why this is needed/used for? I don't understand why it was added just from reading https://github.com/qemu/qemu .

filipesilva (approved):

.github/workflows/build.yml
neotyk: commented.
Also this: https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md

PR #1871, opened 2021-11-24
docs: ADR Athens System Design.

neotyk:


doc/adr/0017-system-design-lan-party.dot
```diff
@@ -0,0 +1,77 @@
...
+        cSE -> "Optimistic Update" [label="Apply optimistically";];
+        cSE -> SendIt [label="Send AGO over WS";];
+        "Optimistic Update" -> cResolution [label="Composed of AGO";];
+        cResolution -> cTransact;
+        cTransact -> cDatascript;
+
+        cAck -> cDatascript [label="Updates DB, cleaning Optimistic state";];
+        cDatascript -> UI [style=dotted; label="Render based on DB context"; color=blue];
+        cDatascript -> cResolution [style=dotted; label="Resolved in DB context"; color=blue];
+        cDatascript -> cSE [style=dotted; label="Creates AGO in DB context"; color=blue];
```
filipesilva, line 26: I don't understand what this arrow means. Can you explain?

doc/adr/0017-system-design-lan-party.dot
```diff
@@ -0,0 +1,77 @@
...
+        sPresence [label="Forward to all connected clients";];
+        sForward [label="Forward AGO to all connected clients";];
+        sEventStore [label="Event Store";];
+        
+        wsReceive -> isAGO;
+        isAGO -> sAGO [label="Yes, AGO";];
+        sAGO -> sResolution [label="1";];
+        sAGO -> sAck [label="2";];
+        sAGO -> sForward [label="3";];
+        sAGO -> sEventStore [label="4";];
```
filipesilva, line 48: What are the numbers here meant to convey? Is it order of operations?

If it is order, then 4 is actually the very first one and blocks the others (as it should).

doc/adr/0017-system-design-lan-party.dot
filipesilva: Thinking a bit more about it, I think it means that semantic events create AGO using the DB. Is that right? If so, then I think all blue arrows are pointing in the wrong direction.

e.g. `cDatascript` doesn't use `AGO resolution` for `Resolved in DB context`, instead `AGO resolution` uses `cDatascript`

filipesilva (approved):
lgtm, except for the two comments

doc/adr/0017-system-design-lan-party.dot
neotyk: State of Datascript DB influences resolution.

doc/adr/0017-system-design-lan-party.dot
neotyk: Yes, it's order of execution, I'll update.

doc/adr/0017-system-design-lan-party.dot
neotyk: Updated

PR #1872, opened 2021-11-25
doc: expand and rename event log adr

filipesilva:


neotyk (approved):
I like what you did here!

PR #1874, opened 2021-11-25
add a simple kubenetes.yaml file

bbigras:
as discussed on discord:

﻿https://discord.com/channels/708122962422792194/800442937602342922/913545032760455212

### DIFF
--- a/kubenetes.yaml
+++ b/kubenetes.yaml
@@ -1,0 +1,86 @@
+# TODO:
+# - volumes
+# - 'depends_on' equivalent (if possible)
+#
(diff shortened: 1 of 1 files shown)


bbigras:
a traefik ingress should be something like:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: athens
  annotations:
    kubernetes.io/ingress.class: "traefik"
spec:
  rules:
  - host: athens.my_domain.com
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: athens
            port:
              number: 3010
```

PR #1875, opened 2021-11-26
fix: Fix the issue with merging blocks using the delete key

hustodemon:
The problem happened when:
1. there are 2 consecutive non-empty blocks
2. user changes the first one (e.g. deletes a character)
3. user moves the cursor to the end of the first block and hits the
   delete key

Instead of merging visible the blocks, the first block gets "reverted"
to the old state (as if the 2nd step didn't happen).

Fixed by transacting the block to the db before merging the blocks.


neotyk:
👍  Thank you @hustodemon

hustodemon:
Well, thank you for the fix @neotyk  ;)

PR #1876, opened 2021-11-26
Basic warnings for unsupported browsers

hustodemon:
1. Display a warning when JS is disabled in the browser (previously the
   browser just displayed a blank page)
2. Display a warning for browsers other than chrome/chromium (previously
   a JS error popup was displayed). See https://github.com/athensresearch/athens/issues/1718


filipesilva (approved):
Thanks for taking this on @hustodemon! I'm going to merge it now but it might be a bit until it shows up in the https://athensresearch.github.io/athens/ URL. We're not deploying it right now since it's pinned to the 1.0 release, and we're trying to avoid making a new stable release for 1.0 unless really needed.

Thanks again!

hustodemon:
Awesome, thanks @filipesilva !

PR #1879, opened 2021-11-27
feat: insert markdown link with cmd+k

tangjeff0:
Request by @cortexfutura

### DIFF
--- a/src/cljs/athens/views/blocks/textarea_keydown.cljs
+++ b/src/cljs/athens/views/blocks/textarea_keydown.cljs
@@ -492,6 +492,24 @@
          (set-cursor-position target (+ start n)))))))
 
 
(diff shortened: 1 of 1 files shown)


neotyk:
It's conflicting with Athena invocation.
Can we do some other shortcut for it?

lukaskawerau:
> It's conflicting with Athena invocation.
> Can we do some other shortcut for it?

Cmd-k is the universal shortcut for linking things, including here on Github. Imo Athens should conform to that convention. One option would be that the trigger does different things when a block is focused/you're typing inside a block and when no block is focused.

tangjeff0:
Happy medium solution:

- `cmd-p` for always-whenever Athena, in any context
- `cmd-k` for inserting links when cursor is in a block

PR #1880, opened 2021-11-29
save load cli

sid597:


src/clj/athens/self_hosted/event_log.clj
```diff
@@ -197,7 +217,7 @@
   ((:reconnect-fn comp))
 
   ;; Create ledger if not present.
-  (ensure-ledger! comp)
+  (ensure-ledger! comp )
```
filipesilva, line 220: nit: please remove extra space

src/clj/athens/self_hosted/event_log.clj
```diff
@@ -6,12 +6,28 @@
...
+(defn create-fluree-comp
+  [address]
+  (let [conn-atom    (atom nil)
+        reconnect-fn (fn []
+                       (when-some [conn @conn-atom]
+                         (fdb/close conn))
+                       (reset! conn-atom (fdb/connect address)))
+        comp         {:conn-atom    conn-atom
+                      :reconnect-fn reconnect-fn}]
+    (reconnect-fn)
```
filipesilva, line 27: It would be better if the creation of the fluree comp did not automatically connect, because it is surprising and it is reasonable to do these two different steps at two different times.

src/clj/athens/self_hosted/event_log.clj
filipesilva, line 220: nit: please remove extra space

src/clj/athens/self_hosted/event_log.clj
```diff
@@ -41,7 +57,9 @@
 (defn deserialize
   [{id   "event/id"
     data "event/data"}]
-  [id (edn/read-string data)])
+  [(if (str/blank? id)
```
filipesilva, line 60: This really should not be done. We shouldn't make up a new `event/id` out of thin air. If an event does not have an ID, it is not valid.

What's the case where this was necessary? Do we have events in our graphs that don't have an id?

script/save-load-cronjob.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Some of the strategies to calculate the filename
+#  - Timestamp
+#  - Monotonically increasing no.
+#  - No. of files in folder + 1
+
+# I am going with the 3rd option because it is the simplest one to follow.
+# Using `find` instead of `ls`: https://github.com/koalaman/shellcheck/wiki/SC2012
+FILENUMBER=$(find $FOLDER | wc -l)
+FILENAME="backup-no-$FILENUMBER"
```
filipesilva, line 48: Nit: a timestamp would be more useful and does not require counting files.

src/clj/athens/self_hosted/event_log.clj
```diff
@@ -6,12 +6,28 @@
...
+    [fluree.db.api :as fdb])
+  (:import
+    (java.util
+      UUID)))
 
 
 (def ledger "events/log")
 
 
+(defn create-fluree-comp
```
filipesilva, line 18: Can you move this function to `athens.self-hosted.components.fluree` instead, and also use it there inside the `defrecord` `start` method?

script/save-load-cronjob.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# command to save ledger
+
+yarn cli:save -f "$FILENAME"
```
filipesilva, line 52: This command will not work in our servers. They do not contain a clone of this repo containing `package.json` (where this script is defined), and even if the file was there you would also need all the dependencies.

Instead, this cron job should be running a built executable (e.g. an uberjar file probably). We build an uberjar for the server as well. It's not very obvious how to do this so we probably need to pair on it.

src/cljc/athens/athens_datoms.cljc
```diff
@@ -168,6 +168,7 @@
   (let [op (bfs/build-paste-op (datascript/empty-db common-db/schema)
                                welcome-page-internal-representation)
         welcome-page (common-events/build-atomic-event op)
+        ;; TODO NOW this is wrong need to wrap in build-atomic-event to create a uuid
```
filipesilva, line 171: This looks like it should be part of the PR.

filipesilva (changes requested):

script/save-load-cronjob.sh
neotyk: +1 for using date in filename `date +%F-%H-%M` and file extension would be nice (.edn)

src/clj/athens/self_hosted/event_log.clj
neotyk: Currently second event in every ledger has this situation, how we add Welcome page shortcut.

filipesilva: Ok I see now, it's related to that TODO. I agree we need to handle it. A comment here would be nice for the next time we see it.

src/clj/athens/self_hosted/event_log.clj
sid597: I am not sure about this, currently we need to automatically connect on creation of fluree comp in all the cases where this is being used.

 Removing it from here and doing it as a separate step for each case would result in same code in different places then we would have to maintain it all the places.

filipesilva: That's a good argument. I'm fine either way then.

I see you've already refactored the fn out. But now reconnect-fn is never called. You need to either:
- call reconnect-fn inside create-fluree-comp
- or call reconnect-fn where create-fluree-comp is called

Up to you which one you prefer.

sid597: yeah I messed up that part, I first removed it, came up with this arg but did not revert the removal :) Now it's done I went with the first one

filipesilva (approved):

PR #1881, opened 2021-11-29
fix: remove nils from inline presence too

filipesilva:
Same as toolbar presence is doing already.


src/cljs/athens/self_hosted/presence/views.cljs
```diff
@@ -100,6 +100,7 @@
                   :background "var(--background-color)"}}]
         (->> @users
              (map user->person)
+             (remove nil?)
```
filipesilva, line 103: This is the same as what we're doing above in line 67. The `user->person` fn can return nil, so we have to remove them, otherwise we get errors like the one we saw today.

sid597 (approved):

PR #1886, opened 2021-12-03
Structure parser

neotyk:
So we can handle Big Blocks


dev/notebooks/parser_notebook.clj
```diff
@@ -0,0 +1,57 @@
+;; # Structure parsing shenanigans
```
filipesilva, line 1: It was really nice to look at output and usage with the notebook.

test/athens/parser/structure_test.clj
```diff
@@ -0,0 +1,171 @@
+(ns athens.parser.structure-test
+  (:require
+    [athens.parser.structure :as sut]
+    [clojure.test :as t :refer [deftest is testing]]))
+
+
+(defmacro parses-to
```
filipesilva, line 7: Nit: would be nice for have this macro in another ns, like `athens.common.utils` (which is already cljc and has a macro), and then this ns could be cljc.

filipesilva (approved):

test/athens/parser/structure_test.clj
neotyk: moved it to `athens.common.utils`

PR #1891, opened 2021-12-06
build: use uberdeps to create parametrized executable uberjars

filipesilva:


filipesilva:
@sid597 this should make it easy to create another executable uberjar for the athens cli. Just create new package.json scripts:

```
    "cli:compile": "clojure -M -e \"(compile 'athens.cli.core)\"",
    "cli:uberjar": "clojure -M:uberdeps --main-class athens.cli.core --target target/athens-cli.jar",
```

Use whatever the cli main file path is, I don't remember it so I just used `athens.cli.core`.

deps.edn
```diff
@@ -1,4 +1,4 @@
-{:paths ["src/clj" "src/cljs" "src/cljc" "src/js" "src/gen" "test"]
+{:paths ["src/clj" "src/cljs" "src/cljc" "src/js" "src/gen" "test" "classes"]
```
neotyk, line 1: Why do we need compilation output to be treated as sources?
do we generate stuff that our code depends on?

filipesilva: Added some changes that shouldn't need classes to be in the global paths, lmk if you have a better way.

neotyk (approved):

PR #1892, opened 2021-12-07
feat: templates

filipesilva:
Use `/template` or type `;;` to look for a block to use as a template.

That blocks children will be inserted in place of the calling block, as if pasted.


mikejones3:
I see this is added, how do I create a template?

filipesilva:
Any block works as a template. When you use the slash command and pick template, you'll see a search box for blocks. Start typing some of the block content to search for a block. Pick a block to use that blocks children as the template.

PR #1894, opened 2021-12-09
Deploy clerk notebooks to vercel along with storybook

filipesilva:


.github/workflows/build.yml
```diff
@@ -294,3 +255,27 @@ jobs:
...
+         run: yarn storybook:static
+
+       - name: Build static Clerk site
+         run: yarn notebooks:static
+
+       - uses: amondnet/vercel-action@v20
+         with:
+           vercel-token: ${{ secrets.VERCEL_TOKEN }}
+           vercel-org-id: ${{ secrets.VERCEL_ORG_ID }}
+           vercel-project-id: ${{ secrets.VERCEL_PROJECT_ID }}
```
filipesilva, line 277: @tangjeff0 need you to set these three secrets. Am around to pair if you want.

The first one you can get from https://vercel.com/account/tokens.

The two other ones you can get from https://github.com/marketplace/actions/vercel-action#project-linking. You don't need to keep the `.vercel` folder around after getting them.

filipesilva:
Example comments and output in a PR on my fork: https://github.com/filipesilva/athens/pull/4

tangjeff0:
Briefly, why do we need to use Vercel Actions when Vercel GitHub integration was sufficient before? @filipesilva

tangjeff0 (approved):

tangjeff0:
`build / release-devtools` is failing because it uses `build.yaml` does not use the `with` secrets that Vercel needs.

PR #1899, opened 2021-12-13
fix(data-loss): Quitting athens while editing a block displays no warning and causes data loss. 

### LINKED ISSUE #1898: Potential Dataloss while editing a block
**Problem**
While editing a block, if the app is quit or refreshed, no warning for data not being saved is shown and all new edits made to the block are deleted

**Athens Version**
2.0.0 Beta 4

sawhney17:
Brings up a popup window warning user that edits may not have been saved if user is currently editing a block. 

Discussed in the data loss channel on discord. 

this will close #1898

### DIFF
--- a/src/cljs/athens/listeners.cljs
+++ b/src/cljs/athens/listeners.cljs
@@ -220,11 +220,11 @@
     EventType.BEFOREUNLOAD
     (fn [e]
       (let [synced? @(subscribe [:db/synced])
-            idEditing? @(subscribe [:editing/uid])
+            isEditing? @(subscribe [:editing/uid])
             remote? (electron-utils/remote-db? @(subscribe [:db-picker/selected-db]))]
         (cond
           (and (or (not synced?)
-                   (not (= nil idEditing?)))
+                   (not (= nil isEditing?)))
                (not @force-leave))
           (do
             ;; The browser blocks the confirm window during beforeunload, so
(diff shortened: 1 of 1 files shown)


src/cljs/athens/listeners.cljs
```diff
@@ -220,17 +220,19 @@
     EventType.BEFOREUNLOAD
     (fn [e]
       (let [synced? @(subscribe [:db/synced])
+            isEditing? @(subscribe [:editing/uid])
```
neotyk, line 223: `:editing/uid` signifies only that user has entered editing state, it doesn't mean there are unsaved changes yet.

sawhney17: This approach makes sense to me. If i'm in a text area, i'm likely in it with the intention to add data to it. I definitely wouldn't want to be kicked out by accident and would much prefer a popup while i'm writing. 

Would an alternate approach to this issue involve saving at intervals while typing in the block without necessarily having left it? 

I'm not quite sure how i would go about identifying if there are unsaved changes which is why this approach made sense to me. 

This is similar to how logseq does it with command r with a popup dialog confirming with the user, no matter if they have unsaved changes or not.

neotyk: This is a good MVP approach.
We'll need to expose editing dirty state, for next iteration of it.

Thank you for this contribution

neotyk, line 223: `isEditing?` is a mix between JS and CLJ naming.
In Clojure we'd just say `editing?`.
Change this and we're merging.

sawhney17: Got it, fixed now. Ready for merging i think

neotyk: @sawhney17 🙇‍♂️

PR #1904, opened 2021-12-15
test: add client e2e tests

filipesilva:


test/e2e/example.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { expect } from '@playwright/test';
+import { test } from './electron-test';
+
+test('basic test', async ({ page }) => {
```
filipesilva, line 4: We could probably do all of this in cljs. But all examples for playwright tests are in TS, and we are just getting started on them, so it's easier to start with TS.

test/e2e/electron-test.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+    await run(electronApp);
+    if (process.env.CI) {
+      // TODO: Should await on close here but it doesn't seem to work on the CI.
+      // If you edit the client:e2e:only:verbose script to debug on pw:* you can see
+      // more detailed logs on CI and it seems like the electron process responds
+      // that it closed, but playwright does not recognize. Maybe we need to
+      // update electron beyond v12.
+      electronApp.close();
+      await new Promise(resolve => setTimeout(resolve, 1000));
```
filipesilva, line 33: Don't have a better way to do this for CI, but this works. See the comment above for more details.

neotyk (approved):
👍

PR #1912, opened 2021-12-18
Fix docker-compose.yml

siraht:
The docker-compose.yml file has been incorrectly formatted for the version of compose used, for months now. I've fixed it by adding dashes in front of `depends_on` services as well as removing the deprecated `condition` fields within the `depends_on` fields.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
     #   dockerfile: athens.dockerfile
     restart: always
     depends_on:
-      - fluree:
+      - fluree
     ports:
       - 3010:3010 # Change this according to CONFIG_EDN.
     volumes:
@@ -35,7 +35,7 @@ services:
     #  dockerfile: nginx.dockerfile
     restart: always
     depends_on:
-      - athens:
+      - athens
     restart: always
     ports:
       - 80:80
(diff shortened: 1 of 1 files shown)


filipesilva:
Heya @siraht thanks for taking this on. I think something went wrong though, because the whole file shows up as new in the diff and it also shows up as a branch conflict in the merge widget.

I'm a bit confused about the file being incorrectly formatted though... I'd expect docker compose to complain about it, but I haven't seen any error messages or weird behaviour around `depends_on` or `condition`, neither locally nor on the servers we deploy to. I can see the containers waiting for each other to be in healthy states when starting too.

What version of docker and docker compose are you on? The server we run this file on is using Docker 20.10.8, and Docker Compose 1.27.4.

filipesilva:
I also see mentions of it working again as of docker compose 1.27: https://news.ycombinator.com/item?id=25329651

I'm very confused about this tbh... Kind expected to find some authoritative information somewhere in the official docker channels.

siraht:
Hey, sorry about the pull request issues, this was actually my first pull request ever! (And I think I know what I did wrong :P)

Anyways, with an updated Docker installation on Ubuntu (tried on multiple installations) I could not get the docker-compose working except for the configuration here.

PR #1923, opened 2021-12-21
fix: optimistic operations need to be atomic

filipesilva:


src/cljs/athens/events/remote.cljs
```diff
@@ -84,79 +85,137 @@
...
+  "Apply all events in the :memory stage of event-sync via resolve.
+  Should only be used preceded by rollback."
+  [[db conn]]
+  (let [in-memory-events       (-> db :remote/event-sync :stages :memory)
+        count-in-memory-events (count in-memory-events)
+        reapplied-tx-data      (atom {})]
+
+    (if (= count-in-memory-events 0)
+      (do
+        (log/debug "rollforward skipped, nothing to rollforward")
```
neotyk, line 144: It looks like it's unexpected, we've rolledback to rollfarward (rebase), but there is nothing to rebase, correct?

If so bump this to `log/warn`.

neotyk (approved):

src/cljs/athens/events/remote.cljs
filipesilva: This can happen on `:remote/reject-forwarded-event`, if the rejected event was the only one or not found, and during `:remote/apply-forwarded-event` if there was nothing rolled back at all. I think those are still legitimate cases.

PR #1934, opened 2022-01-04
cronjob script for save (backup) athens server

sid597:


script/save-cronjob.sh
```diff
@@ -46,5 +49,11 @@ FOLDER=/var/lib/athens/backups/
 TIMESTAMP=$(date +%F-%H-%M)
 FILENAME="${FOLDER}${TIMESTAMP}.edn"
 
+# If java not installed then install it
+
+if [ -z "$(which java)" ]; then
+  apt -y install openjdk-11-jre-headless
```
neotyk, line 55: This assumes the linux distro, let's just warn here and `exit 1` so it's clear failure.

sid597: yeah you are right I was thinking of this being used on servers which are linux but people can also do this while testing locally on other platforms

neotyk (approved):

tangjeff0:
Please make the name of this PR more descriptive @sid597

tangjeff0 (approved):

PR #1935, opened 2022-01-04
idle save

filipesilva:
- fix: idling for 2s on a block with changes will save it
- fix: don't flicker old state on save


neotyk (approved):
LGTM

PR #1943, opened 2022-01-06
moving with arrow keys MVP

sid597:


tangjeff0 (changes requested):
Demo: https://www.loom.com/share/b4dece1603bc4870b588c062a604713d

When going UP or LEFT and the block ABOVE is indented further to the right than the original block and is closed, the caret loses focus and disappears.

I noticed this bug while testing this PR, but it seems to have also existed before this PR on main, so it should be addressed in a separate PR.

---

Please add comments for other cases. Otherwise the code looks fine, and the application is behaving as we specified.

src/cljs/athens/views/blocks/textarea_keydown.cljs
```diff
@@ -415,14 +416,34 @@
 
       ;; Else: navigate across blocks
       ;; FIX: always navigates up or down for header because get-caret-position for some reason returns the wrong value for top
+
+      ;; going LEFT at **0th index** should always go to **last index** of block **above**
```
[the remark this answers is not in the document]
sid597: addressed this case with comment below

src/cljs/athens/views/blocks/textarea_keydown.cljs
```diff
@@ -415,14 +416,34 @@
...
+      ;; last index is special - always go to last index when going up or down
```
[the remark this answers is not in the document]
sid597: same addressed this case with comment below (439)

src/cljs/athens/views/blocks/textarea_keydown.cljs
```diff
@@ -371,7 +371,8 @@
         down?                            (= key-code KeyCodes.DOWN)
         left?                            (= key-code KeyCodes.LEFT)
         right?                           (= key-code KeyCodes.RIGHT)
-        header                           (db/v-by-ea (db/e-by-av :block/uid uid) :block/header)]
+        header                           (db/v-by-ea (db/e-by-av :block/uid uid) :block/header)
```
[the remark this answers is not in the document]
sid597: I also don't know about this, did not check this while working on this issue. Will look into it

sid597:
Yeah in addition to "When going UP or LEFT and the block ABOVE is indented further to the right than the original block and is closed, the caret loses focus and disappears." there is another bug where we lose the caret while moving UP from the 1st block. We should address both these bugs in a separate issue.

src/cljs/athens/views/blocks/textarea_keydown.cljs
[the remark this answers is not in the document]
neotyk: I don't think it's used anymore
```clojure
(->> (d/datoms @dsdb :eavt) 
                (filter #(= :block/header (second %)))
                count)
```

src/cljs/athens/views/blocks/textarea_keydown.cljs
```diff
@@ -415,14 +416,34 @@
...
+      ;; last index is special - always go to last index when going up or down
+
+      ;; NOTE: Using 99999999999 is a hack, if the previous block has less character than mentioned then the default
+      ;;       caret position will be last position. Otherwise, we would have to calculate the no. of characters in the
+      ;;       block we are moving to, this calculation would be done on client side and, I am not sure if the calculation
+      ;;       would be correct because between calculation on client side and block data on server can change.
+
+      (or (and left? start?)
+          (and up? end?))         (do (.. e preventDefault)
+                                      (dispatch [:up uid 99999999999]))
```
neotyk, line 430: It would be better to use sentinel `:end` instead of magic number here, and in handler for both `:up` & `:down` calculate what `:end` means.

src/cljs/athens/views/blocks/textarea_keydown.cljs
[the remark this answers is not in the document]
tangjeff0: I remember this now. We don't use this anymore. Please delete it @sid597

src/cljs/athens/views/blocks/textarea_keydown.cljs
tangjeff0: Did you read his comments @neotyk ?

sid597: Alex to calculate the end we would have to look at the end of block to which caret is moving. This calculation being done on client side could be wrong

sid597: For example 

Current block text: hello world
Next block text   : Hello world this

To move from current block to next block, we would have to calculate :end position which would be position after "this". 

But it could be possible that between the client calculating the position and sending it to server some other client added more thus making 

Next block text: Hello world this is sid

neotyk: > Did you read his comments @neotyk ?

Yes

neotyk: I do get it that it can change.
And that we can feed `setCursorPostion` too big value, it's ok.
What I'd prefer to see here is meaningful semantics, this magic number is accidental complexity of implementation,
we really mean here stick to `:end` of block.
Yes we'll resolve it to a big number, please let's do it close to implementation that requires this accidental complexity.

tangjeff0 (approved):

PR #1954, opened 2022-01-12
enhance: Keyboard shortcut for navigating forward and backward

sid597:
shortcut for:
mac - meta + left/right
linux - alt + left/right
windows - alt + left/right

### DIFF
--- a/src/cljs/athens/electron/boot.cljs
+++ b/src/cljs/athens/electron/boot.cljs
@@ -61,12 +61,13 @@
                                       :dispatch-n [[:navigate :home]
                                                    [:left-sidebar/toggle]]}
 
-                                     ;; if nth time, remember dark/light theme and last page
+                                     ;; if nth time, remember dark/light theme
                                      {:when       :seen?
                                       :events     :reset-conn
                                       :dispatch-n [[:fs/update-write-db]
                                                    [:db/sync]
-                                                   [:restore-navigation]
+                                                   ;; [:restore-navigation]
+                                                   [:posthog/set-super-properties]
                                                    [:loading/unset]]
                                       ;; This event ends the async flow successfully.
                                       :halt?      true}
--- a/src/cljs/athens/electron/core.cljs
+++ b/src/cljs/athens/electron/core.cljs
@@ -3,6 +3,7 @@
     [athens.electron.boot]
     [athens.electron.db-picker]
     [athens.electron.fs]
+    [athens.electron.monitoring.core]
     [athens.electron.window]
     [athens.util :as util]
     [day8.re-frame.async-flow-fx]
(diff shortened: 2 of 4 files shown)


tangjeff0 (changes requested):

src/cljs/athens/listeners.cljs
```diff
@@ -82,45 +83,55 @@
...
-  (let [{:keys [key-code ctrl meta shift alt]} (util/destruct-key-down e)
-        editing-uid @(subscribe [:editing/uid])]
+  (let [{:keys [key-code
+                ctrl
+                meta
+                shift
+                alt]}    (util/destruct-key-down e)
+        editing-uid      @(subscribe [:editing/uid])
+        start?          (if editing-uid
+                          (textarea-keydown/block-start? e)
```
tangjeff0, line 93: Why are you we checking for `block-start?` and `block-end`?

src/cljs/athens/listeners.cljs
```diff
@@ -82,45 +83,55 @@
...
+                                      KeyCodes.D     (router/nav-daily-notes)
+                                      KeyCodes.G     (router/navigate :graph)
+                                      KeyCodes.A     (router/navigate :pages)
+                                      KeyCodes.T     (dispatch [:theme/toggle])
+                                      nil)
+      (util/navigate-key? meta alt) (condp = key-code
+                                      KeyCodes.LEFT  (when (or (nil? editing-uid)
+                                                               start?)
+                                                       (.back js/window.history))
+                                      KeyCodes.RIGHT (when (or (nil? editing-uid)
```
tangjeff0, line 131: Why `(nil? editing-uid)` here and also `if` statements for `editing-uid` on lines 92 and 95? Seems redundant.

neotyk (approved):
too good

src/cljs/athens/views/blocks/textarea_keydown.cljs
```diff
@@ -374,6 +374,9 @@
         [char-offset _]                  (get-end-points target)]
 
     (cond
+      (util/navigate-key? keys-destruct) (cond
```
sid597, line 377: Need to handle the case for textarea navigation here because in mac navigation while in textarea does not work if we try to do this in listners. The issue is somehow `meta` keypress is not bubbled up and for now we (alex and me) could not figure out why is it so.

sid597 (commented):
self review

tangjeff0 (approved):
Linux: two cases where it doesn't work:
- when caret is focused on a title (block or page)
- after navigating to a page without children, the caret loses focus and it doesn't work

Mac: must be focused in on a block for navigation to work at all. Linux works without focus (sometimes). This is because we do textarea keybindings, and not global keybindings.

Still approving because this seems like's incrementally better. Getting this right without using a keybinding library is more effort than perhaps worth.

PR #1971, opened 2022-01-18
Undo block move

sid597:


src/cljc/athens/common_events/resolver/undo.cljc
```diff
@@ -40,6 +40,38 @@
...
+                                    true
+                                    false)
+        parent-ref                (if parent-page?
+                                    (:node/title parent)
+                                    (:block/uid parent))
+        {prev-sib-uid :block/uid} (common-db/nth-sibling evt-db uid -1)
+        prev-block-pos            (cond
+                                    (= order 0) :parent
+                                    :else       :prev-sib)
+        position                  (cond
```
filipesilva, line 65: Can you use the new `common-db/get-position` here instead? It was merged into the undo branch yesterday.

filipesilva (changes requested):
LGTM, nice tests!

Would just like to use the new helper instead.

src/cljc/athens/common_events/resolver/undo.cljc
```diff
@@ -58,17 +58,10 @@
         parent-ref                (if parent-page?
                                     (:node/title parent)
                                     (:block/uid parent))
-        {prev-sib-uid :block/uid} (common-db/nth-sibling evt-db uid -1)
-        prev-block-pos            (cond
-                                    (= order 0) :parent
-                                    :else       :prev-sib)
         position                  (cond
```
filipesilva: You can replace this one too, all you should need is:
```
position                    (common-db/get-position evt-db uid)
````

filipesilva (approved):

PR #1979, opened 2022-01-19
enhance(db-picker): allow users to remove dbs from list

tangjeff0:
v1 - https://www.loom.com/share/e94c905163e54f4d937c823d4aabbd20
v2 in response to Alex and Stuart's feedback - https://www.loom.com/share/f52ea62f5053457191b63a40bc7f7971

### DIFF
--- a/src/cljs/athens/electron/db_menu/db_list_item.cljs
+++ b/src/cljs/athens/electron/db_menu/db_list_item.cljs
@@ -3,9 +3,9 @@
     ["@material-ui/icons/Clear" :default Clear]
     ["@material-ui/icons/Link" :default Link]
     [athens.electron.db-menu.db-icon :refer [db-icon]]
+    [athens.electron.dialogs :as dialogs]
     [athens.style :refer [color]]
     [re-frame.core :refer [dispatch]]
-    [reagent.core :as r]
     [stylefy.core :as stylefy :refer [use-style]]))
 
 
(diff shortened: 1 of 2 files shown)


src/cljs/athens/electron/db_menu/db_list_item.cljs
```diff
@@ -59,10 +61,21 @@
...
+        remove-db-click-handler (fn [e]
+                                  (dispatch [:db-picker/remove-db db])
+                                  (.. e stopPropagation))]
+    (fn []
+      [:div (use-style db-list-item-style)
+       (if is-current
+         [:div.body.is-current
+          [db-list-item-content {:db db}]]
+         [:button.body.button {:onClick      #(dispatch [:db-picker/select-db db])
+                               :onMouseOver  mouse-over-handler
```
neotyk, line 76: Why do we want to show UI element when cursor enters the area?
Sounds like "surprise"

src/cljs/athens/electron/db_menu/db_list_item.cljs
```diff
@@ -59,10 +61,21 @@
...
 (defn db-list-item
   [{:keys [db is-current]}]
-  [:div (use-style db-list-item-style)
-   (if is-current
-     [:div.body.is-current
-      [db-list-item-content {:db db}]]
-     [:button.body.button {:onClick
-                           #(dispatch [:db-picker/select-db db])}
-      [db-list-item-content {:db db}]])])
+  (let [state                   (r/atom {:show-remove-button false})
```
neotyk, line 64: If you really have to have local state - and please don't have it, make it part of reframe -
you don't need to have a map with just 1 key, simpler would be to have `show? (r/atom false)`.

neotyk (changes requested):

neotyk (approved):

PR #1991, opened 2022-01-22
fix: small paste bugs

tangjeff0:
Demos

1a. https://www.loom.com/share/811939853f6646058b481860f0e1d429

1b. https://www.loom.com/share/cd507d3e31a240fa95d5ae65551a03d8

---

1. fixes subtle issues with paste
  a. pasting while there is an unsaved local string overwrote it; now it doesn't
  b. pasting when on a block that is open with children previously added blocks to the top of the page; now it adds them to the top of the children
2. wasn't able to get tests to runs in Cursive in my REPL, so temporarily using the `fixture/integration-test-fixture`. will remove once the general logic for this PR or I can get the tests to compile in my REPL
3. there is a more general problem regarding 1a. oftentimes user will trigger some sort of graph op while they have an unsaved local-string edit, e.g. paste, indent, enter. Seems like there should be a more elegant, generic solution that does better local-string handling.
4. currently no `:editing/uid` is dispatched after paste events, user just says on the same block. `:editing/uid` should be dispatched, but want to wait for a larger refactor here (I think depending on Alex's async-flow-fx)

### DIFF
--- a/src/cljc/athens/common_events/bfs.cljc
+++ b/src/cljc/athens/common_events/bfs.cljc
@@ -105,32 +105,44 @@
 
   ([db internal-representation]
    (composite/make-consequence-op {:op/type :block/paste} (internal-representation->atomic-ops db internal-representation nil)))
-  ([db uid internal-representation]
-   (let [current-block-parent-uid             (:block/uid (common-db/get-parent db [:block/uid uid]))
+  ([db uid local-str internal-representation]
+   (let [current-block-parent-uid (:block/uid (common-db/get-parent db [:block/uid uid]))
          {:block/keys [order
                        children
                        open
-                       string]}               (common-db/get-block db [:block/uid uid])
+                       string]} (common-db/get-block db [:block/uid uid])
          ;; The parent of block depends on:
          ;; - if the current block is open and has chidren : if this is the case then we want the blocks to be pasted
          ;;   under the current block as its first children
          ;; - else the parent is the current block's parent
-         current-block-parent?                (and children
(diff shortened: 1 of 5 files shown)


src/cljc/athens/common_events/bfs.cljc
```diff
@@ -105,30 +105,44 @@
...
-         extra-ops                            (if empty-block? [(graph-ops/build-block-remove-op db uid)] [])]
+         new-block-order          (cond
+                                    empty-block? order
+                                    current-block-parent? 0
+                                    :else (inc order))
+         block-position           (cond
+                                    empty-block? current-block-parent-uid
+                                    current-block-parent? uid
+                                    :else current-block-parent-uid)
+         ;; do we need compat position? the 3 cases are X, :first, and :after. I'm not sure about the case of empty-block?
```
filipesilva, line 138: You probably still want compat-position to convert the uid for a page into title, unless you want to do that resolution yourself.

filipesilva (approved):
It's not very clear to me what the generic solution to local-string is. The crux of it is timing: sometimes you want to do stuff because you left the block, and sometimes you know the action is going to leave the block so you need to roll the exit logic in.

In a sense the abstraction here is to "commit" the block changes. But I don't think this would work very well with re-frame's resolver right now, because it'd be easy to mess up order (e.g. commit then paste-op, or paste-op then commit) based on the dispatch sequence.

Maybe the answer is again the async flow stuff. Edit block -> exit block -> block commit -> actions.

PR #1992, opened 2022-01-22
fix: pressing up and the previous block has a closed child

tangjeff0:
https://www.loom.com/share/e231434d9eb44bd8aeb59bc58f1443c6

I think there's another case where something similar happens: backspace and merge when previous block is more indented.

Is this something that needs tests?

### DIFF
--- a/src/cljs/athens/db.cljs
+++ b/src/cljs/athens/db.cljs
@@ -386,12 +386,13 @@
 
 (defn deepest-child-block
   [id]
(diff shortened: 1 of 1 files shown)


filipesilva (approved):
Only sort of test that would catch this is e2e test. Would you like to pair on making one?

tangjeff0:
Maybe if there's a regression. Seems like this sort of should be encoded somewhere eventually though. @filipesilva

PR #1994, opened 2022-01-23
feat: basic API

filipesilva:


src/clj/athens/self_hosted/web/api.clj
```diff
@@ -0,0 +1,81 @@
...
+   :body x})
+
+
+;; Routes with inline handlers.
+
+(defn make-routes
+  [datascript fluree config]
+  (let [conn        (:conn datascript)
+        ;; TODO: handle pw
+        ;; password    (-> config :config :password)
```
filipesilva, line 58: This will really need to be done, even for a MVP.

PR #2006, opened 2022-01-27
feat: show references inline

filipesilva:


src/cljs/athens/views/blocks/core.cljs
```diff
@@ -361,11 +483,17 @@
            [presence/inline-presence-el uid]
 
            (when (and (> (count _refs) 0) (not= :block-embed? opts))
-             [block-refs-count-el (count _refs) uid])]
+             [block-refs-count-el (count _refs) #(swap! state update :inline-refs/open not)])]
```
tangjeff0, line 486: If normal click, open inline refs. **If shift-click, open in right sidebar.**

tangjeff0 (changes requested):
Looks good other than shift-click behavior!

src/cljs/athens/views/blocks/core.cljs
filipesilva: fixed

tangjeff0 (approved):

PR #2016, opened 2022-02-06
feat: speedy window re-open on macOS

hyferg:
Normal mac behavior on clicking the red x is to hide the window so it can be opened instantly from the dock and not have to got through the whole initial boot cycle.

You might want to now move the (prevent-save) function onto this close listener so that on macos it can keep syncing in the background since the window is now hidden but not destroyed.

I'd move (prevent-save) to the window "close" event myself but I have no experience using clojure.

### DIFF
--- a/src/cljs/athens/main/core.cljs
+++ b/src/cljs/athens/main/core.cljs
@@ -77,6 +77,9 @@
   )
 
 
+(def quitting (atom false))
+
+
 (defn init-browser
   []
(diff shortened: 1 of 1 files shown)


filipesilva (approved):

filipesilva:
@hyferg added some more code so that cmq+q still quits and clicking on the dock icon still reopens/shows. The `prevent-save` fn seems to be fine since it only really runs when the browser window is destroyed, not hidden. Thanks for the contrib, should be in the next release!

PR #2023, opened 2022-02-13
Add athens-web docker container for self hosting

ww-daniel-mora:
This change introduces an athens-web docker container for self hosting the web client. As is, the change is incomplete and inconsistent with the other docker containers. This change has the following work items unresolved:
* How will the container be built? I have added a docker build to the package.json as a matter of convenience, however, the other containers are not built the same way.
* What would the image be called? Here it is named athens-web.
* Should the docker-compose include the web container
* What documentation would need to be updated

I request feedback from the maintainers on how to resolve these questions


neotyk:
@ww-daniel-mora Thank you for this Pull Request.
I'm not sure about the value standalone web build provides.

It naturally belongs to self-hosted setup, and as such should be served by our Nginx container as static asset.
It this situation it would become web interface of self-hosted deployment, once we solve SSL situation.

ww-daniel-mora:
@neotyk thank you for the feedback. Let me take a swing at getting this served by your existing nginx container. What ssl issue are you referring

ww-daniel-mora:
Whoops hit the close button but accident!

ww-daniel-mora:
@neotyk Okay I've updated the change to use the existing nginx container. This change resolves my questions except for documentation which I would be happy to contribute to. Have a look and let me know what you think

ww-daniel-mora:
@neotyk bump on this pr 🙏

ww-daniel-mora:
@filipesilva thank you for the update. I'm happy that this functionality is being added. I'll keep an eye on the updates so I can try it out when it is published!

PR #2025, opened 2022-02-17
fix: navigate to page from athena when result is for a block

sid597:


tangjeff0 (changes requested):
- Code is correct, but I think we should actually `navigate-uid` for blocks.
- Make sure this works for both enter and click handlers.
- Can you try writing e2e tests for this?
- Let me think about what `:editing/uid` should be going forward. I think it is being called multiple times right now.

sid597:
> Code is correct, but I think we should actually navigate-uid for blocks
I was thinking about this too but it was not clear to me what happens when we navigate to a child block whose parent is closed.

>  Make sure this works for both enter and click handlers.
I tested click handlers are working for athena navigation

> Can you try writing e2e tests for this?
Yup

tangjeff0:
> I was thinking about this too but it was not clear to me what happens when we navigate to a child block whose parent is closed.

If you go directly to the child block as a `block-page`, it doesn't matter if parent is closed.

sid597:
> If you go directly to the child block as a block-page, it doesn't matter if parent is closed.

Oh ok so the intention is to open as `block-page` will do this, I had the wrong impression that we would like to navigate to block on its parent page.

sid597:
> Deleting a page leads to an error:

This was introduced by monitoring I think we should add this there and track there

tangjeff0:
> Oh ok so the intention is to open as block-page will do this, I had the wrong impression that we would like to navigate to block on its parent page.

That would be cool, but I think that requires more UX thought, similar to `open-close` in the case of collaboration. For now, I'd rather zoom in directly on the block than to go to the page and not know where the block is, in the case that it's deeply nested.

tangjeff0 (changes requested):

tangjeff0 (approved):

PR #2029, opened 2022-02-17
fix: boot sequence and db-dump async flows

neotyk:
Added `:stage/success-db-load` and `:stage/fail-db-load`,
those are used in boot flow.

### DIFF
--- a/src/cljs/athens/events.cljs
+++ b/src/cljs/athens/events.cljs
@@ -45,7 +45,27 @@
     (let [conn (d/create-conn common-db/schema)]
       (doseq [[_id data] athens-datoms/welcome-events]
         (atomic-resolver/resolve-transact! conn data))
-      {:dispatch [:reset-conn @conn]})))
+      {:async-flow {:id             :db-in-mem-load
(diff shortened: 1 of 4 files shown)


src/cljs/athens/events.cljs
```diff
@@ -45,7 +45,27 @@
     (let [conn (d/create-conn common-db/schema)]
       (doseq [[_id data] athens-datoms/welcome-events]
         (atomic-resolver/resolve-transact! conn data))
-      {:dispatch [:reset-conn @conn]})))
+      {:async-flow {:id             :db-in-mem-load
```
filipesilva, line 48: Why async flow here, where it just really does the sync reset conn, but no async flow in `:fs/read-and-watch`?

I think what you did in `:fs/read-and-watch` (just add `:stage/success-db-load` to the dispatch-n) should work here just as well.

filipesilva (approved):

src/cljs/athens/events.cljs
neotyk: I've changed `:fs/read-and-watch` to use correct approach too

PR #2030, opened 2022-02-18
undo textarea

filipesilva:
- rfct: add a few e2e utils
- feat: use the same undo for local and remote
- test: add undo e2e
- fix: always use athens undo/redo
- rfct: remove old single player undo


test/athens/common_events/left_sidebar_test.cljc
```diff
@@ -1,212 +0,0 @@
-(ns athens.common-events.left-sidebar-test
```
tangjeff0, line 1: What happened here?

test/e2e/undo-input.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    // All three blocks should be there.
+    await expect(page.locator('.block:has-text("one")')).toBeVisible();
+    await expect(page.locator('.block:has-text("two")')).toBeVisible();
+    await expect(page.locator('.block:has-text("three")')).toBeVisible();
+
+    // Undo once, removing the last block input and split.
+    await page.keyboard.press(undoShortcut);
+
+    // Last block should be gone.
+    await expect(page.locator('.block:has-text("one")')).toBeVisible();
```
tangjeff0, line 29: How is `expect` async here? I thought this would have to be written as

`expect(await page.locator('.block:has-text("one")')).toBeVisible();`

test/e2e/utils.ts
```diff
@@ -26,24 +28,72 @@ export const createLocalAthensDB = async (page:Page, dbName:string) => {
...
-    await page.click('.textarea >> nth=-1');
-    return page.press('.textarea  >> nth=-1', 'Shift+Tab');
+    await page.click(lastBlockSelector);
+    return page.press(lastBlockSelector, 'Shift+Tab');
 };
+
+
+export const goToDailyPages = async (page:Page) => {
+    // The sixth button is the daily notes button.
+    // TODO: find a better way to address this button, maybe tooltip?
```
tangjeff0, line 67: I think we could CSS classes to the Typescript components and select them that way.

test/e2e/utils.ts
```diff
@@ -26,24 +28,72 @@ export const createLocalAthensDB = async (page:Page, dbName:string) => {
...
+    await page.click('button:nth-child(6)');
+}
+
+export const waitForBoot = async (page:Page) => {
+    // Wait for an element that signals the app has finished loading.
+    // Normally on e2e tests we'd load a page, but on a electron app we should rely
+    // only on visible behaviour.
+    // TODO: This isn't necessary on production builds, but is necessary for dev
```
tangjeff0, line 75: Makes sense to me

tangjeff0 (approved):

tangjeff0:
What's needed to get `:editing/uid` to get set as well after undo/redo?

tangjeff0:
Also, if user undos after creating a page while the user is on that page, they are left with an empty page.

image

There's a more general problem here of how to undo/redo not just datascript state but also DOM state:

- cursor focus
- page
- scroll
- (what else)?

Seems hard :) https://twitter.com/steveruizok/status/1495317652046290944

test/athens/common_events/left_sidebar_test.cljc
filipesilva: I think we refactored these events into atomic events, disabled the test here, and never deleted it. I only found it because I removed the `.resolve` ns due to removing undo.

test/e2e/utils.ts
filipesilva: Usually for e2e we're looking at things a user could see in some way. A CSS class doesn't fit the bill, but a particular icon, text, or tooltip text does.

filipesilva:
We probably need some guardrails around large undo effects, stuff like deleting a whole page or moving like 40 blocks. Think that's out of scope of this PR though.

Unsure what concretely we need to do to set the cursor on the right place, but I am sure it's doable with what we have right now. We need to have a stronger concept of "block to focus", and save it on the undo/redo stack.

test/e2e/undo-input.spec.ts
filipesilva: I think the expectation itself is async, because I know it can wait a certain amount of time for it to be or not be visible. If the locator itself was the async part here, I don't think it could wait to not be visible.

test/e2e/utils.ts
tangjeff0: Great explanation!

PR #2040, opened 2022-02-22
web build e2e

filipesilva:
- fix: athena navigation test should wait for boot
- test: use primarily web build for e2e


neotyk (approved):
LFG

neotyk (approved):
LFG

PR #2041, opened 2022-02-23
Fix: multi block selection copy from context menu

sid597:


sid597:
could not find a way to add e2e tests because clipboard support is not good on playwright, there is also an issue related to clipboard of test passing in headful state and consistently failing on headless for chromium. 

References:
https://github.com/microsoft/playwright/issues/8114
https://github.com/microsoft/playwright/issues/11654

PR #2042, opened 2022-02-23
Selecting multiple blocks when the root parent is out of the viewport forces user to scroll upwards in a very jarring way

sid597:


test/e2e/utils.ts
```diff
@@ -113,10 +113,3 @@ export const deleteCurrentPage = async (page:Page) => {
     await page.click('button:has-text("Delete Page")');
     await page.click('button:nth-child(6)');
 }
-
```
filipesilva, line 116: @sid597 please don't remove this function, it's being used.

PR #2046, opened 2022-02-24
fix: handle loading states on reconnect in a sane way

filipesilva:


src/cljs/athens/events.cljs
```diff
@@ -513,10 +513,15 @@
...
 
-(reg-event-db
+(reg-event-fx
   :conn-status
-  (fn [db [_ status]]
-    (assoc db :connection-status status)))
+  (fn [{:keys [db]} [_ to-status]]
+    (let [from-status (:connection-status db)]
+      {:db (assoc db :connection-status to-status)
+       :dispatch-n [(condp = [from-status to-status]
```
neotyk, line 521: Is `:dispatch-n [nil]` happy or complaining?

src/cljs/athens/self_hosted/client.cljs
```diff
@@ -64,13 +64,12 @@
        (rf/dispatch [:remote/connection-failed])))))
 
 
-(defn- close-reconnect-timer!
+(defn close-reconnect-timer!
```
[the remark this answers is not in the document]
filipesilva: Mistake of mine while changing things, good catch!

src/cljs/athens/events.cljs
filipesilva: `:dispatch-n [nil]` is happy but `:dispatch nil` is not.

PR #2051, opened 2022-03-01
feat: don't warn on exit while editing

filipesilva:
This change was first introduced in
https://github.com/athensresearch/athens/pull/1899.

It made a lot of sense at the time because it was so easy to lose
content that was being edited.
But now we auto-save after 2s of inactivity in a block, so I don't think
it's as big of a problem anymore.

I do get the warning popup very often so I think now it's doing more
harm than good.

### DIFF
--- a/src/cljs/athens/listeners.cljs
+++ b/src/cljs/athens/listeners.cljs
@@ -223,13 +223,11 @@
     EventType.BEFOREUNLOAD
     (fn [e]
       (let [synced? @(subscribe [:db/synced])
-            editing? @(subscribe [:editing/uid])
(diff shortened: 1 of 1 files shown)


filipesilva:
cc @sawhney17

sawhney17:
I think it makes sense to go ahead with this change now that auto save was happenign. the main issue initially was with regards to long blocks but in two seconds of time, I think it's a good buffer, in the _worst_ case it's 2-3 words of loss, but that's acceptable especially considering the likliness of accidentally hitting command r or command q mid sentence.

PR #2061, opened 2022-03-03
docs: add Reactive DataScript ADR

filipesilva:


tangjeff0 (approved):

tangjeff0:
Do you have any comments on possible future next steps for greater optimization?

Any thoughts on doing on replacing posh with another lib our doing our own reactive layer over datascript?

filipesilva:
I think there's further possible optimizations that would let us comfortably render pages with more than 5k-10k, but better than rendering everything quickly once is not having to render everything at once. This would be the infinite scroll-like approaches that @neotyk suggested. We'd only render roughly what's visible on screen, and thus only create Posh subscriptions for on-screen blocks.

I don't have any real reason to try and replace Posh at this point. It seems fit for the general usage we're giving it. The work we did when finding out soft limits does not reveal any obvious problem areas.

PR #2062, opened 2022-03-03
fix: set a 1MB event limit

filipesilva:


filipesilva:
@sid597 would like your help testing the limit, since you already have test code.

@neotyk do you think this is a reasonable way of adding the limit check, or do you have a better idea?

filipesilva:
Realized this later on: this limit also applies to db dump will cause our team server to stop working. We can just exclude db dump from the limit, but the question of what limit there is for db dump still remains.

src/cljc/athens/common_events.cljc
```diff
@@ -1,7 +1,67 @@
...
+
+(def serialization-type :json)
+
+
+(defn serialize
+  [event]
+  (let [writer #?(:cljs (transit/writer serialization-type)
+                  :clj  (transit/writer (ByteArrayOutputStream. 4096)
+                                        serialization-type))]
+    (transit/write writer event)))
```
filipesilva: Where did the `(.toString out)` from `->transit` go? Is it not needed anymore?

src/cljc/athens/common_events.cljc
```diff
@@ -1,7 +1,67 @@
...
+
+(defn valid-serialized-event?
+  [serialized-event]
+  (< (count serialized-event) max-event-size-in-bytes))
+
+
+(defn validate-serialized-event
+  [serialized-event]
+  (when-not (valid-serialized-event? serialized-event)
+    (ex-info "Serialized event is larger than 10 MB" {})))
```
filipesilva, line 32: This should say `1 MB` instead of `10 MB`.

src/cljs/athens/self_hosted/client.cljs
```diff
@@ -102,8 +97,14 @@
          (log/debug "event-id:" (pr-str (:event/id data))
                     ", type:" (pr-str (:event/type data))
                     "WSClient sending to server")
-         (.send connection (transit/write (transit/writer :json) data))
-         {:result :sent})
+         (let [serialized-event (common-events/serialize data)
+               errors           (common-events/validate-serialized-event serialized-event)]
+           (if errors
+             (do (log/warn "Received invalid serialized event:" (pr-str errors))
```
filipesilva, line 103: This message should be `Tried sending invalid serialized event`

filipesilva:
I will need to test we don't have any 1mb event before merging by replaying the team log.

src/cljc/athens/common_events.cljc
filipesilva: This is already fixed in a subsequent commit.

neotyk:
When testing, I can't even paste 80KB of text into the app.

neotyk:
LGTM

neotyk (approved):

PR #2068, opened 2022-03-08
fix: Allow to start with empty line

neotyk:
Used to be severely broken if block would have a block level
element with 1st line empty like following blockquote
```
>
> Text
```


filipesilva (approved):
LGTM but should have a test for the newly enabled behaviour

PR #2072, opened 2022-03-09
datascript persist

filipesilva:
- feat: add athens.self-hosted.web.persistence ns
- feat: add :datascript :persist-base-path to server config
- rfct: make most persistence ns fns private
- rfct: use /datascript/persist instead of just /persist/
- feat: persist server datascript db

### DIFF
--- a/src/clj/athens/self_hosted/event_log.clj
+++ b/src/clj/athens/self_hosted/event_log.clj
@@ -113,6 +113,7 @@
           ;; This is a terrible way to do what we want.
           ;; Instead we should filter out all the events we don't care about on the fluree query.
           ;; But when I (filipe) tried to do that, it made each events-page query take 30s instead of 0.3s.
+          ;; See https://github.com/fluree/db/issues/160
           ;; This seems good enough for now.
           ;; TODO: figure out a performant way to do this.
           (drop-while (fn [[id]]
--- a/src/clj/athens/self_hosted/event_log.clj
+++ b/src/clj/athens/self_hosted/event_log.clj
@@ -73,7 +73,7 @@
 (defn- event-id->subject-id
   [db event-id]
   (first @(fdb/query db {:select "?event"
-                         :where  [["?event" "event/id", event-id]]})))
+                         :where  [["?event" "event/id", (str event-id)]]})))
 
 
 (defn last-event-id
--- a/src/clj/athens/self_hosted/event_log.clj
+++ b/src/clj/athens/self_hosted/event_log.clj
@@ -76,6 +76,21 @@
                          :where  [["?event" "event/id", event-id]]})))
 
 
+(defn last-event-id
+  [fluree]
+  (-> fluree
+      :conn-atom
+      deref
+      (fdb/db ledger)
+      (fdb/query {:selectOne {"?event" ["*"]}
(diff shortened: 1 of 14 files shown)


src/clj/athens/self_hosted/web/persistence.clj
```diff
@@ -0,0 +1,102 @@
...
+         d/serializable
+         json/write-str
+         (spit path))
+    (delete-others! persist-base-path id)
+    path))
+
+
+(defn load
+  [persist-base-path]
+  (when-some [path (-> (list persist-base-path) first)]
```
neotyk, line 64: So when there would be multiple "snapshot" files, we'd just take first one?

neotyk (approved):
LGTM

src/clj/athens/self_hosted/web/persistence.clj
filipesilva: Yes, with no sorting or anything. The only case I think this can happen, given that `save!` will delete other snapshots, is if the process crashes after `save!` makes a new snapshot but before deleting the others.

If this happened multiple times, for whatever reason, then you'd end up with multiple snapshots. But the next time `save!` finishes successfully, all others will be deleted.

I think this is a pretty solid failure mode. It's even safe with multiple processes. Wink wink.

src/clj/athens/self_hosted/event_log.clj
```diff
@@ -55,25 +55,71 @@
...
+     (->> (utils/range-mapcat-while f empty?)
+          (map deserialize)
+          ;; The point here is to get all events since event-id.
+          ;; We're getting all the events, dropping every one until the first one we care about,
+          ;; then dropping that one too.
+          ;; This is a terrible way to do what we want.
+          ;; Instead we should filter out all the events we don't care about on the fluree query.
+          ;; But when I (filipe) tried to do that, it made each events-page query take 30s instead of 0.3s.
+          ;; This seems good enough for now.
+          ;; TODO: figure out a performant way to do this.
```
filipesilva, line 117: See https://github.com/fluree/db/issues/160

src/cljc/athens/common_events/resolver/atomic.cljc
```diff
@@ -34,7 +34,9 @@
                                         (if ref-block-exists?
                                           ref-block
                                           (throw (ex-info "Ref block does not exist" {:block/uid ref-uid})))
-                                        (common-db/get-parent db [:block/uid ref-uid]))
+                                        (if-let [parent (common-db/get-parent db [:block/uid ref-uid])]
+                                          parent
+                                          (throw (ex-info "Ref block does not have parent" {:block/uid ref-uid}))))
```
neotyk, line 39: What has happened here? Why did we start throwing this error here in this PR?
Also no test reflecting this change in behaviour.

Same comment for `:block/move`.

filipesilva: While replaying the team graph log I saw a nullpointer error here (specifically, when trying to reorder children if there's no parent). The nullpointer error was not caught by our current trycatch and would crash the process. 

I think the right thing is to throw a legible error here. You're right that there's no tests and there should be. Adding now.

PR #2081, opened 2022-03-15
fix: default config should use docker values

filipesilva:
Otherwise, adding things like a password via env var means the user has
to also include all docker specific config items.

This commit partially reverts
https://github.com/athensresearch/athens/commit/712759dabe37e4f5a6a61ebc1733e594021c65db.
Paths in the commit above were changed afterwards by https://github.com/athensresearch/athens/commit/f62f829f2cf9347446571ab3c9594443237d2ed3.


neotyk:
LGTM

PR #2089, opened 2022-03-21
fix block move after last block

sid597:


sid597:
Tests to perform

- [x] as first block
- [x] as the last block
- [x] as the first child of a block
- [x] as the sibing of the parent

I think other drag and drop result in one of the above cases

neotyk:
@sid597 can we get some tests with it?

sid597:
> @sid597 can we get some tests with it?

yes

neotyk (approved):
LGTM

PR #2096, opened 2022-03-27
feat: don't autoblock non-chrome browsers

filipesilva:
Close https://github.com/athensresearch/athens/issues/1718
Close https://github.com/athensresearch/athens/issues/1705

I tried this PR on firefox and it seems to open ok, so I guess the `case` problem in those issues went away.

### DIFF
--- a/resources/public/index.html
+++ b/resources/public/index.html
@@ -32,17 +32,17 @@
         localStorage.setItem("day8.re-frame-10x.show-panel","\"true\"")
         localStorage.setItem("day8.re-frame-10x.using-trace?","\"true\"")
       }
-      function isBrowserSupported() {
-        return userAgent.indexOf("chrome/") > -1;
+      function isBrowserUnsupported() {
+        return userAgent.indexOf("AppleWebKit/") > -1;
       }
       function showUnsupportedBrowserWarning() {
           let warnDiv = window.document.createElement("div");
-          warnDiv.innerText = "Only browsers based on Chrome/Chromium are supported";
+          warnDiv.innerText = "Safari based browsers based are not supported";
           window.document.body.appendChild(warnDiv);
       }
     </script>
     <script>
-      if (!isBrowserSupported()) {
+      if (isBrowserUnsupported()) {
         showUnsupportedBrowserWarning();
       } else {
         var electron = isElectron()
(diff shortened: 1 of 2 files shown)


filipesilva:
Best reviewed with the `hide whitespace` option.

tangjeff0:
Browsers

- [ ] Brave
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

OSes

- [ ] Windows
- [ ] Linux
- [ ] Mac

filipesilva:
I tested on Mac with chrome/brave/edge/firefox and they seem to work. 

Safari on mac does not work, neither does iOS Safari or Chrome (same engine). Safari on mac throws this error:
> SyntaxError: Invalid regular expression: invalid group specifier name

I traced it down to some lookbehinds we have on regexes, which aren't supported on Safari yet (https://caniuse.com/?search=lookbehind, https://github.com/athensresearch/athens/pull/2096).

resources/public/index.html
```diff
@@ -32,17 +32,17 @@
         localStorage.setItem("day8.re-frame-10x.show-panel","\"true\"")
         localStorage.setItem("day8.re-frame-10x.using-trace?","\"true\"")
       }
-      function isBrowserSupported() {
-        return userAgent.indexOf("chrome/") > -1;
+      function isBrowserUnsupported() {
+        return userAgent.indexOf("applewebkit/") > -1;
```
filipesilva, line 36: This isn't right, the user agent for chrome under macos is `'mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/100.0.4896.75 safari/537.36'`

filipesilva: Brave is also the above.

Firefox is `"mozilla/5.0 (macintosh; intel mac os x 10.15; rv:98.0) gecko/20100101 firefox/98.0"`.

Safari is `"mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/605.1.15 (khtml, like gecko) version/15.4 safari/605.1.15"`

filipesilva: I've updated this check and can verify it works as follows:

MacOS:
- Safari shows the `Safari based browsers based are not supported` message
- Chrome, Firefox and Brave don't

iOS:
- All browsers show the `Safari based browsers based are not supported` message.

tangjeff0 (approved):

waldyrious:
Just for the record, I confirmed the [demo site](https://athensresearch.github.io/athens) now works on Ubuntu in both Firefox v99.0 and Brave v1.37.116 (Chromium 100.0.4896.127).

PR #2100, opened 2022-03-29
feat: Link Creation Reporting

neotyk:


tangjeff0 (approved):

tangjeff0:
Seems good, but didn't look as closely at structural parser. We should do Eng Testing before release to make sure linking still works for in-memory db and is reported in posthog @neotyk

src/cljc/athens/common_events/graph/ops.cljc
```diff
@@ -195,3 +196,39 @@
...
+        new-blocks        (->> block-save-ops
+                               (map #(select-keys (:op/args %)
+                                                  [:block/uid :block/string])))
+        new-block-uids    (->> new-blocks
+                               (map :block/uid)
+                               set)
+        new-structures    (->> new-blocks
+                               (map :block/string)
+                               (map structure/structure-parser->ast))
+        old-block-strings (->> new-block-uids
```
filipesilva, line 214: What happens on `:block/save` over a uid that doesn't exist in the db? That should be an invalid operation. But does it crash the runtime or is it handled correctly?

src/cljc/athens/common_events/graph/ops.cljc
```diff
@@ -195,3 +196,39 @@
...
+                                 set))
+        old-links         (links old-structures
+                                 #{:page-link :hashtag :block-ref}
+                                 {:hashtag :page-link})
+        new-links         (links new-structures
+                                 #{:page-link :hashtag :block-ref}
+                                 {:hashtag :page-link})
+        removed-links     (set/difference old-links new-links)
+        added-links       (set/difference new-links old-links)]
+    [removed-links added-links]))
```
filipesilva, line 234: Since this fn operates over atomic operations and is in the graph ops ns, it would be nice if the output uses identity in the sense defined in `athens.common-events.graph.schema`.

Right now it outputs what I think is parser names like `[:block-ref "123"]` and `[:page-link "123"]`, but in `athens.common-events.graph.schema` we have identity defined as `[:block/uid "123"]` and `[:page/title "123"]` instead.

src/cljc/athens/common_events/graph/ops.cljc
```diff
@@ -195,3 +196,39 @@
                            (map :block/new)
                            set)]
     new-uids))
+
+
+(defn structural-diff
```
filipesilva, line 201: nit: I think this function is mostly doing the same as linkmaker (especially the 1-arity), except it looks at the atomic op string instead of datoms.

Linkmaker 1-arity looks like this:
```clj
     (let [datoms        (d/datoms db :eavt)
           linkmaker-txs (into []
                               (comp (keep parseable-string-datom)
                                     (mapcat (partial update-refs db)))
                               datoms)]

```

It gets entity ids and strings from the db, and creates a tx via `update-refs`:

```clj
(defn update-refs-tx
  "Return the tx that will update lookup ref's :block/refs from before to after.
   Both before and after should be sets of lookup refs."
  [lookup-ref before after]
  (let [[only-before only-after] (data/diff before after)
        to-tx (fn [type ref] [type lookup-ref :block/refs ref])]
    (set (concat (map (partial to-tx :db/retract) only-before)
                 (map (partial to-tx :db/add) only-after)))))

(defn update-refs
  "Returns updated refs for eid.
   Returns nil if eid is no longer in db."
  [db [eid string]]
  (when-let [lookup-ref (eid->lookup-ref db eid)]
    (let [before     (block-refs-as-lookup-refs db lookup-ref)
          after      (string-as-lookup-refs db string)]
      (update-refs-tx lookup-ref before after))))
```

The overall logic here is the same I think: given a db, a entity, and a string, check the difference in refs between what's in the db right now and what's in the string. But on this function you don't want the output as a tx, I think you just want the `data/diff` between the lookups, because `data/diff` gives you `[removed-links added-links]`.

filipesilva (changes requested):
Only comment that I really want addressed is the one about the missing uid.

src/cljc/athens/common_events/graph/ops.cljc
```diff
@@ -195,3 +196,39 @@
...
+  "Calculates removed and added links (block refs & page links)"
+  [db ops]
+  (let [block-save-ops    (contains-op? ops :block/save)
```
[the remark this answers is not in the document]
neotyk: Do you mean like in `:node/title`?

src/cljc/athens/common_events/graph/ops.cljc
neotyk: then old-block-strings will have `nil` instead of old string

src/cljc/athens/common_events/graph/ops.cljc
neotyk: For now this data is used for reporting only, and it's ok this way.
We could change the returned format, but why bother if it's not used like that?

src/cljc/athens/common_events/graph/ops.cljc
neotyk: I've tried `data/diff` it's output is confusing with all these `nil` all over the place.
Maybe linkmaker could use the structural-diff, though it's entangled with Atomic Ops.

I agree that similar things are happening in both places.
We don't have a common abstraction between the two just now.

src/cljs/athens/electron/monitoring/core.cljs
```diff
@@ -58,3 +58,34 @@
...
+  :reporting/block.link
+  (fn [{:keys [_db]} [_ {:keys [source count]}]]
+    {:posthog/capture-event! {:event-name "feature/block.link"
+                              :opts-map   {:source        source
+                                           :links-created count}}}))
+
+
+(defn build-reporting-link-creation
+  [added source]
+  (let [{:keys [page-link block-ref]} (reduce (fn [agg [link-type _]]
```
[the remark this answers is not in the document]
neotyk: look-alike, not the same tho

src/cljc/athens/common_events/graph/ops.cljc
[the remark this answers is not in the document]
filipesilva: Yes, those are refs too.

src/cljc/athens/common_events/graph/ops.cljc
filipesilva: Then let's wait until the third case where we're doing the same similar thing to create the abstraction.

filipesilva (approved):

src/cljc/athens/common_events/graph/ops.cljc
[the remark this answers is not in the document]
neotyk: Those should be captured too for completeness

PR #2118, opened 2022-04-05
fix: use long instead of bigint for event log

filipesilva:


.clj-kondo/rewrite-clj/rewrite-clj/config.edn
```diff
@@ -0,0 +1,5 @@
+{:lint-as
```
filipesilva, line 1: This was auto-added by a recent clj-kondo.

PR #2129, opened 2022-04-11
permalink

filipesilva:
- feat: support loading a url on first boot on web client
- feat: support copying permalink

### DIFF
--- a/src/cljs/athens/router.cljs
+++ b/src/cljs/athens/router.cljs
@@ -76,48 +76,46 @@
           nav-page?      (= :page-by-title route-name)
           controllers    (rfc/apply-controllers (:controllers old-match) new-match)
           loading?       (:loading? db)]
-      (when-not loading?
-        (if nav-page?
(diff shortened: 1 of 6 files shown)


src/cljs/athens/router.cljs
```diff
@@ -233,3 +231,37 @@
...
+    ;; Replace history with a version without the graph param.
+    (.. url -searchParams (delete graph-param-key))
+    (js/history.replaceState js/history.state nil url)
+    graph-id))
+
+
+(defn create-url-with-graph-param
+  "Create a URL containing graph-id."
+  [graph-id]
+  (let [url (js/URL. js/window.location)]
```
neotyk, line 264: this is weird in case of electron, because `file:///Users/neotyk/src/athens/resources/public/index.html?graph=https%3A%2F%2Fteam.athensresearch.org#/`

filipesilva: I should remove the permalink button on electron, did so now.

PR #2140, opened 2022-04-14
feat: Navigate back when user deletes current page

neotyk:


filipesilva (approved):
LGTM but has conflicts

PR #2156, opened 2022-04-22
fix: `:page/rename` & `:page/merge` w/o regex injection.

neotyk:


src/cljc/athens/common_db.cljc
```diff
@@ -280,12 +279,12 @@
...
-         (let [[string kw] (if title
+         (let [[text kw]   (if title
                              [title :node/title]
                              [string :block/string])
-               new-str (string/replace string
-                                       (patterns/linked old-title)
-                                       (str "$1$3$4" new-title "$2$5"))]
+               old-wrapped (str "[[" old-title "]]")
+               new-wrapped (str "[[" new-title "]]")
+               new-str     (string/replace text old-wrapped new-wrapped)]
```
filipesilva, line 287: I think you also need to replace the the `#name-here` format here. Our parser recognizes plain `#alex` as well as `#[[alex]]` and renders both as a ref.

filipesilva (approved):

src/cljc/athens/common_db.cljc
filipesilva: On the web preview version I made these three bullets:
```
- #alex
- #[[alex]]
- [[alex]]
```

Then renamed the page from `alex` to `alexxx`, and ended up with these:
```
- #alex
- #[[alexxx]]
- [[alexxx]]
```

neotyk: that plus one more case: `alex` to `al ex` is handled
thanks for spotting it

neotyk: @filipesilva last commit fixes it

PR #2167, opened 2022-04-27
Revert "fix: row title is addressed by full name"

filipesilva:
Reverts athensresearch/athens#2166


filipesilva:
Reverting fix, need to merge in https://github.com/athensresearch/athens/pull/2165 first

PR #2172, opened 2022-04-27
enhance: remove devtool

tangjeff0:
Never really worked, and the other datascript devtool should be better (but also doesn't work right now).


src/cljs/athens/views/devtool.cljs
```diff
@@ -1,394 +0,0 @@
...
-    [cljs.pprint :as pp]
-    [clojure.core.protocols :as core-p]
-    [clojure.datafy :refer [nav datafy]]
-    [datascript.core :as d]
-    [datascript.db]
-    [me.tonsky.persistent-sorted-set]
-    [re-frame.core :refer [subscribe dispatch]]
-    [reagent.core :as r]
-    [reagent.ratom]
-    [sci.core :as sci])
```
filipesilva, line 16: Can you remove this dependency from `deps.edn` as well please?

filipesilva (approved):

src/cljs/athens/views/devtool.cljs
tangjeff0: Yes

PR #2196, opened 2022-05-25
Support Safari on Mac and iOS

filipesilva:
Followup from https://github.com/athensresearch/athens/pull/2096#issuecomment-1083101498.

This PR removes lookbehind on regexes, and the check that prevents Safari from loading the app.


src/cljs/athens/views/athena.cljs
```diff
@@ -22,14 +23,21 @@
...
-                       [:> Text {:class "result-highlight"
-                                 :key i} part]
-                       part))
-                   (str/split txt query-pattern)))))
+  (if-not query
+    txt
+    (let [query-pattern (patterns/highlight query)]
+      (doall
+        (map-indexed (fn [i part]
+                       (println i part query-pattern (re-find query-pattern part))
```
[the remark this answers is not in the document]
filipesilva: removed

src/cljs/athens/views/athena.cljs
```diff
@@ -231,7 +239,7 @@
                      :prefix   "Create page"
                      :preview  nil
                      :type     :page
-                     :query    query
+                     :query    nil
```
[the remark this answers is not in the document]
filipesilva: Because this `result-el` is to show the `Create page: <page name>` result, and shouldn't highlight `page name`. I changed `highlight-match` to not try to do any highlighting if there's no query.

test/athens/parser/impl_test.cljc
```diff
@@ -231,10 +231,17 @@
     (utils/parses-to sut/inline-parser->ast
 
                      ;; Any ASCII punctuation character may be backslash-escaped
+                     ;; NOTE: not working in JS environment same as in JVM
```
filipesilva, line 234: @neotyk would like your input on specialcasing this test. There's another test (a latex one) that is specialcased for cljs too. 

I think this isn't a big issue because the hashtag tests already test that word chars before a hashtag will stop it from being a hashtage.

test/athens/cljs_parser_test.cljs
```diff
@@ -1,816 +0,0 @@
-(ns athens.cljs-parser-test
```
neotyk, line 1: Why do we want to get rid of it?

filipesilva: I rolled it into the `test/athens/parser/impl_test.clj -> impl_test.cljc` change, but am not sure if that's the right thing to do. Seemed to be the same tests except for the clj/cljs differences.

PR #2203, opened 2022-06-01
Comments using metadata

sid597:
- [x] Comments using list data model
- [x] Threads using list data model
- [ ] Comments using `comment/on` data model
- [ ] Threads using `comment/on` data model


sid597:
see #2221

PR #2224, opened 2022-06-17
Support audio files

aldosolorzano:


src/cljs/athens/parse_renderer.cljs
```diff
@@ -265,6 +265,12 @@
                                       :borderRadius "md"
                                       :alt   alt
                                       :src   url}])
+     :url-audio            (fn [{url :src alt :alt}]
```
aldosolorzano, line 268: This is not working, dunno how if chakra ui supports audio

src/cljs/athens/electron/utils.cljc
```diff
@@ -70,6 +71,7 @@
    :id         base-dir
    :base-dir   base-dir
    :images-dir (.resolve (path) base-dir IMAGES-DIR-NAME)
+   :audios-dir (.resolve (path) base-dir AUDIOS-DIR-NAME)
```
aldosolorzano, line 74: Even with this change, when I tried to access to `:audios-dir` it returns nil

src/cljs/athens/electron/audio.cljs
```diff
@@ -0,0 +1,55 @@
+(ns athens.electron.audio
+  (:require
+    [athens.common.utils :as common.utils]
+    [athens.db :as db]
+    [athens.electron.utils :as electron.utils]
+    [re-frame.core :as rf]))
+
+(defn save-audio
```
aldosolorzano, line 8: This and the save-image function share a big portion of code, maybe makes sense to refactor the common code to copy files

src/cljc/athens/parser/impl.cljc
```diff
@@ -98,6 +99,7 @@ strikethrough = <#'~~(?!\\s)'>
 
 link = md-link
 image = <'!'> md-link
+audio = <'^'> md-link
```
aldosolorzano, line 102: I just did some copy and paste, but I don't fully understand how the interpreter works

tangjeff0:
Hi @aldosolorzano, thank you so much for the PR! Some questions around this:

- We think we will move away from markdown syntax eventually to be more rich-text. I'm not sure if we want to support more syntax like `^[]()`, which is not really standard in markdown.
- I wasn't able to see anything rendered when I linked to an audio file. What kind of UI did you want to present?
- Random question: what kind of audio files did you want to link to in Athens?

aldosolorzano:
> Hi @aldosolorzano, thank you so much for the PR! Some questions around this:

Hi @tangjeff0 thank you for answering and sorry for taking long to reply, now I have some time to try again this.

> * We think we will move away from markdown syntax eventually to be more rich-text. I'm not sure if we want to support more syntax like `^[]()`, which is not really standard in markdown.

Yes agree with you, I randomly chose that syntax as I could find an standard, to what are you planning to migrate ?

> * I wasn't able to see anything rendered when I linked to an audio file. What kind of UI did you want to present?

Yes, that's the problem I had, I was able to copy the audio to the athens folder and convert it as `^[](path/to/file)` when you drag the file but I cant' render it with a simple player. I tried simple html <audio> tag and it didn't work. I think that a super simple play and stop could work to test if it makes sense to have it and invest more time.

> * Random question: what kind of audio files did you want to link to in Athens?

Me and my brother like to record audio from natural places, such as ocean, woods, birds etc... and use them for research and to make music. One day I was talking to him while using athens and it clicked me, I could use athens as a way to share notes over an audio file and not send a whatsapp explaining my idea and then audio. Also as I personal log for adding ideas for the specific recordings and have a centralize place.
Probably dealing with audio can become complicated and heavy, but as an experiment I think is worth it to give it try.

PR #2226, opened 2022-06-19
Improve page headers UX

shanberg:
- Moves page header URL controls into new 'Properties' popover control in page headers, replacing previous toggle
- Adds scaffolding for future properties in the same popover
- Adds simple validation and logic to the header URL control to prevent setting bad states and make it easier to clear


filipesilva (changes requested):
On the web client using the in-memory db, the first time you click on the cover image input box, the box disappears. I think because it saves an empty string and re-renders the page.

The properties box shouldn't be there if cover image isn't turned on via feature flags, as there's no other property it controls atm.

tangjeff0:
Subsumed by #2235

PR #2231, opened 2022-06-26
feat: more consistent style for refs

shanberg:
Use a dotted gold border to indicate embedded content

Addresses inline and block embeds in content.

CleanShot 2022-06-26 at 15 17 31@2x


tangjeff0:
holding off

PR #2234, opened 2022-07-01
fix: don't show empty dbs section when none exist

shanberg:
- Hides "Other databases" in DB menu when none exist


src/cljs/athens/electron/db_menu/core.cljs
```diff
@@ -57,36 +57,37 @@
...
+                                 :is-disabled (= sync-status :synchronising)
+                                 :is-current false
+                                 :key key}]))
+              (when (= :synchronising sync-status)
+                [:> VStack {:align "center"
+                            :background "background.vibrancy"
+                            :backdropFilter "blur(0.25ch)"
+                            :justify "center" :position "absolute" :inset 0}
+                 [:> Spinner]
+                 [:> Text "Syncing..."]])]])
```
filipesilva, line 90: Shouldn't the sync message still be there regardless? I think with these changes it'll be hidden when there's no other db.

shanberg:
OBE

PR #2241, opened 2022-07-09
fix: docker-compose.yml only need 1 restart in each services

cloudsbird:
It's Minor fix.

I see in our docker-compose.yml file in nginx service part the `restart: always` is written twice so I fix it.


filipesilva (approved):
Thanks for this fix!

PR #2253, opened 2022-07-18
pass string to block/move

sid597:


src/cljs/athens/events.cljs
```diff
@@ -1598,7 +1598,9 @@
   :block/move
   (fn [_ [_ {:keys [source-uid target-uid target-rel local-string] :as args}]]
     (log/debug ":block/move args" (pr-str args))
-    (let [sentry-tx (close-and-get-sentry-tx "block/move")
+    (let [local-string (or local-string
```
neotyk, line 1601: sentry tx best be created 1st so it captures most of processing

PR #2254, opened 2022-07-18
enhancement: better position, clearer text for reaction tooltip

shanberg:
CleanShot 2022-07-18 at 09 08 36@2x

- Reaction tooltip now linked to whole reaction button
- Tooltip should neatly concatenate list of users who reacted, ("stuart, jeff, and sid reacted with 👍 "


sid597 (approved):
GG

PR #2257, opened 2022-07-27
property drag and drop

filipesilva:
- fix: drop after/before prop moves block to first
- fix: drag and drop properties


tangjeff0 (approved):
Seems like an improvement, but dragging the property key should probably do the same thing.

tangjeff0:
Dragging block onto it's child makes them disappear https://www.loom.com/share/ad4293138373425480f7e70b15261f75

filipesilva:
@tangjeff0 fixed both issues. Note that the `Dragging block onto it's child makes them disappear` bug was unrelated to properties, and present in main as well.

PR #2259, opened 2022-07-28
search perf

filipesilva:
- fix: autocomplete should be autoComplete
- rfct: optimize search-in-block-content for comments
- rfct: remove unused history listener


sid597 (approved):
lgtm gg.

PR #2263, opened 2022-07-29
enhance: use settings as a modal, not as a page

tangjeff0:
before
image

after


filipesilva (approved):

tangjeff0:
For some reason, using 2 `when`s is slower than using one `cond`.

2 whens
image

cond
image

PR #2266, opened 2022-07-30
Improve block performance

shanberg:
Outline renderer blocks watch to see when they're in the viewport.

When they are, they'll render their more expensive parts. When they aren't, they'll render leaner versions of themselves.


src/cljs/athens/views/blocks/types/default.cljs
```diff
@@ -361,7 +361,7 @@
 
               (when (= @drag-target :before) [drop-area-indicator/drop-area-indicator {:placement "above"}])
 
-              [editor/editor-component
+              [:f> editor/editor-component
```
filipesilva, line 364: What does this change do?

filipesilva (approved):
LGTM

src/cljs/athens/views/blocks/types/default.cljs
tangjeff0: https://github.com/reagent-project/reagent/blob/master/doc/ReactFeatures.md#function-components

PR #2269, opened 2022-08-08
perf: leaner reactions

shanberg:
Adds separate menu section for reactions to reduce use of heavy emoji picker component


src/cljs/athens/views/blocks/types/default.cljs
```diff
@@ -316,9 +314,9 @@
                                                          :onClick  #(ctx-menu/handle-click-comment % uid)
                                                          :icon     (r/as-element [:> ChatIcon])}])
                                          (when reactions-enabled?
-                                           [:> MenuItem {:children "Add reaction"
-                                                         :onClick  show-emoji-picker-fn
-                                                         :icon     (r/as-element [:> ThumbUpFillIcon])}])
+                                           [:<>
+                                            [:> MenuDivider]
+                                            [reactions/reactions-menu-list uid "Stuart"]])
```
filipesilva, line 319: Don't think `"Stuart"` should be here.

filipesilva (approved):
LGTM but shows
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at eval (js/compiled/cljs-runtime/module$node_modules$$chakra_ui$tooltip$dist$chakra_ui_tooltip_cjs_dev.js:14:8)
    at div
```

can you take a look before merging?

PR #2272, opened 2022-08-10
fix: enter on prop should split block

filipesilva:
- fix: enter on prop should split block
- fix: block-el should prioritize the reactive props


neotyk (approved):
LGTM

PR #2277, opened 2022-08-12
Add location indicator

shanberg:
- Add Location indicator component to toolbar
- Should show current page or current block

TODO
- Link to actual current location
- Populate with real actions, not fake actions


tangjeff0:
subsumed by #2280

PR #2285, opened 2022-08-17
Fix bugs in comments

sid597:
- don't get notified if someone comments on a block you authored
- see no. of messages on inbox icom
- show comments box when block is closed


tangjeff0 (changes requested):
see comment in Athens

tangjeff0 (approved):

PR #2288, opened 2022-08-18
feat: import from roam using internal representation

filipesilva:
This enables importing from roam into Athens CKM by using the merge button in db picker.

You can test it using an EDN export of https://roamresearch.com/#/app/athensresearch for a real db (~5mb), or https://roamresearch.com/#/app/fstest6 for a mostly empty db (~4kb).

Suffers heavily from re-frame-10x being on by default (see https://github.com/athensresearch/athens/pull/2303). 

I suggest testing this PR with an in memory db via the vercel deploy, and against a local server via a dev build with re-frame-10x turned off.


src/cljc/athens/common_events/bfs.cljc
```diff
@@ -128,6 +128,8 @@
   "Convert internal representation to the vector of atomic operations that would create it.
   :block/save operations are grouped at the end so that any ref'd entities are already created."
   [db internal-representation default-position]
+  (when-not (vector? internal-representation)
```
tangjeff0, line 131: This will cause some notification ops to fail. I saw that some of these were lists, not vectors. @sid597

src/cljc/athens/common_events/resolver/atomic.cljc
```diff
@@ -18,17 +18,22 @@
...
-                                      :block/create event-ref
-                                      :block/edits  event-ref}
-        position-tx                  (condp = (position/position-type position)
-                                       :child    (position/add-child db uid position event-ref)
-                                       :property (position/add-property db uid position event-ref))
-        tx-data                      (into [new-block] position-tx)]
-    tx-data))
+  (let [{:block/keys [uid position]} args]
+    (if (common-db/block-exists? db [:block/uid uid])
+      ;; Treast :block/new on an existing block as a :block/move instead.
```
tangjeff0, line 23: `Treats`

tangjeff0: I don't understand why we'd want to do this tho

tangjeff0 (approved):

tangjeff0:
I tested importing https://roamresearch.com/#/app/help/ into web build -- 3.3MB. Didn't test on desktop.

src/cljc/athens/common_events/resolver/atomic.cljc
filipesilva: It's not so much "why do we want to do this" as "what do we do when this happens". Currently, when this happens, we end up with a block on two parents. That's wrong.

You can't really prevent the operation from being issued so it is more sensible to have semantics for its resolution. The two major options here are "error out" or "do something that's not wrong". I already ran into a case (repeat imports of the same content, and not just whole db imports) that I want to support, so I chose "do something that's not wrong".

src/cljs/athens/views/right_sidebar/events.cljs
```diff
@@ -105,7 +105,7 @@
                           (filter #(= (:name %) block-page-uid))
                           first
                           :source-uid)]
-      {:fx [[:dispatch [:properties/update-in [:block/uid update-uid] [(shared/ns-str "/items/type")]
+      {:fx [[:dispatch [:graph/update-in [:block/uid update-uid] [(shared/ns-str "/items/type")]
```
filipesilva, line 112: I need to check if `:properties/update-in` needs to be updated anywhere else.

src/cljs/athens/import/roam.cljs
```diff
@@ -60,94 +61,84 @@
...
-  (->> (d/q '[:find [?pages ...]
-              :in $ ?fn ?shared
-              :where
-              [_ :node/title ?pages]
-              [(?fn ?shared ?pages)]]
-            roam-db
-            gett
-            shared-pages)
-       sort))
+;; roam page -> IR
```
filipesilva, line 77: nit: remove comments

src/cljc/athens/common_events/bfs.cljc
filipesilva: Fixed by not throwing on lists

neotyk (approved):

PR #2298, opened 2022-08-20
Fix misc icon issues

shanberg:
- Fixes incorrect icon size in collapsed toolbar
- Re-exports icons without extra junk


tangjeff0:
Getting an error when trying to open the preview

PR #2299, opened 2022-08-20
Reuse comments toast

shanberg:
Comments toggle toast should now be reused instead of creating a new one every time


tangjeff0 (changes requested):
if you wait for toast to disappear, and you click on toggle again, there is no toast again

PR #2303, opened 2022-08-22
perf: Do not load `re-frame-10x` by default

neotyk:
Also bumped `re-frame-10x` to a recent version.


tangjeff0:
Did you say you changed something with inline subscriptions? @neotyk

tangjeff0:
Looks like you did it in #2278

PR #2315, opened 2022-08-26
Notifications for assigned task

sid597:
Notification is created for
- the user who is assigned the task
- user who assigned the task


neotyk (approved):
LGTM

PR #2320, opened 2022-08-26
rfct: nicer comments, comment anchors have menu

shanberg:


shanberg:
> Noice: how about adding reactions to comments?

We have the UI for this, just need to add it as an option to the menu

PR #2321, opened 2022-08-27
perf: wait to update sidebar size on graph

shanberg:
- Right sidebar resizes with frontend state, saves to graph after a second
- Layout animation is disabled while resizing

### DIFF
--- a/src/cljs/athens/views.cljs
+++ b/src/cljs/athens/views.cljs
@@ -2,7 +2,6 @@
   (:require
     ["/components/App/ContextMenuContext" :refer [ContextMenuProvider]]
     ["/components/Layout/MainContent" :refer [MainContent]]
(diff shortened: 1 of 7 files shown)


tangjeff0 (dismissed):
When I resize the right sidebar very quickly, there's some lag, which causes the ResizeController to have different final width than the other layout elements:

image

tangjeff0:
> Given that the graph can break in its current state

I don't think we have full certainty that this broke the graph, though one followed the other. It might've, but the root issue still hasn't been solved from a data perspective.

tangjeff0:
The other thing is that I think there is probably a more generic way to be doing this, because I think we might have this pattern again.

tangjeff0 (approved):

PR #2325, opened 2022-08-29
fix: right sidebar should trigger toolbar when scrolled

shanberg:
- Scroll state now lives in layout state, not apptoolbar
- Scroll state now contains a list of elements scrolled past their respective readbility limits
- Right sidebar uses an invisible box with inView, for convenience


tangjeff0 (approved):
Nice, this was bothering me

PR #2329, opened 2022-08-30
Create sibling for block with just comments

sid597:


src/cljs/athens/events.cljs
```diff
@@ -1092,27 +1092,36 @@
...
+                                              (.getAttribute "data-uid"))
+                                      (str parent-uid "-embed-" embed-id))
+        root-block?                (boolean (:node/title parent))
+        context-root-uid           (get-in rfdb [:current-route :path-params :id])
+        new-uid                    (common.utils/gen-block-uid)
+        has-children?              (seq (common-db/sorted-prop+children-uids @db/dsdb [:block/uid uid]))
 
         {:keys [value start]} d-key-down
         event                 (cond
+                                (and (not-empty (get block-properties ":comment/threads"))
```
tangjeff0, line 1118: can you give this a condition a name? bit hard to read this.

src/cljs/athens/events.cljs
```diff
@@ -1092,27 +1092,36 @@
...
+                                     has-children-without-prop?
```
tangjeff0, line 1119: I don't understand the name of this. Shouldn't it be that it doesn't have children?

src/cljs/athens/events.cljs
```diff
@@ -1092,27 +1092,36 @@
...
+                                     (= start (count value)))
```
tangjeff0, line 1120: I think this condition should be given a name, and the rest of the cases should use it.

tangjeff0 (changes requested):

tangjeff0 (approved):

PR #2343, opened 2022-09-05
feat: Task indent & unindent.

neotyk:


tangjeff0:
The caret gets swallowed sometimes, but still an improvement.

PR #2344, opened 2022-09-05
build: use react 18

filipesilva:


package.json
```diff
@@ -125,11 +125,11 @@
     "luxon": "^2.0.2",
     "nedb": "^1.8.0",
     "polished": "^4.1.3",
-    "react": "17.0.1",
+    "react": "18.2.0",
     "react-codemirror2": "^7.2.1",
     "react-colorful": "^5.4.0",
     "react-day-picker": "^7.4.10",
-    "react-dom": "17.0.1",
+    "react-dom": "18.2.0",
```
filipesilva, line 132: re-frame-10x is using the old bootstrap and will error on startup, but I think it might be ok? See https://github.com/day8/re-frame-10x/issues/367

shanberg: @filipesilva does this issue cause everything to fall back to 17? I assumed it would.

filipesilva: I don't think so. Just that instantiation of react rendering, I think. You can have multiple react "instances" on the same page, controller different parts of the html, without interacting with each other.

shanberg (approved):

PR #2352, opened 2022-09-07
don't hide notifications and add green mark on the bullet in outline

sid597:


tangjeff0:
Nice, can you remove the old context-menu code now too?

PR #2353, opened 2022-09-07
Revert "Revert "Merge pull request #2344 from filipesilva/react-18""

filipesilva:
Reverts athensresearch/athens#2351


filipesilva:
Problems we found after the react 18 update:
- typing `[[` creates `[][]`
- dev auto reload doesn't work
- some toasts not working

filipesilva:
So more problems and details:
- typing `a [[` will work, it's just the very first bracket that doesn't get the cursor in the right place (inside)
- typing `a ((((((` will add the closing parens, but if you leave that bullet you will something like `a ))))`, and a warning on the console

filipesilva:
Tracked the pair chair problem to `athens.views.blocks.textarea-keydown`, specifically with how `handle-pair-char` ends up calling `(set-cursor-position target new-idx)`. For some reason it doesn't seem to do anything now if the pair chair is the first char.

filipesilva:
The `))))` problem is weirder... it looks like the bullet has the right content, the textbox too, but the bullet when rendered without editing doesn't?

filipesilva:
Also need to check if the 200ms txs on team graph are related, I saw them around this time but now they are gone.

PR #2376, opened 2022-09-21
Show title on blocks too

shanberg:
Before, only pages contributed a title to the toolbar.
Now blocks and routes (e.g. Daily Notes, Graph) will too.


filipesilva (approved):
I don't think this is working for blocks:
image

shanberg:
> I don't think this is working for blocks:
> 
> image
> 
> 

It should only appear after you've scrolled down and the native title is offscreen, otherwise it’s redundant. Is that not happening?

PR #2385, opened 2022-09-25
refactor, polish notifications

shanberg:
Goal: Clean up notifications so they're easier to use and more likely to get used

Major
- Notifications grouped by subject (page, block)
- Titles and content are parsed
- Titles limited to 2 lines
- More space for notifications so the text is easier to read
- Refactored implementation with less magic

Minor
- Style adjustments


filipesilva:
IMHO @sid597 is the person that can help here, not me. Nothing important registers for me but he might have insight on the changes in src/cljs/athens/views/notifications/popover.cljs.

PR #2407, opened 2022-10-05
.block-page and .node-page class names added

alwx:
After the recent code changes the `.block-page` and `.node-page` class names were removed => the following line in `athens.views.blocks.editor/find-selected-items` returns `nil` which makes it impossible to do manipulations with multi-line texts.
```clj
(or (.. target (closest ".node-page"))
    (.. target (closest ".block-page")))
```
Screenshot 2022-10-05 at 13 32 04

This PR fixes that.


filipesilva (approved):
LGTM but will leave for sid to move forward.

filipesilva:
@alwx thanks for fixing this! Was hitting the problem myself, but hadn't gotten around to taking care of it.

